PR #10, opened 2019-04-10
Make it work with dev model in k8s cluster

vincent-pli:
I don't know, maybe tekton/pipeline changed after dashboard code nail down.
I try with dev model by `npm run start` it's not work with my latest pipeline, issues is mainly about k8s api, see the diff, thanks

After the change I could see the list of pipelines, but cannot go further, when I click on the item, another exception raised, seems the `status.condition` is null.

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -13,12 +13,12 @@ limitations under the License.
 
 import { deleteRequest, get, post, put } from './comms';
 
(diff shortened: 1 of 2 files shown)


skaegi:
Thanks @vincent-pli -- we're still wiring the two together. Please give us a few hours...

PR #11, opened 2019-04-10
Backend build changes and install yaml

dibbles:
Pretty simple set of changes, delivering:

- Dockerfile to build back end code (needs extending to include front end?)
- Yaml to install built code
- Readme with some notes for now on how to build the backend (needs modifying and relocating once front end included?)
- Modified endpoints for health/readiness to not require a namespace on the endpoint path in the yaml


dibbles:
ah.... looks like I need to change author email on my commit

PR #17, opened 2019-04-10
Add create manual PipelineRun API

jessm12:
This PR adds:

 - POST endpoint on /pipelinerun
 - PipelineRun create implementation method
 - Insecure Tasks and Pipeline
 - README update
 - Create PipelineResource method
 - main.go functionality to run Pipeline0 on install
 - Tests for PipelineRun creation and resource permutations
 - Dockerfile_test

Tests will need to be added under the tests folder once added to test the helm and registry secret


jessm12:
@a-roberts @ncskier

a-roberts:
@ncskier can you please take a look? Jess and I worked on this together so I'd be a terrible reviewer.

a-roberts:
@jessm12 I've been thinking about this and discussed with @skaegi about the best way to get in, we're doing a lot at once here so I reckon we should split this PR up as follows

1) Dockerfile_test first, and instructions for contributors to test first using it since we don't yet have automatic builds
2) any readme changes: but we should add the API definitions under a developer's guide (e.g. here are our endpoints, here are our return codes/responses)
3) the API + tests for PipelineRun
4) code for kicking off Pipeline0 (out of main.go)
5) we noticed in Gopkg.lock eventing-sources - need to check we have nothing importing knative unnecessarily in our codebase

skaegi:
@jessm12 @a-roberts -- talked on the side and this PR is going to get broken into pieces and re-submitted

PR #20, opened 2019-04-10
Global types

vtereso:
Utils.go is serving as the registration file except for the Resource struct, which is a global value so this helps separate responsibilities. A lot of Kubernetes repositories use this system of defining types within a types.go file when the value is applicable outside of a single file e.g. https://github.com/kubernetes/api/blob/master/core/v1/types.go


a-roberts:
Thanks @vtereso, this is a good idea and looks ok to me once your CLA is resolved 😄

a-roberts:
@vtereso you already know this based on our Slack chat but:

```
Step 9/9 : RUN CGO_ENABLED=0 NAMESPACE=default GOOS=linux go test -v
 ---> Running in e681d1f819b0
# github.com/tektoncd/dashboard/pkg/endpoints [github.com/tektoncd/dashboard/pkg/endpoints.test]
./register.go:18:2: imported and not used: "github.com/tektoncd/dashboard/vendor/github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
```

just needs resolving and everything else passes and LGTM! And then the conflict and your CLA.

a-roberts:
/approve

i tested this builds fine and our unit tests pass, thanks @vtereso!

Issue #25, opened 2019-04-11
Deployment should be labelled 


jessm12:
Under `install/tekton-dashboard-deployment.yaml`, the deployment should have a label that allows for easier manipulation of it, such as getting and deleting the deployment

a-roberts:
This is merged 😄 thanks

PR #26, opened 2019-04-11
Add deployment label

jessm12:
Add app label `app=tekton-dashboard-deployment` for issue https://github.com/tektoncd/dashboard/issues/25 and rename deployment


a-roberts:
Yeah, this gives us pod names of the format `tekton-dashboard-xxxxx` and the deployment indeed has the label, looks good to me!

/approve

Issue #16, opened 2019-04-10
Logs are not displayed when clicking on a step in a PipelineRun


AlanGreene:
The front-end and back-end handling of TaskRun logs are not currently in sync and need to be aligned to ensure the new combined project works as expected. The front-end `Log` container and presentational components should be updated to handle the changed `taskrunlog` response format returned by the API.

AlanGreene:
I have a fix for this on my fork. Waiting for CLA approval before opening PR.

PR #32, opened 2019-04-11
Update front end to handle change in taskrunlog response format

AlanGreene:
Fixes https://github.com/tektoncd/dashboard/issues/16

Update front end code to handle changed log response format for TaskRun / Step:
- from `String`
- to `{ ..., StepContainers: [{ Name: ..., Logs: [String]}] }`

### DIFF
--- a/src/components/Log/Log.js
+++ b/src/components/Log/Log.js
@@ -11,6 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+/* eslint-disable react/no-array-index-key */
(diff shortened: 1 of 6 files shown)


a-roberts:
I discussed this with Alan over slack, it's so the UI can pick up the changes @vtereso made recently to the way we present task run information and logs - so it's in a more presentable way on the UI.

Alan mentions all of his testing works (we don't yet have the two hooked up so it's difficult to test automatically/easily) so I'm gonna go ahead and approve this one and get it in.

/approve

Issue #24, opened 2019-04-11
REST API for creating PipelineRuns


jessm12:
Under this issue add a REST API endpoint that will take parameters and create a new `PipelineRun` for a specified `Pipeline`

When creating a new `PipelineRun` the information that will need to be provided to the POST method is as follows:

- The name of the `Pipeline` to use for that `PipelineRun`
- Information relating to the `PipelineResources` for a `PipelineRun`, if required 
- Information relating to the repository for that `PipelineRun`
- Helm and Registry secret names, if required

The request will be namespaced and this is the namespace that the PipelineRun will occur in

a-roberts:
/assign

I've worked on this too with @jessm12 and we should have a PR coming up soon for folks to try and test 😸

a-roberts:
Merged, thanks all for taking a look too 🎉 we're adding the API definition next.

PR #33, opened 2019-04-11
Add create PipelineRun API

jessm12:
For issue https://github.com/tektoncd/dashboard/issues/24, this PR adds:

- POST endpoint on /pipelinerun
- PipelineRun create implementation method
- Create PipelineResource method
- Tests for PipelineRun creation and resource permutations
- Env values in `tekton-dashboard-deployment.yaml`

(Tests will need to be added to test the helm and registry secret once infrastructure is in place for this)


install/tekton-dashboard-deployment.yaml
```diff
@@ -33,3 +33,16 @@ spec:
         env:
         - name: PORT
           value: "9097"
+        # Currently PipelineRuns MUST happen in the same namespace as the install namespace of the Tekton dashboard
+        # the current requirement is that the below namespace values match 
+        - name: PIPELINE_RUN_NAMESPACE
+          value: "replace_me"
+        - name: INSTALLED_NAMESPACE
+          value: "replace_me"
```
ncskier: Please remove the `PIPELINE_RUN_NAMESPACE` and `INSTALLED_NAMESPACE` environment variables. I can't find them being referenced in the code, so I'm not sure what they're supposed to add.

Also, the `createPipelineRun` function seems capable of creating a PipelineRun in any namespace, so why would we want to limit it to one `PIPELINE_RUN_NAMESPACE`?

pkg/endpoints/pipeline.go
```diff
@@ -112,44 +139,63 @@ func (r Resource) getPipeline(request *restful.Request, response *restful.Respon
...
 
 	pipelines := r.PipelineClient.TektonV1alpha1().Pipelines(namespace)
 	pipeline, err := pipelines.Get(name, metav1.GetOptions{})
 	if err != nil {
 		logging.Log.Errorf("could not retrieve the pipeline called %s in namespace %s", name, namespace)
-		return *pipeline, err
+		return v1alpha1.Pipeline{}, err
 	}
-	logging.Log.Debug("Found the pipeline definition OK")
+	logging.Log.Debugf("Found the pipeline definition OK")
```
ncskier: No need for the `f` because you are only logging a string.

pkg/endpoints/pipeline.go
```diff
@@ -112,44 +139,63 @@ func (r Resource) getPipeline(request *restful.Request, response *restful.Respon
...
+	} else {
+		// no errors so return the pipeline run list
+		logging.Log.Debugf("+%v", pipelinerunList.Items)
+		response.WriteEntity(pipelinerunList)
+	}
+}
+
+// GetAllPipelineRunsImpl ...
+/* Returns a pointer to a list of PipelineRuns in a given namespace,
+optionally with a repository query as a parameter */
```
ncskier: Please put this all in one multi-line comment, so we aren't mixing comment styles.

pkg/endpoints/pipeline.go
```diff
@@ -112,44 +139,63 @@ func (r Resource) getPipeline(request *restful.Request, response *restful.Respon
...
+
+	pipelineruns := r.PipelineClient.TektonV1alpha1().PipelineRuns(namespace)
+
 	var pipelinerunList *v1alpha1.PipelineRunList
 	var err error
 	if repository != "" {
-		server, org, repo := getGitValues(repository)
+		server, org, repo, err := getGitValues(repository)
+		if err != nil {
+			logging.Log.Errorf("there was an error getting the Git values: %s", err)
```
ncskier: Should `GetAllPipelineRunsImpl` return an `AppError` here since there was an error getting the Git values?

pkg/endpoints/pipeline.go
```diff
@@ -166,6 +212,112 @@ func (r Resource) getPipelineRun(request *restful.Request, response *restful.Res
...
+
+	pipelineName := pipelineRunData.PIPELINENAME
+
+	saName := ""
+	providedSaName := os.Getenv("PIPELINE_RUN_SERVICE_ACCOUNT")
+	if providedSaName != "" {
+		saName = providedSaName
+	} else {
+		saName = "default"
+	}
```
ncskier: Can we move this `provideSaName` logic into a reusable function? It seems like "general purpose" logic. I think the control flow of using the `default` service account name if the `PIPELINE_RUN_SERVICE_ACCOUNT` does not exist should be documented in a general location, such as the utils.

install/tekton-dashboard-deployment.yaml
```diff
@@ -33,3 +33,16 @@ spec:
...
+        # The service account to run PipelineRun pods under
+        # any secrets required for authentication will need to be patched to this service account
+        - name: PIPELINE_RUN_SERVICE_ACCOUNT
+          value: "replace_me"
```
ncskier: Can we add a comment that if this is left blank, the `default` service account will be used? I think that's what this code snippet is for: https://github.com/tektoncd/dashboard/pull/33/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yaml#diff-2a268056787157d73c4168498850de86R244.

Or maybe even leave this value blank by default, because the user does not necessarily need to provide a service account?

pkg/endpoints/pipeline.go
```diff
@@ -166,6 +212,112 @@ func (r Resource) getPipelineRun(request *restful.Request, response *restful.Res
...
+	logging.Log.Infof("PipelineRuns will be created with the service account %s", saName)
+
+	startTime := getDateTimeAsString()
+
+	// Assumes you've already applied the yml: so the pipeline definition and its tasks must exist upfront.
+	generatedPipelineRunName := fmt.Sprintf("devops-pipeline-run-%s", startTime)
+
+	pipeline, err := r.getPipelineImpl(pipelineName, namespace)
+	// getPipelineImpl returns a null pipeline object if it doesn't exist, so be safe and check for that
+	if err != nil || pipeline.Name == "" {
```
ncskier: I don't think you need to check if the `pipeline.Name == ""` here because `getPipelineImpl` returns an error if the pipeline is not found (https://github.com/tektoncd/dashboard/pull/33/files#diff-2a268056787157d73c4168498850de86R151).

Also, you can remove the  comment `// getPipelineImpl returns a null pipeline object if it doesn't exist, so be safe and check for that` because you changed the `getPipelineImpl` to return `v1alpha1.Pipeline{}` instead of nil.

jessm12:
Thanks for your detailed comments @ncskier 🙂 I definitely agree with your comment about the scope and we did have this as a consideration. I also had the expectation that the scope would be increased to allow for any kind of Pipeline to be run, whether it be user provided or provided by us. 

At the moment however I believe our aim was to support all Pipeline templates within our templates repository so this is what this PR was intended to support, we should document this I agree so once the API definition readme is updated I will be happy to add this.

I think to increase the scope of PipelineRun creation would require more discussion, time and structural changes. Maybe @mnuttall @a-roberts can comment on whether this is reasonable that we add this functionality supporting Pipeline templates within our repo initially with a view to increasing this to all (user provided or not) Pipelines in future

install/tekton-dashboard-deployment.yaml
jessm12: Good spot, looks like these aren't needed anymore, I have removed them

install/tekton-dashboard-deployment.yaml
jessm12: Yes, I have removed the replace_me and added a comment here to say that default will be used if left blank

pkg/endpoints/pipeline.go
jessm12: true, removing both

ncskier:
Ok, thanks @jessm12. That helps clarify things for my PR review 🙂 If I'm not mistaken, you're creating a separate PR for readme changes (https://github.com/tektoncd/dashboard/pull/17#issuecomment-481780979)? So you can document the limitations in that PR.

Also, I can create a new issue after I'm done with my review to discuss increasing the scope of PipelineRun creation.

pkg/endpoints/pipeline.go
```diff
@@ -59,6 +67,25 @@ type BuildRequest struct {
...
+type ManualPipelineRun struct {
+	PIPELINENAME      string `json:"pipelinename"`
+	PIPELINERUNTYPE   string `json:"pipelineruntype,omitempty"`
+	GITRESOURCENAME   string `json:"gitresourcename,omitempty"`
+	IMAGERESOURCENAME string `json:"imageresourcename,omitempty"`
+	GITCOMMIT         string `json:"gitcommit,omitempty"`
+	REPONAME          string `json:"reponame,omitempty"`
+	REPOURL           string `json:"repourl,omitempty"`
+	HELMSECRET        string `json:"helmsecret,omitempty"`
+	REGISTRYSECRET    string `json:"registrysecret,omitempty"`
```
ncskier: Sorry; this is another question about the scope of the PipelineRun creation 😬

Is there a reason why we don't want to specify the `SERVICEACCOUNT` in the `ManualPipelineRun` struct? This is a pretty strict design decision to say that the user can only create a PipelineRun using one service account. Since we only use the `PIPELINE_RUN_SERVICE_ACCOUNT` here https://github.com/tektoncd/dashboard/pull/33/files#diff-2a268056787157d73c4168498850de86R240, it seems to me like the code would function the same way if the user was able to specify whatever service account they want for the PipelineRun. Then we would also be able to remove the restriction of one `PIPELINE_RUN_SERVICE_ACCOUNT`.

However, please let me know if there is a benefit to restricting the user to one service account that I am not seeing at the moment. I know this used to be a bigger PR, so there might be a bigger-picture that I'm missing out on.

pkg/endpoints/pipeline.go
```diff
@@ -59,6 +67,25 @@ type BuildRequest struct {
...
+}
```
ncskier: I have the same comment regarding the `DOCKER_REGISTRY_LOCATION` environment variable that I had for the `PIPELINE_RUN_SERVICE_ACCOUNT` (https://github.com/tektoncd/dashboard/pull/33/files#r274901277). It seems unnecessary to restrict the user to one docker registry, when we can add this as a field in the `ManualPipelineRun` struct and allow users to use any docker registry they please.

Again, let me know if I'm missing out on a bigger-picture here.

pkg/endpoints/pipeline.go
jessm12: I agree! especially as ServiceAccount is something that can be defined in the spec for a PipelineRun, I will remove the env variable and take ServiceAccount as a field for the struct 👍

pkg/endpoints/pipeline.go
jessm12: I think you're right here too, with this new implementation it makes sense to allow the user to provide a value for this rather than having to hardcode one at install time!

pkg/endpoints/pipeline.go
```diff
@@ -11,23 +13,22 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package endpoints
```
ncskier: This is a nit, but please leave `package endpoints` below the copyright. It looks like we're kind of inconsistent with this in our repo, but I think most of the `package endpoints` lines are below the copyright. Also, I checked a file in the kubernetes and tektoncd/pipeline repos and they both have the `package NAME` below the copyright: https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1alpha1/artifact_bucket.go and https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1alpha1/artifact_bucket.go.

pkg/endpoints/websocket.go
```diff
@@ -1,3 +1,5 @@
+package endpoints
```
ncskier: See previous comment https://github.com/tektoncd/dashboard/pull/33/files?file-filters%5B%5D=.go&file-filters%5B%5D=.lock#r274927769

pkg/endpoints/utils.go
```diff
@@ -1,3 +1,5 @@
+package endpoints
```
ncskier: See previous comment https://github.com/tektoncd/dashboard/pull/33/files?file-filters%5B%5D=.go&file-filters%5B%5D=.lock#r274927769

pkg/endpoints/test-utils.go
```diff
@@ -1,3 +1,5 @@
+package endpoints
```
ncskier: See previous comment https://github.com/tektoncd/dashboard/pull/33/files?file-filters%5B%5D=.go&file-filters%5B%5D=.lock#r274927769

pkg/endpoints/pipeline_test.go
```diff
@@ -1,3 +1,5 @@
+package endpoints
```
ncskier: See previous comment https://github.com/tektoncd/dashboard/pull/33/files?file-filters%5B%5D=.go&file-filters%5B%5D=.lock#r274927769

pkg/endpoints/pipeline.go
jessm12: Yeah initially we had moved them above the copyright as the go linter was complaining that the package should be at the top of the file but it looks like the convention in projects we're working around is to keep the copyright first

pkg/endpoints/pipeline.go
```diff
@@ -112,44 +139,63 @@ func (r Resource) getPipeline(request *restful.Request, response *restful.Respon
...
 func (r Resource) getAllPipelineRuns(request *restful.Request, response *restful.Response) {
 	namespace := request.PathParameter("namespace")
-	pipelineruns := r.PipelineClient.TektonV1alpha1().PipelineRuns(namespace)
 	repository := request.QueryParameter("repository")
 
 	logging.Log.Debugf("In getAllPipelineRuns: namespace: `%s`, repository query: `%s`", namespace, repository)
 
+	pipelinerunList, appError := r.GetAllPipelineRunsImpl(namespace, repository)
+	if appError.ERROR != nil {
+		logging.Log.Debugf("+%v", pipelinerunList.Items)
```
ncskier: Logging the `pipelinerunList.Items` after the error isn't helpful to the user. We know the function returns an empty list `v1alpha1.PipelineRunList{}` on errors, so this log is just going to print an empty list.

Furthermore, this log seems sketchy because if the `GetAllPipelineRunsImpl()` function returned `nil` along with its error message, this log could result in a crash (we know this won't happen because it just returns `v1alpha1.PipelineRunList{}`, but it still isn't a great coding practice).

pkg/endpoints/pipeline.go
```diff
@@ -112,44 +139,63 @@ func (r Resource) getPipeline(request *restful.Request, response *restful.Respon
...
+		utils.RespondError(response, appError.ERROR, appError.CODE)
+	} else {
+		// no errors so return the pipeline run list
+		logging.Log.Debugf("+%v", pipelinerunList.Items)
```
ncskier: I think you should add a description string to this debug message, so it is easier for the user to track down where the error is coming from, and what it is referring to. For example, `logging.Log.Debugf("PipelineRun list: +%v", pipelinerunList.Items)`

pkg/endpoints/pipeline.go
```diff
@@ -112,44 +139,63 @@ func (r Resource) getPipeline(request *restful.Request, response *restful.Respon
...
+		response.WriteEntity(pipelinerunList)
+	}
+}
+
+// GetAllPipelineRunsImpl ...
+/* Returns a pointer to a list of PipelineRuns in a given namespace,
```
ncskier: It looks like it returns a list of PipelineRuns instead of a pointer to the list.

pkg/endpoints/pipeline.go
```diff
@@ -414,12 +610,19 @@ func getGitValues(url string) (gitServer, gitOrg, gitRepo string) {
 			repoURL = strings.TrimPrefix(url, "http://")
 		}
 	}
+
+	// example at this point: github.com/tektoncd/pipeline
+	numSlashes := strings.Count(repoURL, "/")
+	if numSlashes < 2 {
+		return "", "", "", errors.New("Url didn't match the requirements (at least two slashes)")
+	}
 	repoURL = strings.TrimSuffix(repoURL, "/")
```
ncskier: Move the line `repoURL = strings.TrimSuffix(repoURL, "/")` above the line `// example at this point: github.com/tektoncd/pipeline`. If you have the url `github.com/tektoncd/`, then it will pass your test of having at least 2 slashes, but will not have the gitRepo. So, you'll want to trim the suffix `/` first.

pkg/endpoints/pipeline.go
ncskier: Please add the `repository` to this error message.

pkg/endpoints/pipeline.go
ncskier: Instead of saying the `repository` parameter is `optional`, please document in the comment that if the repository parameter is an empty string, then `GetAllPipelineRunsImpl` will return all the PipelineRuns in the given namespace.

There are no optional parameters in Go, so this change avoids some confusion and better explains the functionality.

pkg/endpoints/pipeline.go
```diff
@@ -112,44 +139,63 @@ func (r Resource) getPipeline(request *restful.Request, response *restful.Respon
...
 		pipelinerunList, err = pipelineruns.List(metav1.ListOptions{LabelSelector: match})
-		logging.Log.Debugf("+%v", pipelinerunList.Items)
 	} else {
 		pipelinerunList, err = pipelineruns.List(metav1.ListOptions{})
-		logging.Log.Debugf("+%v", pipelinerunList.Items)
 		if err != nil {
-			utils.RespondError(response, err, http.StatusNotFound)
-			return
+			return v1alpha1.PipelineRunList{}, AppError{err, "", http.StatusNotFound}
 		}
```
ncskier: Please move this error check & return block outside of the else statement, because we check for an error getting the list here, but we do not check for an error getting the list with the `match` LabelSelector.

```suggestion
		}
	if err != nil {
		return v1alpha1.PipelineRunList{}, AppError{err, "", http.StatusNotFound}
	}
	return *pipelinerunList, AppError{}
```

pkg/endpoints/pipeline.go
```diff
@@ -166,6 +214,101 @@ func (r Resource) getPipelineRun(request *restful.Request, response *restful.Res
...
+
+	createResponse := r.CreatePipelineRunImpl(pipelineRunData, namespace)
+	if createResponse.CODE == 204 {
+		response.WriteHeader(http.StatusNoContent)
+	} else { // anything other than 204 is an error - RespondError
+		utils.RespondError(response, createResponse.ERROR, createResponse.CODE)
+	}
+}
+
+/*CreatePipelineRunImpl - Create a new manual pipeline run and resources in a given namespace */
```
ncskier: I think you should add some info from this comment `// Assumes you've already applied the yml: so the pipeline definition and its tasks must exist upfront.` about how we are assuming that the Pipeline & Task definitions already exist.

pkg/endpoints/pipeline.go
```diff
@@ -166,6 +214,101 @@ func (r Resource) getPipelineRun(request *restful.Request, response *restful.Res
...
+	pipeline, err := r.getPipelineImpl(pipelineName, namespace)
+	if err != nil {
+		errorMsg := fmt.Sprintf("could not find the pipeline template %s in namespace %s", pipelineName, namespace)
+		logging.Log.Errorf(errorMsg)
+		return &AppError{err, errorMsg, http.StatusPreconditionFailed}
+	}
+
+	logging.Log.Debugf("Found the pipeline template %s OK", pipelineName)
+
+	registryURL := regsitryLocation
```
ncskier: You're just changing the name of the existing variable here. Why not just name it `registryURL` to begin with? 😄

pkg/endpoints/pipeline.go
```diff
@@ -403,8 +555,42 @@ func getDateTimeAsString() string {
 	return strconv.FormatInt(time.Now().Unix(), 10)
 }
 
+// defines and creates a resource of a specifed type and returns a pipeline resource reference for this resource
+// takes a manual pipeline run data struct, namespace for the resource creation, resource name to refer to it and the resource type
```
ncskier: It seems weird that this `createPipelineResource` function requires a `ManualPipelineRun` parameter. It is possible to create Tekton PipelineResources by itself. So, I think it would be clearer to rename the function to indicate that its use is intended for creating PipelineResources for a given `ManualPipelineRun`. For example, maybe a name like `createPipelineResourceForPipelineRun` would be more clear?

pkg/endpoints/pipeline.go
```diff
@@ -403,8 +555,42 @@ func getDateTimeAsString() string {
...
+
+	var paramsForResource []v1alpha1.Param
+	var typeToUse v1alpha1.PipelineResourceType
+	// Unique names are required so timestamp them.
+	if resourceType == v1alpha1.PipelineResourceTypeGit {
+		typeToUse = resourceType
+		paramsForResource = []v1alpha1.Param{{Name: "revision", Value: resourceData.GITCOMMIT}, {Name: "url", Value: resourceData.REPOURL}}
+	} else if resourceType == v1alpha1.PipelineResourceTypeImage {
+		urlToUse := fmt.Sprintf("%s/%s:%s", registryURL, strings.ToLower(resourceData.REPONAME), resourceData.GITCOMMIT)
+		typeToUse = resourceType
```
ncskier: Please remove the lines `typeToUse = resourceType` and `var typeToUse v1alpha1.PipelineResourceType`, because it's just changing the name of the variable `resourceType` to `typeToUse` in both cases of this `if` statement. If the variable should be named `typeToUse`, then change it in the function arguments list.

pkg/endpoints/pipeline.go
```diff
@@ -403,8 +555,42 @@ func getDateTimeAsString() string {
...
+		paramsForResource = []v1alpha1.Param{{Name: "url", Value: urlToUse}}
+	}
+
+	pipelineResource := definePipelineResource(resourceName, namespace, paramsForResource, typeToUse)
+	createdPipelineImageResource, err := r.PipelineClient.TektonV1alpha1().PipelineResources(namespace).Create(pipelineResource)
+	if err != nil {
+		logging.Log.Errorf("could not create pipeline %s resource to be used in the pipeline, error: %s", resourceType, err)
```
ncskier: Please return an error here. The function should return an error if it fails to create the PipelineResource.

pkg/endpoints/pipeline_test.go
```diff
@@ -1145,3 +1085,245 @@ func TestPipelineRunUpdateStatusAlreadySet412(t *testing.T) {
...
+	}
+
+	_, err := r.PipelineClient.TektonV1alpha1().Pipelines("ns1").Create(&pipeline1)
+	if err != nil {
+		t.Errorf("FAIL: error creating test pipeline: %s", err)
+	}
+
+	httpWriter := httptest.NewRecorder()
+
+	badRequestPipelineRunBody := strings.NewReader(`{"fielddoesnotexist": "pipeline",}`)
```
ncskier: Is there supposed to be a comma here? `"pipeline",`

pkg/endpoints/pipeline_test.go
```diff
@@ -1145,3 +1085,245 @@ func TestPipelineRunUpdateStatusAlreadySet412(t *testing.T) {
...
+
+	badRequestPipelineRunRequest := dummyHTTPRequest("POST", "http://wwww.dummy.com:8383/v1/namespaces/ns1/pipelinerun", badRequestPipelineRunBody)
+	badRequestPipelineRunRequestRestful := dummyRestfulRequest(badRequestPipelineRunRequest, "ns1", "")
+	resp := dummyRestfulResponse(httpWriter)
+	r.createPipelineRun(badRequestPipelineRunRequestRestful, resp)
```
ncskier: I'm seeing these lines repeated over and over again in your tests:
```
httpWriter := httptest.NewRecorder()

pipelineRunBody := strings.NewReader(`TEXT`)

pipelineRunRequest := dummyHTTPRequest("POST", "http://wwww.dummy.com:8383/v1/namespaces/ns1/pipelinerun", pipelineRunBody)
pipelineRunRequestRestful := dummyRestfulRequest(pipelineRunRequest, "ns1", "")
resp := dummyRestfulResponse(httpWriter)
r.createPipelineRun(pipelineRunRequestRestful, resp)
```
Please make a function for the repeated code to cut down own code bloat and make the tests easier to read.

pkg/endpoints/pipeline_test.go
```diff
@@ -1145,3 +1085,245 @@ func TestPipelineRunUpdateStatusAlreadySet412(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "Pipeline1",
+		},
+		Spec: v1alpha1.PipelineSpec{},
+	}
+
+	_, err := r.PipelineClient.TektonV1alpha1().Pipelines("ns1").Create(&pipeline1)
+	if err != nil {
+		t.Errorf("FAIL: error creating test pipeline: %s", err)
+	}
```
ncskier: Please make this into a function:
```
pipeline1 := v1alpha1.Pipeline{
	ObjectMeta: metav1.ObjectMeta{
		Name: "Pipeline1",
	},
	Spec: v1alpha1.PipelineSpec{},
}

_, err := r.PipelineClient.TektonV1alpha1().Pipelines("ns1").Create(&pipeline1)
if err != nil {
	t.Errorf("FAIL: error creating test pipeline: %s", err)
}
```
It is repeated almost every time in each test function.

pkg/endpoints/pipeline_test.go
```diff
@@ -1145,3 +1085,245 @@ func TestPipelineRunUpdateStatusAlreadySet412(t *testing.T) {
...
+
+	for _, pipelineResource := range pipelineResourceList.Items {
+		if pipelineResource.Spec.Type == "git" {
+			numberOfGitResources++
+		}
+	}
+
+	if numberOfGitResources != 1 {
+		t.Errorf("FAIL: expected to find a single created PipelineResource of type Git, but found the number of Git PipelineResources to be %d", numberOfGitResources)
+	}
```
ncskier: Instead of just checking the number of Git resources, can you please verify that the PipelineResource created has the same content as what you provided (resource name prefix, gitcommit, repourl).

pkg/endpoints/pipeline_test.go
```diff
@@ -1145,3 +1085,245 @@ func TestPipelineRunUpdateStatusAlreadySet412(t *testing.T) {
...
+
+	for _, pipelineResource := range pipelineResourceList.Items {
+		if pipelineResource.Spec.Type == "image" {
+			numberOfImageResources++
+		}
+	}
+
+	if numberOfImageResources != 1 {
+		t.Errorf("FAIL: expected to find a single created PipelineResource of type image, but found the number of image PipelineResources to be %d", numberOfImageResources)
+	}
```
ncskier: Same as my previous comment about the git PipelineResource. Please verify that the `pipelineImageResource` has the same content as the one you created.

pkg/endpoints/pipeline_test.go
```diff
@@ -1145,3 +1085,245 @@ func TestPipelineRunUpdateStatusAlreadySet412(t *testing.T) {
...
+		}
+	}
+
+	if numberOfGitResources != 1 {
+		t.Errorf("FAIL: expected one created PipelineResource of type git but found the number of git PipelineResources to be %d", numberOfGitResources)
+	}
+
+	if numberOfImageResources != 1 {
+		t.Errorf("FAIL: expected one created PipelineResource of type image but found the number of image PipelineResources to be %d", numberOfImageResources)
+	}
```
ncskier: Same as the above comment about verifying content. You should be able to create functions that you can reuse here.

pkg/endpoints/pipeline_test.go
jessm12: This comma is intended so that the request is registered as a 'bad request'

pkg/endpoints/pipeline.go
```diff
@@ -260,23 +260,35 @@ func (r Resource) CreatePipelineRunImpl(pipelineRunData ManualPipelineRun, names
 	resources := []v1alpha1.PipelineResourceBinding{}
 
 	if pipelineRunData.GITRESOURCENAME != "" {
-		gitResource = r.createPipelineResource(pipelineRunData, namespace, pipelineRunData.GITRESOURCENAME, v1alpha1.PipelineResourceTypeGit)
+		gitResource, err = r.createPipelineResourceForPipelineRun(pipelineRunData, namespace, pipelineRunData.GITRESOURCENAME, v1alpha1.PipelineResourceTypeGit)
+		if err != nil {
+			errorMsg := fmt.Sprintf("Could not create the PipelineResource of type Git with provdided name %s", pipelineRunData.GITRESOURCENAME)
+			logging.Log.Errorf(errorMsg)
```
ncskier: Doesn't need to be `Errorf`, can use `Error`

pkg/endpoints/pipeline.go
```diff
@@ -260,23 +260,35 @@ func (r Resource) CreatePipelineRunImpl(pipelineRunData ManualPipelineRun, names
...
+		}
 		resources = append(resources, v1alpha1.PipelineResourceBinding{Name: pipelineRunData.GITRESOURCENAME, ResourceRef: gitResource})
 	}
 
 	if pipelineRunData.IMAGERESOURCENAME != "" {
-		imageResource = r.createPipelineResource(pipelineRunData, namespace, pipelineRunData.IMAGERESOURCENAME, v1alpha1.PipelineResourceTypeImage)
+		imageResource, err = r.createPipelineResourceForPipelineRun(pipelineRunData, namespace, pipelineRunData.IMAGERESOURCENAME, v1alpha1.PipelineResourceTypeImage)
+		if err != nil {
+			errorMsg := fmt.Sprintf("Could not create the PipelineResource of type Image with provdided name %s", pipelineRunData.IMAGERESOURCENAME)
+			logging.Log.Errorf(errorMsg)
```
ncskier: Doesn't need to be `Errorf`, can use `Error`

pkg/endpoints/pipeline.go
```diff
@@ -292,20 +304,23 @@ func (r Resource) CreatePipelineRunImpl(pipelineRunData ManualPipelineRun, names
 		params = append(params, v1alpha1.Param{Name: "helm-secret", Value: pipelineRunData.HELMSECRET})
 	}
 	// PipelineRun yaml defines references to resources
-	newPipelineRunData := definePipelineRun(generatedPipelineRunName, namespace, serviceAccount, pipelineRunData.REPOURL, pipeline, v1alpha1.PipelineTriggerTypeManual, resources, params)
+	newPipelineRunData, err := definePipelineRun(generatedPipelineRunName, namespace, serviceAccount, pipelineRunData.REPOURL, pipeline, v1alpha1.PipelineTriggerTypeManual, resources, params)
+	if err != nil {
+		logging.Log.Errorf("there was a problem defining the pipeline run: %s", err)
```
ncskier: Should we fail/return an error here? If the PipelineRunData failed to be created, then there's no reason to proceed.

install/tekton-dashboard-deployment.yaml
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: tekton-dashboard
-        image: "CHANGE_ME/back-end:latest"
+        image: "jess64/devops-back-end:latest"
```
ncskier: Oops, I think this needs to be changed back too.

vtereso:
As with credentials, we are creating custom structs to encapsulate data, but in the end, we pass everything into the tekton CRD (see https://github.com/tektoncd/dashboard/pull/33/files#diff-2a268056787157d73c4168498850de86R289). Why not just accept the CRDs from the client?

ncskier:
@vtereso Regarding your comment about the tekton CRDs, it is because this issue is focused on a specific type of PipelineRun (see https://github.com/tektoncd/dashboard/pull/33#issuecomment-482502528).

I agree with what you're saying about accepting the tekton CRDs directly from the client, and that is being discussed in this issue: https://github.com/tektoncd/dashboard/issues/47 about generalizing the PipelineRun creation. Please take a look at the issue, and let us know what you think. 🙂

jessm12:
Yeah and I do agree also, this issue and PR was focusing on supporting the Pipelines that are in the sample Pipelines repo and a lot of it came from our original implementations a while ago, we have got an issue open for a more generalised version of creating a PipelineRun which should replace this code in the not too distant future they do I do think a lot of this code could be reused for that - Brandon has linked the issue above 😄

jessm12:
@ncskier this should be good for a final review now please and approve if you're happy! I've merged the latest code and fixed up all the tests 😄

PR #34, opened 2019-04-11
Extension support

akihikokuroda:
Extension support
The dashboard adds the service that has the following items as the back-end extension.
```
  annotations:
    tekton-dashboard-endpoints: /
  labels:
    tekton-dashboard-extension: "true"
```


akihikokuroda:
I signed it!

cmd/dashboard/main.go
```diff
@@ -75,6 +75,10 @@ func main() {
 	resource.RegisterHealthProbes(wsContainer)
 	resource.RegisterReadinessProbes(wsContainer)
 
+	logging.Log.Info("Search extensions")
```
a-roberts: I think `logging.Log.Infof("Searching for extensions in the namespace %s", installedNamespace)` would be more useful here

pkg/endpoints/utils.go
```diff
@@ -107,3 +112,45 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
 
 	container.Add(wsv4)
 }
+
+// Store all types here that are reused throughout files
+type Extension struct {
+	Name string
+	Url  string
```
a-roberts: URL to make the go linter happy

The comment for this struct is too generic for me, let's include information about the Extension itself?

pkg/endpoints/utils.go
```diff
@@ -107,3 +112,45 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+	Port string
+}
+
+
+func (r Resource) RegisterExtension(container *restful.Container, namespace string) {
```
a-roberts: Missing a method comment

pkg/endpoints/utils.go
```diff
@@ -107,3 +112,45 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+	logging.Log.Info("Adding API for extensions")
+	svcs, err := r.K8sClient.CoreV1().Services(namespace).List(metav1.ListOptions{LabelSelector: "tekton-dashboard-extension=true"})
```
a-roberts: As this is a super important string (`tekton-dashboard-extension`) let's get it declared upfront as a const somewhere easy to spot, and well documented

pkg/endpoints/utils.go
```diff
@@ -107,3 +112,45 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+	if err != nil {
+		logging.Log.Debug("no extension found")
```
a-roberts: What if there was actually an error? Why not errorf?

pkg/endpoints/utils.go
```diff
@@ -107,3 +112,45 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+		return
+	}
+	ws := new(restful.WebService)
+	ws.
+		Path("/").
+		Consumes(restful.MIME_JSON).
+		Produces(restful.MIME_JSON)
+	for _, svc := range svcs.Items {
```
a-roberts: Big nit but let's newline this (I think just on saving in vscode you'd get it anyway)

pkg/endpoints/utils.go
```diff
@@ -107,3 +112,45 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+		for key, url := range svc.ObjectMeta.Annotations{
+			if key == "tekton-dashboard-endpoints" {
```
a-roberts: Let's reference this as a string we're gonna declare based on a previous comment

pkg/endpoints/utils.go
```diff
@@ -107,3 +112,45 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+				logging.Log.Debugf("extension Url: %s", url)
```
a-roberts: Extension URL: ...

pkg/endpoints/utils.go
```diff
@@ -107,3 +112,45 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+                                ext := Extension { Name: svc.ObjectMeta.Name, Url: url, Port: getPort(svc) }
+				ws.Route(ws.POST(url).To(ext.HandleExtension))
+			}
+		}
+	}
+	container.Add(ws)
+}
+
+func (ext Extension) HandleExtension(request *restful.Request, response *restful.Response) {
+	target, _ := url.Parse("http://" +  ext.Name + ":" + ext.Port + "/")
```
a-roberts: Do we need a todo here for handling https eventually? 

Missing method comment

pkg/endpoints/utils.go
```diff
@@ -107,3 +112,45 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+	proxy := httputil.NewSingleHostReverseProxy(target)
+	proxy.ServeHTTP(response, request.Request)
+}
+
+func getPort(svc corev1.Service) string {
```
a-roberts: Missing method comment

akihikokuroda:
I addressed all review comments.

pkg/endpoints/router.go
```diff
@@ -99,3 +117,50 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+	ws.
+		Path("/").
+		Consumes(restful.MIME_JSON).
+		Produces(restful.MIME_JSON)
+
+	for _, svc := range svcs.Items {
+		for key, url := range svc.ObjectMeta.Annotations{
+			if key == urlKey {
+				logging.Log.Debugf("extension URL: %s", url)
+                                ext := Extension { Name: svc.ObjectMeta.Name, URL: url, Port: getPort(svc) }
```
a-roberts: Massive whitespace here we don't need 😱

pkg/endpoints/router.go
```diff
@@ -99,3 +117,50 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+				ws.Route(ws.POST(url).To(ext.HandleExtension))
+			}
+		}
+	}
+	container.Add(ws)
+}
+
+// HandleExtension - this route request to the extention path to the extention service
```
a-roberts: routes*

Hmm, sentence is a bit weird - so this routes a request to the extension path to the extension service? Is it going to the extension service via the path, really?

pkg/endpoints/router.go
```diff
@@ -99,3 +117,50 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+func (ext Extension) HandleExtension(request *restful.Request, response *restful.Response) {
+	target, _ := url.Parse("http://" +  ext.Name + ":" + ext.Port + "/")
+	proxy := httputil.NewSingleHostReverseProxy(target)
+	proxy.ServeHTTP(response, request.Request)
+}
+
+// getPort - this get the port of the service
```
a-roberts: gets*

pkg/endpoints/router.go
```diff
@@ -99,3 +117,50 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
 
 	container.Add(wsv4)
 }
+
+// Back-end extension: Requests to the URL are passthrough to the Port of the Name service (extension)
```
a-roberts: passed through*

pkg/endpoints/router.go
```diff
@@ -99,3 +117,50 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+// "label: tekton-dashboard-extension=true" in the service defines the extention
```
a-roberts: extension*

pkg/endpoints/router.go
```diff
@@ -99,3 +117,50 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+// "annotation: tekton-dashboard-endpoints=<URL>" spacifies the path for the extension
```
a-roberts: specifies*

a-roberts:
@akihikokuroda shall we also add something into our DEVELOPMENT.md?

akihikokuroda:
Yes, I'll do that.

a-roberts:
@AlanGreene can you please take a look and see if this would offer everything you want? Specifically for an extension we will offer you the name, URL and port as JSON

AlanGreene:
I may be missing something obvious (very new to Go) but I don't see how the frontend would:
- retrieve a list of registered extensions
- retrieve a JS bundle or other static resources provided by the extension
  - the code appears to proxy POST requests but would also need to support GET
  - how will we handle caching of the static resources?
    - standard URL for extension JS bundle? Would need some mechanism to handle cache busting
    - include content-based hash in JS bundle name (approach I've taken in my sample) to allow for long-term caching? Then we need to provide the correct name / URL to the frontend

We also need to provide a display name appropriate for use in the UI for navigation / page headers etc. unless the extension name can be used for this.

How do we handle potential clashes of extension name? Is this something we're likely to encounter?

I'm also not entirely clear on how the frontend would address an extension via the proxy. What endpoint is the backend exposing that the frontend will consume?

akihikokuroda:
I can add the REST API that returns the list of extensions.  The (kubernetes) service name should be unique in the namespace so we can return it with the path for the extension.
I'll add the GET handling as well as PUT and DELETE.

a-roberts:
We discussed this today on a call together hosted by @dibbles, so in the interest of transparency here are the notes I've taken.

For creating extensions we should also be including:

- A display name to use (add to this PR)
- the URL or path to js bundle is where all of the paths go so we needn't care from the backend (add to this PR)
- We can use the current Name field as the extension path e.g. at /extension#*Name*

Extensions with their own breadcrumbs can be handled on the frontend.

Aki can add a GET method that return a list extensions, any particular format you want @AlanGreene  (so in a similar way we've arranged taskrun logs)?

@AlanGreene can you confirm this sounds about right to you? Note that currently we're doing the discovery at startup time as opposed to using an informer (services having certain labels).

akihikokuroda:
I added the GET extensions REST API.  It returns the display name and the bundle location from the service annotations as well as the path and the service name of the extension.  I also added the go unit test for the extension registration.

pkg/endpoints/router.go
```diff
@@ -98,6 +113,72 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+			extensions = append(extensions, ext)
+		}
+	}
+	logging.Log.Debugf("extension: %+v", extensions)
+	container.Add(ws)
+}
+
+// HandleExtension - this routes request to the extension service
+func (ext Extension) HandleExtension(request *restful.Request, response *restful.Response) {
+	target, _ := url.Parse("http://" +  ext.Name + ":" + ext.Port + "/")
```
dibbles: do we really want to ignore logging any returned error here - I appreciate there shouldn't be one but.....?

pkg/endpoints/router.go
```diff
@@ -98,6 +113,72 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+	proxy := httputil.NewSingleHostReverseProxy(target)
+	proxy.ServeHTTP(response, request.Request)
+}
+
+/* Get all extensions in a given namespace */
+func (r Resource) getAllExtensions(request *restful.Request, response *restful.Response) {
+	namespace := request.PathParameter("namespace")
```
dibbles: per previous comments ..... do we really want namespace?  The only extensions processed have been in the namespace to which we were installed.

pkg/endpoints/router.go
```diff
@@ -98,6 +113,72 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+	logging.Log.Debugf("In getAllExtensions: namespace: %s", namespace)
+	logging.Log.Debugf("extension: %+v", extensions)
+
```
dibbles: General note for all changes ..... inconsistency in messages.

Debug/Info should probably start with a capital - error messages should be all lower case

dibbles: Is it at all possible that the UI might be trying to use this REST endpoint to read the extensions while extensions are still being processed by main.go? .... if so protection required - don't think this is currently an issue as we don't serve content from main.go until after processing extensions has completed, but if this ever changed to be an informer rather than a static listing protection would be needed.

pkg/endpoints/router.go
```diff
@@ -98,6 +113,72 @@ func (r Resource) RegisterReadinessProbes(container *restful.Container) {
...
+ 				DisplayName: svc.ObjectMeta.Annotations[displayNameKey],
+				BundleLocation: svc.ObjectMeta.Annotations[bundleLocationKey]}
+			// extension handler is registered at the url
+			ws.Route(ws.GET(url).To(ext.HandleExtension))
+			ws.Route(ws.POST(url).To(ext.HandleExtension))
+			ws.Route(ws.PUT(url).To(ext.HandleExtension))
+			ws.Route(ws.DELETE(url).To(ext.HandleExtension))
+			extensions = append(extensions, ext)
+		}
+	}
```
dibbles: DO we need code to check the annotations have been set?  All we know at this point is that we have a service with the required label ..... not that it has the annotations as required, with values

akihikokuroda: These 2 annotations must be optional for some extensions that only extends REST API.

pkg/endpoints/router.go
akihikokuroda: I add the check.

cmd/dashboard/main.go
```diff
@@ -75,6 +75,10 @@ func main() {
 	resource.RegisterHealthProbes(wsContainer)
 	resource.RegisterReadinessProbes(wsContainer)
 
+	installedNamespace := os.Getenv("INSTALLED_NAMESPACE")
```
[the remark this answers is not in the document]
akihikokuroda: I take out the namespace from the API URL.  I'll update the yaml file.

pkg/endpoints/router.go
a-roberts: Another comment here is that if someone comes along and installs an extension after the fact (a new labelled service), the client side UI, presumably, is going to try and register that because your get endpoint would render it

We should persistently store (e.g. in a configmap) the extension at install time *or* use an informer and then users just need to restart their browser to pick it up (until we offer it over a socket).

a-roberts: Also bear in mind annotation spec at https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/ doesn't let us spaces, we could client side swap certain chars for spaces if we were desperate...

akihikokuroda: For now (My current issue requirement understanding), the extensions are registered only at the pod start up.  
Thanks for the annotation spec.  I didn't know the restrictions.

akihikokuroda:
It's ready now.

pkg/endpoints/router.go
```diff
@@ -154,22 +155,34 @@ func (r Resource) RegisterExtension(container *restful.Container, namespace stri
...
 		}
 	}
-	logging.Log.Debugf("extension: %+v", extensions)
+	logging.Log.Debugf("Extension: %+v", extensions)
 	container.Add(ws)
+
+	logging.Log.Info("Adding API for extension")
+	wsv1 := new(restful.WebService)
+	wsv1.
+		Path("/v1/extension")
```
dibbles: minor point "/v1/extensions" might be nicer than  "/v1/extension" - also would need to change DEVELOPMENT.md to match and@AlanGreene would need to match endpoint in ui code.

config/tekton-dashboard-deployment.yaml
```diff
@@ -32,3 +32,7 @@ spec:
         env:
         - name: PORT
           value: "9097"
+        - name: INSTALL_NAMESPACE
```
AlanGreene: should this be `INSTALLED_NAMESPACE` to match https://github.com/tektoncd/dashboard/pull/34/files#diff-51aee488c67e4d9a10cd51b3c5c22e59R78 ?

akihikokuroda: This value is set automatically so it doesn’t need to be set manually.

akihikokuroda:
I made the api /ve/extensions.

a-roberts:
All tests pass, let's get it in, thank you @akihikokuroda!

/lgtm

config/tekton-dashboard-deployment.yaml
AlanGreene: If this is redundant it should be removed to avoid confusion. This is defining an environment variable `INSTALL_NAMESPACE` that's not actually used anywhere and could be easily confused for the `INSTALLED_NAMESPACE` we're expecting?

a-roberts: Yeah, I really don't think it is redundant, PR here. https://github.com/tektoncd/dashboard/pull/77

@akihikokuroda can you please take a look - in the description I've mentioned what I think, lemme know if I'm mistaken please!

akihikokuroda: My bad.  Yes, you are right.

PR #36, opened 2019-04-11
correct image name

akihikokuroda:
correct image name
The back-end is missing in the tekton-dashboard-deployment.yaml file.


akihikokuroda:
I signed it!

Issue #42, opened 2019-04-12
Document APIs using Swagger / OpenAPI spec


AlanGreene:
This would allow us to provide richer documentation in a standard format, as well as interactive examples that would ease development.

Open API spec: https://swagger.io/specification/
Swagger UI example: https://petstore.swagger.io/

Something like:
```
openapi: 3.0.0
info:
  title: Tekton Dashboard
  description: Optional multiline or single-line description in [CommonMark](http://commonmark.org/help/) or HTML.
  version: 0.0.1
servers:
  - url: '{protocol}://{host}:{port}/v1/namespaces/{namespace}'
    description: Optional server description, e.g. Main (production) server
    variables:
      protocol:
        enum:
          - http
          - https
        default: http
      host:
        default: localhost
      port:
        default: 9097
      namespace:
        default: tekton-pipelines
        description: namespace where the Tekton resources (Pipelines, Tasks, etc.) are installed
paths:
  /pipeline/:
    get:
      summary: Returns a list of Pipelines
      description: Optional extended description in CommonMark or HTML.
      parameters:
        - in: path
          name: namespace
          schema:
            type: string
            example: tekton-pipelines
          required: true
          description: namespace in which the Pipelines are installed
      responses:
        '200':
          description: A JSON array of Pipelines
          content:
            application/json:
              schema: 
                type: array
                items: 
                  type: string
  /pipeline/{name}:
    get:
      summary: Returns a sinlge Pipeline
      description: Optional extended description in CommonMark or HTML.
      parameters:
        - in: path
          name: name
          schema:
            type: string
            example: demo-pipeline
          required: true
          description: name of the Pipeline to retrieve
      responses:
        '200':
          description: A JSON object representing the requested Pipeline
          content:
            application/json:
              schema: 
                type: object
```

will produce:
![image](https://user-images.githubusercontent.com/2829095/56046699-bdc88700-5d3b-11e9-9756-ba18f13c8260.png)

![image](https://user-images.githubusercontent.com/2829095/56046718-c7ea8580-5d3b-11e9-8704-751c9c969fb4.png)

akihikokuroda:
PR: https://github.com/tektoncd/experimental/pull/21

akihikokuroda:
PR in dashboard repo: https://github.com/tektoncd/dashboard/pull/126

AlanGreene:
Closing this since we've removed most of our custom APIs in favour of using kube APIs directly.

Issue #41, opened 2019-04-12
Document API definitions


a-roberts:
Now that we have DEVELOPMENT.md we should document our APIs. We'll want return codes, paths, any errors that can happen, and bonus points for any json snippets!

a-roberts:
/assign

I'll start on this, I have a bunch already written up from before we open sourced that I'll push as an initial commit to my branch and then let's look at making it awesome.

AlanGreene:
In line with 'making it awesome' I think it would be useful to document the APIs as a Swagger / OpenAPI spec, new issue to track that https://github.com/tektoncd/dashboard/issues/42

a-roberts:
Changes are now merged, closing 😄

PR #43, opened 2019-04-12
Document our backend API definitions

a-roberts:
Here's what I have so far (I'm done with work for this week and want to share some progress). This is for https://github.com/tektoncd/dashboard/issues/41. 

As @AlanGreene suggested [here](https://github.com/tektoncd/dashboard/issues/42) it would be awesome to have a swagger/openapi spec. It would be very cool to autogenerate these using swagger and https://github.com/go-swagger/go-swagger looks to be suitable.

I'd like to flesh this out by adding **all** response codes next for each API.


a-roberts:
@jessm12 please take a look 😄

Issue #12, opened 2019-04-10
Expose tasks as a top level construct


steveodonovan:
Expose tasks as a top level construct with navigation to tasks and task runs.

New root navigation

![Screen Shot 2019-04-13 at 15 40 49](https://user-images.githubusercontent.com/18128367/56081297-950bc480-5e03-11e9-80b5-2a00d08bca8e.png)

Tasks view. This show all tasks.

![Screen Shot 2019-04-13 at 15 41 00](https://user-images.githubusercontent.com/18128367/56081237-eebfbf00-5e02-11e9-9837-70d136e44ea8.png)

Task Runs for the task selected in the previous navigation with no selected task run.

![Screen Shot 2019-04-13 at 15 41 18](https://user-images.githubusercontent.com/18128367/56081240-fbdcae00-5e02-11e9-9547-d7a4f3cb9357.png)

Selected task no selected step. Header updates with task run name and task run time for the given selected task

![Screen Shot 2019-04-13 at 15 41 26](https://user-images.githubusercontent.com/18128367/56081245-0b5bf700-5e03-11e9-8cbd-b632ad928b62.png)

Task and step selection. Similar to pipeline run navigation

![Screen Shot 2019-04-13 at 15 41 34](https://user-images.githubusercontent.com/18128367/56081253-1f9ff400-5e03-11e9-9ee7-fbe6fdbd74d0.png)

Notification where the selected task has never run.

![Screen Shot 2019-04-13 at 15 42 22](https://user-images.githubusercontent.com/18128367/56081271-4ceca200-5e03-11e9-8f92-8557c1ce1768.png)

@skaegi for review on the above. Alan and Myself iterated on it Thursday and Friday and settled on this layout. It Does add some of the navigation outlined in the design tickets.

PR #49, opened 2019-04-13
Add tasks as a top level navigation

steveodonovan:
@AlanGreene A little wide reaching and large here. Can break this down if you like. The overall design is also still under review in https://github.com/tektoncd/dashboard/issues/12

There are a few pieces here which will be a little cleaner when we have the store in place.


src/containers/Home/Home.js
```diff
@@ -0,0 +1,67 @@
...
+        <StructuredListWrapper border selection>
+          <StructuredListHead>
+            <StructuredListRow head>
+              <StructuredListCell head>Task</StructuredListCell>
+            </StructuredListRow>
+          </StructuredListHead>
+          <StructuredListBody>
+            <StructuredListRow className="pipeline" key="pipeline">
+              <StructuredListCell>
+                <Link to="/pipelines">pipelines</Link>
```
AlanGreene: link text: Pipelines, similarly Tasks below

src/containers/Home/Home.js
```diff
@@ -0,0 +1,67 @@
...
+
+      <main>
+        <StructuredListWrapper border selection>
+          <StructuredListHead>
+            <StructuredListRow head>
+              <StructuredListCell head>Task</StructuredListCell>
+            </StructuredListRow>
+          </StructuredListHead>
+          <StructuredListBody>
+            <StructuredListRow className="pipeline" key="pipeline">
```
AlanGreene: `key` not necessary here

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -17,12 +17,17 @@ import { InlineNotification } from 'carbon-components-react';
...
 
-import PipelineRunHeader from '../../components/PipelineRunHeader';
+import RunHeader from '../../components/RunHeader';
 import StepDetails from '../../components/StepDetails';
 import TaskTree from '../../components/TaskTree';
-import { getStatus } from '../../utils';
+import {
+  getStatus,
+  selectedTask,
+  taskRunStep,
```
AlanGreene: nit: order imports alphabetically for easier scanning

src/containers/TaskRuns/TaskRuns.js
```diff
@@ -0,0 +1,213 @@
...
+    let notification;
+    const tasks = await getTasks();
+    const task = tasks.find(
+      currentTask => currentTask.metadata.name === selectedTaskName
+    );
+    let taskRuns = await getTaskRuns();
+    taskRuns = taskRuns
+      .filter(
+        taskRun =>
+          taskRun.spec.taskRef && taskRun.spec.taskRef.name === selectedTaskName
```
AlanGreene: TODO: discuss additional back end API to retrieve TaskRuns for given Task name.
Not required for this initial PR but should follow up separately.

src/containers/TaskRuns/TaskRuns.js
```diff
@@ -0,0 +1,213 @@
...
+        id: step.name
+      };
+    });
+    return steps;
+  }
+
+  // once redux store is available errors will be handled properly with dedicated components
+  notification(notification) {
+    const { kind, message } = notification;
+    const titles = {};
```
AlanGreene:
```
const titles = {
  info: '...',
  error: '...'
};
```

TODO: 
Also might re-word the info text. 'Task runs not available' could sound like an error.
There's a similar message above: https://github.com/tektoncd/dashboard/pull/49/files#diff-ccd95d56ab624312e691721ff80a1260R101 'Task has never run'.

Let's revisit this as part of the design update.

src/utils/index.js
```diff
@@ -35,3 +35,31 @@ export function getPipelineRunStatusIcon({ reason, status }) {
...
+    return {};
+  }
+
+  const { id, stepName, stepStatus, status, reason, ...definition } = step;
+
+  return {
+    definition,
+    reason,
+    stepName,
+    stepStatus,
```
AlanGreene: What's the difference between `status` and `stepStatus` here?

steveodonovan: Step status is actually the value for the step in the task run (
```
    {
      terminated: {
        exitCode: 1,
        'Error',
        startedAt: '2019',
        finishedAt: '2019',
        containerID: 'containerd://testid'
      }
    }
```
for example

status is the actual succeeded/failed/waiting ...

Bad naming for sure but preexisting. Changing it in this PR might mean hitting other components. Mind if I address it separately ?

AlanGreene: ah yes, ok let's follow up before this gets forgotten 👍

steveodonovan:
@AlanGreene Could I get another once over please ? Thanks

AlanGreene:
The actual change isn't as big as the diff suggests, ~50% tests and refactoring by moving / extraction.

Verified the changes locally, including at smaller screen sizes. It's consistent with the existing design and layout 👍 Core components are now more easily reusable too.

/approve

Issue #44, opened 2019-04-12
Investigate ko for building


a-roberts:
While adding https://github.com/tektoncd/dashboard/pull/27 @vdemeester asked a question in a review comment:  `Out of curiosity, what doesn't work with ko ?`, and I've honestly not tried using `ko` before with this project.

Prove we can use `ko` to build and push and update DEVELOPMENT.md if so.

For tektoncd/pipeline I see lots of kodata folders (e.g. [here](https://github.com/tektoncd/pipeline/tree/master/cmd/controller)).

It would be very cool to recommend doing a `ko apply` and it'll just build and deploy our containers; at the least for now we can make sure it builds and deploys the backend using what's at `install`.

Useful resources:
- #ko channel at https://knative.slack.com
- https://github.com/google/ko

We've had success building and pushing knative repositories and tektoncd/pipeline both locally and to public Dockerhub, including updating the readme for tektoncd/pipeline to use a local registry. It's likely we can be more inline with what's done for tektoncd/pipeline and if so let's advocate that.

skaegi:
Let's make this a priority as it lets other devs try the dashboard out easily. We also want to make it very easy to grab the latest pipeline and mix with the latest dashboard.

I'm working on #14 and need something to mix the front-end with the back-end and will take a look at using ko but feel free to assign to yourself and do this just for the back-end if you're ready.

a-roberts:
Yeah, I'm gonna look at this now (I've been working on a compatibility tester using Travis and the runs take *ages*). https://triggermesh.com/2019/03/07/go-binaries-kubernetes-ko/ this actually looks really easy: notice the gopath style image location.

/assign

a-roberts:
I was stumped as I didn't realise the folder needed to be called `config`: we were using `config` to store frontend config files!

I've got `ko apply -f config` working fine now, pushing to my Dockerhub, so I'm updating our docs now and the webpack const definition to read from config_frontend instead of config for the json.

a-roberts:
Merged a couple days ago, closing 🎉

PR #55, opened 2019-04-16
Advocate using ko in developer docs

a-roberts:
This is for https://github.com/tektoncd/dashboard/issues/44

I want us to be consistent with how tektoncd/pipeline is built and worked on so please do take a look, poking @vdemeester @jessm12 @dibbles and @vtereso 👼 , lemme know if there's something I've missed or messed up!


DEVELOPMENT.md
```diff
@@ -54,6 +56,44 @@ _Adding the `upstream` remote sets you up nicely for regularly
 - We've had good success using Docker Desktop: ensure your Kubernetes cluster is healthy and you have plenty of disk space allocated as PVs will be created for PipelineRuns.
 - Ensure you can push images to a Docker registry - the above listed requirements are only for local development, otherwise we pull in the tooling for you in the image.
 
+1. `GOPATH`: If you don't have one, simply pick a directory and add
+   `export GOPATH=...`
+1. `$GOPATH/bin` on `PATH`: This is so that tooling installed via `go get` will
+   work properly.
+1. `KO_DOCKER_REPO`: The docker repository to which developer images should be
```
vdemeester: There is also the `ko.local` hack (if you use `minikube` or docker-for-mac) as it doesn't require you to push to a registry, it directly load the images on the k8s docker daemon.
Can be added in a follow-up though :wink:

a-roberts: Mb, I only pushed the docs up on my first commit, addressed now

Good point, I remember using `ko.local` and advocated for it when building tektoncd/pipeline, I'm happy for that to be done as a follow-up also (we personally stick to using dockerhub so we needn't spin up and manage our own registry)

webpack.dev.js
```diff
@@ -1,7 +1,7 @@
 const merge = require('webpack-merge');
 
 const common = require('./webpack.common.js');
-const { API_DOMAIN, PORT } = require('./config/config.json');
+const { API_DOMAIN, PORT } = require('./config-frontend/config.json');
```
vdemeester: Isn't it `config_frontend` ? (looking at the folder)

a-roberts:
@AlanGreene FYI as I've renamed the config folder, lemme know if this is cool with you too from a frontend perspective and then I'll merge this

a-roberts:
@AlanGreene thanks, updated and noticed a ref in DEVELOPMENT.md too, lemme know if I've missed any others please, I've done an `npm install` and an `npm test` too

AlanGreene:
Verified the coverage reports and running the client locally too. Looks good 👍

PR #57, opened 2019-04-16
Add query parameters to filter pipelineruns/taskruns by name

vtereso:
From the below snippet of code, being able to grab all such Run objects based on their reference is likely a desirable filter. This PR adds query filtering to both pipelineruns and taskruns and adds the corresponding tests .
```
type PipelineRun struct {
	metav1.TypeMeta `json:",inline"`
	// +optional
	metav1.ObjectMeta `json:"metadata,omitempty"`

	// +optional
	Spec PipelineRunSpec `json:"spec,omitempty"`
	// +optional
	Status PipelineRunStatus `json:"status,omitempty"`
}
type PipelineRunSpec struct {
	PipelineRef PipelineRef     `json:"pipelineRef"`
...
}
type PipelineResourceRef struct {
	// Name of the referent; More info: http://kubernetes.io/docs/user-guide/identifiers#names
	Name string `json:"name"`
	// API version of the referent
	// +optional
	APIVersion string `json:"apiVersion,omitempty"`
}
```

### DIFF
--- a/pkg/endpoints/pipeline.go
+++ b/pkg/endpoints/pipeline.go
@@ -138,6 +138,7 @@ func (r Resource) getAllPipelineRuns(request *restful.Request, response *restful
 	pipelinerunInterface := r.PipelineClient.TektonV1alpha1().PipelineRuns(namespace)
 	var pipelinerunList *v1alpha1.PipelineRunList
 	var labelSelector string // key1=value1,key2=value2, ...
+	var fieldSelector string // metadata.something=something, ...
 	var err error
 
 	// repository query filter
@@ -150,20 +151,15 @@ func (r Resource) getAllPipelineRuns(request *restful.Request, response *restful
 		}
 		labelSelector = strings.Join(labels,",")
 	}
-	pipelinerunList, err = pipelinerunInterface.List(metav1.ListOptions{LabelSelector: labelSelector})
+	// name query filter
+	if name != "" {
+		fieldSelector = "metadata.name=" + name
+	}
+	pipelinerunList, err = pipelinerunInterface.List(metav1.ListOptions{LabelSelector: labelSelector,FieldSelector: fieldSelector})
 	logging.Log.Debugf("+%v", pipelinerunList.Items)
 	if err != nil {
 		utils.RespondError(response, err, http.StatusNotFound)
 		return
-	}
-	// name query filter
-	if name != "" {
-		for i := range pipelinerunList.Items {
-			pipelinerun := pipelinerunList.Items[i]
-			if pipelinerun.ObjectMeta.Name != name {
-				pipelinerunList.Items = append(pipelinerunList.Items[:i], pipelinerunList.Items[i+1:]...)
-			}
-		}
 	}
 	response.WriteEntity(pipelinerunList)
 }
@@ -234,19 +230,16 @@ func (r Resource) getAllTaskRuns(request *restful.Request, response *restful.Res
 	logging.Log.Debugf("In getAllTaskRuns, namespace: `%s`, name query: `%s`", namespace, name)
 
 	taskrunInterface := r.PipelineClient.TektonV1alpha1().TaskRuns(namespace)
-	taskrunList, err := taskrunInterface.List(metav1.ListOptions{})
+
+	var fieldSelector string // metadata.something=something, ...
+	// name query filter
+	if name != "" {
+		fieldSelector = "metadata.name=" + name
+	}
+	taskrunList, err := taskrunInterface.List(metav1.ListOptions{FieldSelector: fieldSelector})
 	if err != nil {
 		utils.RespondError(response, err, http.StatusNotFound)
 		return
-	}
-	// name query filter
-	if name != "" {
-		for i := range taskrunList.Items {
-			taskrun := taskrunList.Items[i]
-			if taskrun.ObjectMeta.Name != name {
-				taskrunList.Items = append(taskrunList.Items[:i],taskrunList.Items[i+1:]...)
-			}
-		}
 	}
 	response.WriteEntity(taskrunList)
 }
--- a/pkg/endpoints/pipeline.go
+++ b/pkg/endpoints/pipeline.go
@@ -126,27 +126,43 @@ func (r Resource) getPipelineImpl(name, namespace string) (v1alpha1.Pipeline, er
 	return *pipeline, nil
 }
 
-/* Get all pipeline runs in a given namespace */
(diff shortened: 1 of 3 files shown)


pkg/endpoints/pipeline.go
```diff
@@ -126,27 +126,43 @@ func (r Resource) getPipelineImpl(name, namespace string) (v1alpha1.Pipeline, er
...
-/* Get all pipeline runs in a given namespace */
+/* Get all pipeline runs in a given namespace, filters based on query parameters */
 func (r Resource) getAllPipelineRuns(request *restful.Request, response *restful.Response) {
+	// Request Params
 	namespace := request.PathParameter("namespace")
-	pipelineruns := r.PipelineClient.TektonV1alpha1().PipelineRuns(namespace)
+	// Query Params
 	repository := request.QueryParameter("repository")
+	name := request.QueryParameter("name")
+	logging.Log.Debugf("In getAllPipelineRuns: namespace: `%s`, repository query: `%s`, repository query: `%s`", namespace, repository, name)
```
a-roberts: Repo query twice instead of name query (only on the header part and not the variable used)

pkg/endpoints/pipeline.go
```diff
@@ -126,27 +126,43 @@ func (r Resource) getPipelineImpl(name, namespace string) (v1alpha1.Pipeline, er
...
+	if err != nil {
+		utils.RespondError(response, err, http.StatusNotFound)
+		return
+	}
+	// name query filter
+	if name != "" {
+		for i := range pipelinerunList.Items {
+			pipelinerun := pipelinerunList.Items[i]
+			if pipelinerun.ObjectMeta.Name != name {
+				pipelinerunList.Items = append(pipelinerunList.Items[:i], pipelinerunList.Items[i+1:]...)
```
a-roberts: Could be field selector?

pkg/endpoints/pipeline.go
```diff
@@ -209,18 +225,30 @@ func (r Resource) getTask(request *restful.Request, response *restful.Response)
...
 		utils.RespondError(response, err, http.StatusNotFound)
 		return
 	}
-	response.WriteEntity(taskrunlist)
+	// name query filter
+	if name != "" {
+		for i := range taskrunList.Items {
+			taskrun := taskrunList.Items[i]
+			if taskrun.ObjectMeta.Name != name {
+				taskrunList.Items = append(taskrunList.Items[:i],taskrunList.Items[i+1:]...)
```
a-roberts: fieldSelector again perhaps?

pkg/endpoints/pipeline.go
vtereso: Had misunderstanding with field selector, realized this could be done. I implemented this, but it is not supported on the gotest side. I have been looking into contributing some changes on the fake testing side so that we can do this.

pkg/endpoints/pipeline.go
vtereso: Same as above.

pkg/endpoints/pipeline.go
vtereso: What is the name is specified and not the repository?

a-roberts: We could build up a log string with fmt.Sprintf based on what's being provided and add a bunch of ifs, yeah having just the name is perfectly valid, so is just the repository, so is both

pkg/endpoints/pipeline.go
a-roberts: What part isn't supported on the gotest sides, just using field selector?

If it works for our actual implementation code we should still use it here, and if it doesn't for tests I'm sure we can workaround it (carefully adding/removing runs and checking counts for example, or just stick to the for loop in your test code)

vtereso:
Both instances that we provide list options do not function correctly when executed from our gotests: 
- https://github.com/tektoncd/dashboard/pull/57/files#diff-2a268056787157d73c4168498850de86R239
- https://github.com/tektoncd/dashboard/pull/57/files#diff-2a268056787157d73c4168498850de86R158
The fakeClient does not support filtering, so the previous method of post filtering was fine for these tests, but was inefficient. 
I opened a PR for the fix to this: https://github.com/kubernetes/kubernetes/pull/76735

vtereso:
I can remove the tests in the mean time or we can let this sit, but I don't expect this change to be accepted/shipped any time soon.

a-roberts:
@vtereso keep the changes as is for now but remove the "Request Params" comment in favour of "Path params" like you do elsewhere in this commit and fix the `logging.Log.Debugf("In getAllPipelineRuns: namespace: `%s`, repository query: `%s`, repository query: `%s`", namespace, repository, name)` line please, then let's get this merged

pkg/endpoints/pipeline.go
```diff
@@ -209,18 +227,29 @@ func (r Resource) getTask(request *restful.Request, response *restful.Response)
...
+	}
+	logging.Log.Debugf("In getAllTaskRuns, namespace: %s, parameters: %s", namespace, strings.Join(queryParams,","))
 
-	taskrunlist, err := taskruns.List(metav1.ListOptions{})
+	taskrunInterface := r.PipelineClient.TektonV1alpha1().TaskRuns(namespace)
+
+	var fieldSelector string // metadata.something=something, ...
+	// name query filter
+	if name != "" {
+		fieldSelector = "metadata.name=" + name
```
a-roberts: Can do in the above if on name being set

pkg/endpoints/pipeline_test.go
```diff
@@ -255,6 +255,28 @@ func TestTaskRun(t *testing.T) {
 		t.Errorf("Task2 is not returned: %s, %s", result.Items[0].Name, result.Items[1].Name)
 	}
 
+	// Test getAllTaskRuns function with name query
```
a-roberts: Can we really not have any test for it because field selector doesn't work with fake? I preferred the previous approach if so where we just iterated

vtereso:
(Pipeline/Task)Runs don't have metadata.name (mistake on my part), but rather to access the underlying name, the appropriate field selector would be `spec.(pipeline/taskRef).name`. Since field selectors don't have this functionality yet (https://github.com/kubernetes/kubernetes/issues/53459), post filtering seems like the best we can do.

a-roberts:
Thanks Vincent, only minor nits to go then happy to get this in

a-roberts:
/approve

All tests pass, looks good to me, thanks @vtereso!

a-roberts:
Retrospectively assigning @vtereso as he's now part of our org and implemented this.

/assign vtereso

PR #59, opened 2019-04-16
Patch service account in credential REST API

akihikokuroda:
When the credential REST API creates or deletes the secrets, the secrets are patched into the service account.


pkg/endpoints/credentials.go
```diff
@@ -152,6 +155,21 @@ func (r Resource) createCredential(request *restful.Request, response *restful.R
 		utils.RespondErrorAndMessage(response, err, errorMessage, http.StatusBadRequest)
 		return
 	}
+
+	// Patch service account
+	saName := ""
+	providedSaName := os.Getenv("PIPELINE_RUN_SERVICE_ACCOUNT")
```
a-roberts: We're going to discuss this further next week but we're thinking this should be sent in the body for the request as opposed to getting it from here as it's something we potentially want to get from the user

pkg/endpoints/credentials.go
```diff
@@ -343,3 +370,81 @@ func credentialToSecret(cred credential, namespace string, response *restful.Res
...
+
+// addSecretToSA - Add secret to service account 
+func (r Resource) addSecretToSA(saName string, secretName string , namespaceName string) bool {
+	sa := corev1.ServiceAccount{
+		Secrets: []corev1.ObjectReference {{Name: secretName},},
+	}
+
+	marshaledSA, err := json.Marshal(sa)
+	if err != nil {
+		logging.Log.Errorf("Failed to marshal payload")
```
a-roberts: Lower case f

pkg/endpoints/credentials.go
```diff
@@ -343,3 +370,81 @@ func credentialToSecret(cred credential, namespace string, response *restful.Res
...
+
+	_, err = r.K8sClient.CoreV1().ServiceAccounts(namespaceName).
+		Get(saName, metav1.GetOptions{})
+
+	if err != nil {
+		logging.Log.Infof("Service Account %s doesn't exist. Creating..", saName)
+		_, err = r.K8sClient.CoreV1().ServiceAccounts(namespaceName).
+			Create(&corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: saName,},})
+		if err != nil {
+			logging.Log.Errorf("Error occurred: %s", err.Error())
```
a-roberts: Lower case e for Errorf msg

pkg/endpoints/credentials.go
```diff
@@ -343,3 +370,81 @@ func credentialToSecret(cred credential, namespace string, response *restful.Res
...
+			return false
+		}
+	}
+
+	_, err = r.K8sClient.CoreV1().ServiceAccounts(namespaceName).
+		Patch(saName, typesPatch.StrategicMergePatchType, marshaledSA)
+
+	if err != nil {
+		logging.Log.Errorf("Error occurred: %s", err.Error())
```
a-roberts: Lower case e and let's mention the error occurred while trying to patch the sa

pkg/endpoints/credentials.go
```diff
@@ -343,3 +370,81 @@ func credentialToSecret(cred credential, namespace string, response *restful.Res
...
+		return false
+	}
+	return true
+}
+
+// path - struct for patch operation
+type patch struct {
```
a-roberts: If we're declaring a type this should be Patch

pkg/endpoints/credentials.go
```diff
@@ -343,3 +370,81 @@ func credentialToSecret(cred credential, namespace string, response *restful.Res
...
+	Op              string            `json:"op"`
```
a-roberts: Is Op operation? Guessing so but may be unclear to a newcomer, suggesting Operation as field name

pkg/endpoints/credentials.go
```diff
@@ -343,3 +370,81 @@ func credentialToSecret(cred credential, namespace string, response *restful.Res
...
+	Path            string            `json:"path"`
+}
+
+// removeSecretFromSA - remove secret from service account 
+func (r Resource) removeSecretFromSA(saName string, secretName string , namespaceName string) {
+	logging.Log.Debugf("Service account to patch is %s", saName)
```
a-roberts: Is it to *unpatch*? Or should we just say to *To remove the secret from*?

pkg/endpoints/credentials.go
```diff
@@ -343,3 +370,81 @@ func credentialToSecret(cred credential, namespace string, response *restful.Res
...
+	logging.Log.Debugf("Namespace of service account is %s", namespaceName)
+
+	sa, err := r.K8sClient.CoreV1().ServiceAccounts(namespaceName).
+		Get(saName, metav1.GetOptions{})
+	if err != nil {
+		logging.Log.Errorf("error getting service sccount %s: %s", saName, err.Error())
```
a-roberts: account not sccount

pkg/endpoints/credentials.go
```diff
@@ -343,3 +370,81 @@ func credentialToSecret(cred credential, namespace string, response *restful.Res
...
+	}
+
+	var entry int
+	found := false
+	for i, secret := range sa.Secrets {
+		if secret.Name == secretName {
+			found = true;
+			entry = i
```
a-roberts: Let's break on finding it?

pkg/endpoints/credentials.go
```diff
@@ -343,3 +370,81 @@ func credentialToSecret(cred credential, namespace string, response *restful.Res
...
+	}
+	if found {
+		path := fmt.Sprintf("/secrets/%d", entry)
+		data := []patch {{ Op: "remove", Path: path }} 
+		patch, err := json.Marshal(data)
+		logging.Log.Debugf("Patch JSON:%s", string(patch))
+		_, err = r.K8sClient.CoreV1().ServiceAccounts(namespaceName).
+			Patch(saName, typesPatch.JSONPatchType, patch)
+		if err != nil {
+			logging.Log.Errorf("Error occurred: %+v", err)
```
a-roberts: Lower case e, mention the operation (remove)

pkg/endpoints/credentials_test.go
```diff
@@ -119,6 +125,10 @@ func TestCredentials(t *testing.T) {
 		userPassCred,
 	}
 	readAllCredentialsTest(namespace, expectCreds, "", r, t)
+	// fake patch function for service account may not be implemented
```
a-roberts: Should we just remove these comments then?

pkg/endpoints/credentials_test.go
```diff
@@ -501,3 +515,20 @@ func (r Resource) getK8sCredential(namespace string, name string) (credential cr
 	}
 	return secretToCredential(secret,false)
 }
+
+// Get check service account secret
+func (r Resource) checkSecretInSa(namespace string, nameSa string, nameSecret string, t *testing.T) bool {
+	sa, err := r.K8sClient.CoreV1().ServiceAccounts(namespace).Get(nameSa, metav1.GetOptions{})
+	if err != nil {
+		t.Logf("Couldn't read service account: %+v\n", err)
```
a-roberts: Errorf?

pkg/endpoints/credentials_test.go
```diff
@@ -501,3 +515,20 @@ func (r Resource) getK8sCredential(namespace string, name string) (credential cr
...
+
+// Get check service account secret
+func (r Resource) checkSecretInSa(namespace string, nameSa string, nameSecret string, t *testing.T) bool {
+	sa, err := r.K8sClient.CoreV1().ServiceAccounts(namespace).Get(nameSa, metav1.GetOptions{})
+	if err != nil {
+		t.Logf("Couldn't read service account: %+v\n", err)		
+		return false
+	}
+	for _, secret := range sa.Secrets {
+		t.Logf("Got secret : %+v\n", secret.Name)
```
a-roberts: Debugf?

pkg/endpoints/credentials.go
akihikokuroda: I fix it.

pkg/endpoints/credentials_test.go
akihikokuroda: I take out this line.

pkg/endpoints/credentials.go
akihikokuroda: I see.  The create PipelineRun API will carry the service account name in the body.  So this api probably needs to do the same.

mnuttall: The problem with requesting service account details from the user, is that we don't yet have an API to provide a list of service accounts. So this would be free text entry in the front end, which isn't ideal. I'd be ok with a hardcoded service account, at least until we have a better API around service accounts.

akihikokuroda: This seems OK for now until we found issues for it.  Adding the service account name in the request causes that the service account name related to the secret must be persisted in somewhere.

a-roberts:
Changes look good now @akihikokuroda, thanks, can you please update with master then squash your commits into one, then I'll add the ok to test message?

akihikokuroda:
I opened the clean PR #97.  Closing this.

PR #66, opened 2019-04-23
Add pipelinerunlog API doc

a-roberts:
Should be a straight forward one, there's an inconsistency here too as only the TaskRun logs are formatted (so if you compare /taskrunlog to /pipelinerunlog output, there's a clear difference).

We could format the PipelineRun logs as well so it's easier to display in the UI - I'm thinking by TaskRun and then by Steps (so for each Task we find in the Pipeline, get those logs for each container, and display appropriately).


a-roberts:
@jessm12 please take a look 👼

PR #68, opened 2019-04-23
Install instructions using ko

a-roberts:
We can use ko apply to build, push, deploy ourselves, so suggest it


a-roberts:
@jessm12 please take a look 😄

mnuttall:
@a-roberts could you make it clear that the `kubectl port-forward` command applies to both installation mechanisms? (And then open a web browser etc). 

Secondly, do you think it worth replacing 'if you're logged in to Dockerhub' with a `docker login` step?

a-roberts:
@mnuttall thanks Mark, i agree, updated accordingly, and thanks too for taking a look @vdemeester!

Issue #7, opened 2019-04-09
Dashboard extension support


mnuttall:
It would be great if the dashboard offered an extension mechanism. 'Extensions' could be packaged behind Kube services, with labels of the form, 
```
label: tekton-dashboard-extension: true
label: tekton-dashboard-endpoints: greatThings
```
The extension should then be able to supply panelling into a react component on the dashboard - perhaps a new tab? - whose .js code would be able to interact with the extension via the /greatThings proxied endpoint. Extensions could support managing web hooks, log archiving, or integrations with other third party product.

akihikokuroda:
I can not assign this to myself but I'll work for this.

akihikokuroda:
The service that has the following items is added as the back-end extension.
```  
annotations:
    tekton-dashboard-endpoints: <path for the extension>
  labels:
    tekton-dashboard-extension: "true"
```
PR: https://github.com/tektoncd/dashboard/pull/34

akihikokuroda:
I believe that this can be closed.

PR #74, opened 2019-04-24
Add support for UI extensions

AlanGreene:
https://github.com/tektoncd/dashboard/issues/7

~~Still need to verify against final back end code, likely needs some minor tweaks to API calls before it's ready to merge.~~ Updated to reflect latest discussions on https://github.com/tektoncd/dashboard/issues/80

A future PR will extend this to provide access to a set of front-end APIs that can be consumed by extensions, e.g. accessing redux store.

### DIFF
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -11,8 +11,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React from 'react';
-import { hot } from 'react-hot-loader';
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { hot } from 'react-hot-loader/root';
 import {
   HashRouter as Router,
   Redirect,
@@ -21,6 +22,8 @@ import {
 } from 'react-router-dom';
 
 import {
+  Extension,
+  Extensions,
   Home,
   PipelineRun,
   PipelineRuns,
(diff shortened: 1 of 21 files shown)


src/containers/App/App.js
```diff
@@ -21,40 +22,91 @@ import {
...
+            component={PipelineRuns}
+          />
+          <Route path="/tasks/:taskName/runs" exact component={TaskRuns} />
+          <Route
+            path="/pipelines/:pipelineName/runs/:pipelineRunName"
+            component={PipelineRun}
+          />
+          <Route path="/extensions" exact component={Extensions} />
+          {extensions.map(({ displayName, name, source }) => {
+            if (!displayName) {
```
steveodonovan: What would cause this case ?

src/containers/App/App.js
```diff
@@ -21,40 +22,91 @@ import {
...
+            exact
+            component={PipelineRuns}
+          />
+          <Route path="/tasks/:taskName/runs" exact component={TaskRuns} />
+          <Route
+            path="/pipelines/:pipelineName/runs/:pipelineRunName"
+            component={PipelineRun}
+          />
+          <Route path="/extensions" exact component={Extensions} />
+          {extensions.map(({ displayName, name, source }) => {
```
steveodonovan: move the conditional for display name to a filter ?

src/containers/Extension/Extension.js
```diff
@@ -0,0 +1,60 @@
...
+
+    this.state = { ExtensionComponent };
+  }
+
+  render() {
+    const { displayName, match } = this.props;
+    const { ExtensionComponent } = this.state;
+
+    return (
+      <div className="pipelines">
```
steveodonovan: Should we make these classes more generic ? Can be tackled in a seperate PR/issue

src/containers/Home/Home.js
```diff
@@ -57,11 +58,22 @@ const Home = () => {
                 <Link to="/tasks">Tasks</Link>
               </StructuredListCell>
             </StructuredListRow>
+            {!!extensions.length && (
```
steveodonovan: `!!` ?

src/containers/App/App.js
AlanGreene: The displayname and bundle-location annotations are optional, and only relevant for extensions that provide a UI. If they're omitted there's no need for us to support rendering a UI for that extension, so we skip it when outputting dynamic routes.

src/containers/Extension/Extension.js
AlanGreene: Yeah, haven't changed these yet waiting for the new layout.

src/containers/Home/Home.js
AlanGreene: `!!0 === false` (which React will then conveniently ignore in the output)
however if I left it as just `extensions.length` it'd display `0` on the page

src/containers/App/App.js
AlanGreene: are you thinking something like this?

```js
extensions
  .filter(({ displayName }) => !!displayName)
  .map(({ displayName, name, source }) => (
    <Route
      key={name}
      path={`/extensions/${name}`}
      render={({ match }) => (
        <Extension
          displayName={displayName}
          match={match}
          source={source}
        />
      )}
    />
  ));
```

I went with including it in the `map` originally as it's just a single pass over the extensions (taking advantage of the fact that React skips `nulls` when rendering), but since it should be a relatively small list I'm ok with either.

Issue #14, opened 2019-04-10
Host front-end from back-end


skaegi:
We currently host our front-end in a webpack dev server and proxy to the back-end. We need to figure out how we want our url paths to the back-end to work and then host the results of a webpack build via a mechanism like https://github.com/emicklei/go-restful/blob/master/examples/restful-serve-static.go

PR #78, opened 2019-04-24
Support web project from back and and builders

steveodonovan:
@AlanGreene @a-roberts Could I get a review ? Also try it out. I did have to add the tekton-pipelines service account to the deployment config to make it work but I might be missing something here. Also Ideally the same deployment wouldn't directly reference the `/var/run/ko/web` directory and instead use the variables already available [described here](https://github.com/google/ko#including-static-assets). Very open to suggestions on that one.

I've added a `npm build_ko` task which once run will combine the two in the ko builder. The regular docker build will build the two without any extra steps.

https://github.com/tektoncd/dashboard/issues/14

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -114,17 +114,31 @@ Tekton Dashboard does not involve any custom resource definitions, we only inter
 You can stand up a version of the dashboard on-cluster (to your
 `kubectl config current-context`):
 
+First install and build the npm project.
+
+```shell
+npm install
+```
+
+There is a dedicated npm job for ko builds
+
+```shell
+npm run build_ko
+```
+
+This will build the static resources and add them to the `kodata` directory.
+
 ```shell
 ko apply -f config/
 ```
 
 ## Access the dashboard
 
-To access the backend API:
+To access the dashboard:
 
 `kubectl port-forward $(kubectl get pod -l app=tekton-dashboard -o name) 9097:9097`
 
-Note that we have a big TODO which is to link up the frontend to the backend and to document how to build and run the frontend for yourself.
+The web UI can be accessed at http://localhost:9097
 
 ### Redeploy dashboard
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.12-alpine as builder
+FROM node:10-alpine as nodeBuilder
+USER root
+WORKDIR /go/src/github.com/tektoncd/dashboard
+COPY . .
+RUN npm install
+RUN npm run build
+
+FROM golang:1.12-alpine as goBuilder
 USER root
 RUN apk add curl git
 RUN curl -fsSL -o /usr/local/bin/dep https://github.com/golang/dep/releases/download/v0.5.0/dep-linux-amd64 && chmod +x /usr/local/bin/dep
@@ -24,6 +31,8 @@ RUN addgroup -g 1000 kgroup && \
 USER 1000
 
 WORKDIR /go/src/github.com/tektoncd/dashboard
-COPY --from=builder /go/src/github.com/tektoncd/dashboard/tekton_dashboard_backend .
+ENV WEB_RESOURCES_DIR=./web
+COPY --from=nodeBuilder /go/src/github.com/tektoncd/dashboard/dist ./web
+COPY --from=goBuilder /go/src/github.com/tektoncd/dashboard/tekton_dashboard_backend .
 
 ENTRYPOINT ["./tekton_dashboard_backend"]
--- a/config/tekton-dashboard-deployment.yaml
+++ b/config/tekton-dashboard-deployment.yaml
@@ -32,6 +32,8 @@ spec:
         env:
         - name: PORT
           value: "9097"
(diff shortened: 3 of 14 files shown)


a-roberts:
Taking a look at this now, thank you @steveodonovan!

DEVELOPMENT.md
```diff
@@ -114,17 +114,31 @@ Tekton Dashboard does not involve any custom resource definitions, we only inter
...
 
 ## Access the dashboard
 
-To access the backend API:
+To access the dashboard:
 
 `kubectl port-forward $(kubectl get pod -l app=tekton-dashboard -o name) 9097:9097`
 
-Note that we have a big TODO which is to link up the frontend to the backend and to document how to build and run the frontend for yourself.
+The web UI can be accessed at http://localhost:9097
```
a-roberts: 404 page not found 😞 I see the Go code looks ok, I don't get my logging.Log.Info("Adding web api") message

pkg/endpoints/routes_test.go
```diff
@@ -82,6 +83,21 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
+func TestGetWeb200(t *testing.T) {
+	t.Log("Checking GET route for 200 StatusCode")
```
a-roberts: Nit around inconsistent whitespace throughout this method, should be inline with our other Go test files (we're using tabs)

DEVELOPMENT.md
```diff
@@ -114,17 +114,31 @@ Tekton Dashboard does not involve any custom resource definitions, we only inter
...
 
 ### Redeploy dashboard
```
a-roberts: Let's also update README.md's getting started section? Currently one would follow the instructions and just get the backend, or so they might think. With these changes merged that's going to differ

Dockerfile
```diff
@@ -11,19 +11,24 @@
 
 FROM golang:1.12-alpine as builder
 USER root
-RUN apk add curl git
+RUN apk add curl git nodejs npm
```
AlanGreene: node@10 and npm@6 for consistency
we can look at newer versions once everything's up and running

src/api/index.js
```diff
@@ -14,13 +14,12 @@ limitations under the License.
 import { deleteRequest, get, post, put } from './comms';
 
 export function getAPIRoot() {
-  const apiPrefix = 'api';
```
AlanGreene: need to ensure dev mode still works
so maybe something like (untested):

```js
if (process.env.NODE_ENV === 'development') {
  return '/api'
}

// else get baseURL and return it
```

DEVELOPMENT.md
a-roberts: ^^ above was from my own directory structure being wrong, now I get:

```
{"level":"info","msg":"Adding web api"}
{"level":"info","msg":"Adding v1, and API for pipelines"}
{"level":"info","msg":"Adding API for websocket"}
{"level":"info","msg":"Adding API for health"}
{"level":"info","msg":"Adding API for readiness"}
{"level":"info","msg":"Searching for extensions in the namespace "}
{"level":"info","msg":"Adding API for extensions"}
{"level":"debug","msg":"Extension: []"}
panic: http: multiple registrations for /

goroutine 1 [running]:
net/http.(*ServeMux).Handle(0xc00036c390, 0x1223e06, 0x1, 0x13726e0, 0xc0004a3ec0)
        /usr/local/go/src/net/http/server.go:2377 +0x221
net/http.(*ServeMux).HandleFunc(0xc00036c390, 0x1223e06, 0x1, 0xc0004a3ec0)
        /usr/local/go/src/net/http/server.go:2395 +0x5a
```

DEVELOPMENT.md
```diff
@@ -114,17 +114,31 @@ Tekton Dashboard does not involve any custom resource definitions, we only inter
 You can stand up a version of the dashboard on-cluster (to your
 `kubectl config current-context`):
 
+First install and build the npm project.
+
+```shell
+npm install
```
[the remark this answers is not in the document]
steveodonovan: https://github.com/tektoncd/dashboard/issues/84 tackled in its own prhttps://github.com/tektoncd/dashboard/pull/85

steveodonovan:
@AlanGreene @a-roberts Could I get another review please ? Not to familiar with the extensions pieces.

config/tekton-dashboard-deployment.yaml
```diff
@@ -32,7 +32,9 @@ spec:
         env:
         - name: PORT
           value: "9097"
-        - name: INSTALLED_NAMESPACE
+        - name: WEB_RESOURCES_DIR
+          value: /var/run/ko/web
+        - name: INSTALL_NAMESPACE
```
AlanGreene: looks like a bad merge, should be `INSTALLED_NAMESPACE`

DEVELOPMENT.md
a-roberts: This would be a noddy one to do outside of this PR as part of a general cleanup

a-roberts:
Yeah, looks good to me now @steveodonovan, I want to get this in and run with the main readme next

/approve

If @AlanGreene is happy too I'm good to get this merged

vdemeester:
/lgtm
:bowing_man:

Issue #56, opened 2019-04-16
Task's step(s) showing as success even if they are not


vdemeester:
If a step fails (using `tektoncd/pipeline@master`), dashboard might report it as success even though it is not (hence the task status and the pipeline status in the following screenshot)

![2019-04-16-191519](https://user-images.githubusercontent.com/6508/56230364-5e8cae80-607c-11e9-9809-1fd20ce0943a.png)

Looking at the status, it didn't look at the correct status for the step (the step exited with 1, but the status reported 1)

The screenshot is taken on a run based on the following files : https://sbr.pm/technical/tekton-usage.html#org63ddf32

/kind bug

AlanGreene:
Now that step name is included in task run status (https://github.com/tektoncd/pipeline/pull/657) it should be much easier to reliably match the step / status without relying on order and trying to account for init containers.

/assign

AlanGreene:
@vdemeester issue should be resolved now, let me know if you still get any unexpected step status.

PR #95, opened 2019-04-26
Use newly added name field in step status to find correct results

AlanGreene:
https://github.com/tektoncd/dashboard/issues/56

Instead of attempting to match step status in task runs by index,
accounting for presence / order of init containers, use the
newly added name field in the step status so we can reliably
link steps to their status and logs regardless of order.

This involves updating Gopkg.lock to use github.com/tektoncd/pipeline
revision 3415c18f7728476c132f692d9a55a70b7b194c5d which introduced
the new field.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -158,12 +158,12 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:58dc2abfa0d3901dde142cd636abf13d495f37c125564390c7a5791d1fa29e80"
+  digest = "1:b6fca2d4933db5050e8729d0931acc650c2492daa13decc64ad62d48145f7d20"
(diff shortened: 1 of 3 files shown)


AlanGreene:
/hold
@a-roberts this is the change we discussed earlier in the week. There's no reliable way to identify the correct step status without the new field. If you have no objections to updating the version of pipeline we consume this should be good to go.

a-roberts:
So we're not breaking anything on 0.2 and we're benefitting from a new change that went into master about a week ago and thus the 0.3 release right @AlanGreene?

AlanGreene:
That's right, everything else appears to be working as expected.

a-roberts:
Sounds good to me then, thanks

/approve
/lgtm

PR #99, opened 2019-04-29
Modify the readme for dockerbuild installation method.

dibbles:
The docker build version of the install instruction do not mention docker pushing the image or more importantly the need to modify the location of the web resource bundle, this PR corrects that.


steveodonovan:
https://github.com/tektoncd/dashboard/blob/master/Dockerfile#L34 You shouldn't need to set that value

Issue #86, opened 2019-04-25
Allow namespace selection


steveodonovan:
At the moment the dashboard will only show resources in the default namespace.

This effort is to track listing namespaces and selecting a namespace for the pipeline and task runs views

API `GET /v1/namespaces`

ffahri:
Hey, isn't it much more nice to user give namespace for dashboard backend. So backend list resources by giving namespace variable. (like kube dashboard)

steveodonovan:
@ffahri Yep thats the longer term plan. Haven't started this just yet though, might look to do that in the short term instead.

The design here includes the selection https://github.com/tektoncd/dashboard/issues/35

PR #101, opened 2019-04-29
Add namespace selection

steveodonovan:
@a-roberts Could I get a review please ? Thanks

This is provides a basic api to get the namespaces for a cluster for selection in the UI 

https://github.com/tektoncd/dashboard/issues/86

### DIFF
--- a/pkg/endpoints/cluster.go
+++ b/pkg/endpoints/cluster.go
@@ -1,0 +1,34 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+		http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package endpoints
+
+import (
+  "github.com/emicklei/go-restful"
+  "github.com/tektoncd/dashboard/pkg/logging"
+  "github.com/tektoncd/dashboard/pkg/utils"
+  "net/http"
+
+  metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r Resource) getAllNamespaces(request *restful.Request, response *restful.Response) {
+  namespaces, err := r.K8sClient.CoreV1().Namespaces().List(metav1.ListOptions{});
+  logging.Log.Debugf("In getAllNamespaces - : %s", namespaces)
+
+  if err != nil {
+    utils.RespondError(response, err, http.StatusNotFound)
+    return
+  }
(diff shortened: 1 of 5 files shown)


pkg/endpoints/cluster.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/tektoncd/dashboard/pkg/logging"
+	"github.com/tektoncd/dashboard/pkg/utils"
+	"net/http"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r Resource) getAllNamespaces(request *restful.Request, response *restful.Response) {
+	namespaces, err := r.K8sClient.CoreV1().Namespaces().List(metav1.ListOptions{});
+	logging.Log.Debugf("In getAllNamespaces - : %s", namespaces)
```
a-roberts: I'd have this just be "In getAllNamespaces" in the Debugf

pkg/endpoints/cluster.go
```diff
@@ -0,0 +1,34 @@
...
+
+	if err != nil {
+		utils.RespondError(response, err, http.StatusNotFound)
+		return
+	}
```
a-roberts: Then logging.Log.Debugf("Found namespaces: %s, namespaces") here

pkg/endpoints/cluster_test.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"encoding/json"
+	"net/http/httptest"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestGetAllNamespaces(t *testing.T) {
```
a-roberts: Can you add another short test please for a namespace that doesn't exist then we test that we return the 404?

pkg/endpoints/router.go
```diff
@@ -52,8 +52,9 @@ func (r Resource) RegisterEndpoints(container *restful.Container) {
 		Consumes(restful.MIME_JSON).
 		Produces(restful.MIME_JSON)
 
-	logging.Log.Info("Adding v1, and API for pipelines")
+	logging.Log.Info("Adding v1, and API for k8s resources and pipelines")
 
+	wsv1.Route(wsv1.GET("/").To(r.getAllNamespaces))
```
a-roberts: Document this new endpoint under DEVELOPMENT.md please 😄 I know it's not really going to be used much by end users but at least we'll all know it's exist and what its contract is

pkg/endpoints/cluster_test.go
dibbles: I don't think this is needed @a-roberts, the namesapce is added so there is a namespace to be returned, this is not adding testing getting a specific namespace ... rather getting all namespaces.

a-roberts: Yep you're right, my mistake 😅 will resolve this convo...

steveodonovan:
Thanks, @a-roberts could I get another once over please ?

a-roberts:
Looks perfect to me, thank you @steveodonovan!

/approve
/lgtm

PR #102, opened 2019-04-29
Update data layer to add support for namespace selection

AlanGreene:
https://github.com/tektoncd/dashboard/issues/86

Move specification of namespace to the redux store instead of the API layer.
This will allow us to provide UI for the user to select the namespace to inspect.
Update existing containers and API calls to ensure the selected namespace is passed.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,7 @@ module.exports = {
       "react"
   ],
   "rules": {
+    "no-case-declarations": "off",
     "no-template-curly-in-string": "off",
     "react/jsx-filename-extension": "off",
     "react/destructuring-assignment": "off",
--- a/src/actions/pipeline.js
+++ b/src/actions/pipeline.js
@@ -12,21 +12,24 @@ limitations under the License.
 */
 
 import { getPipelines } from '../api';
+import { getSelectedNamespace } from '../reducers';
 
-export function fetchPipelinesSuccess(data) {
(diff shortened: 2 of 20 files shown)


src/reducers/pipelines.js
```diff
@@ -48,17 +67,27 @@ function errorMessage(state = null, action) {
...
 });
 
-export function getPipelines(state) {
-  return Object.values(state.byName);
+export function getPipelines(state, namespace) {
+  const pipelines = state.byNamespace[namespace];
+  if (!pipelines) {
+    return [];
+  }
+  return Object.values(pipelines).map(id => state.byId[id]);
```
steveodonovan: `return pipelines  ? Object.values(pipelines).map(id => state.byId[id]) : [];`

cleaner ?

src/reducers/pipelines.js
```diff
@@ -48,17 +67,27 @@ function errorMessage(state = null, action) {
...
 
-export function getPipeline(state, name) {
-  return state.byName[name];
+export function getPipeline(state, name, namespace) {
+  const pipelines = state.byNamespace[namespace] || {};
+  const pipelineId = pipelines[name];
+  if (!pipelineId) {
+    return null;
+  }
+  return state.byId[pipelineId];
```
steveodonovan: ditto on the inline conditional ?

AlanGreene:
@steveodonovan updated, thanks

PR #112, opened 2019-05-02
Gofmt'd packages, add makefile to simplify test/format/squashing

vtereso:
# Changes
 
- Add makefile to simplify test/format/squashing
- Ran format rule to format dashboard go files

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -4,11 +4,8 @@
 
 Many of the instructions here replicate what's in the [tektoncd/pipeline development guide](https://github.com/tektoncd/pipeline/blob/master/DEVELOPMENT.md), with a couple of caveats currently.
 
-1. This project currently does not support being built with `ko`
-2. This project has not yet been tested with GKE
-3. The instructions here pertain to building and deploying the backend that lets us interact with Tekton resources, as well as running the frontend locally for development. Instructions for deploying the frontend with the backend are to follow.
-
-We would love to accomplish these tasks and to update this document, contributions welcome!
+- This project has not yet been tested with GKE
+- The instructions here pertain to building and deploying the backend that lets us interact with Tekton resources, as well as running the frontend locally for development.
 
 ### Requirements
 
@@ -16,7 +13,7 @@ You must install these tools:
 
 1. [`go`](https://golang.org/doc/install): The language Tekton Dashboard is built in
 1. [`git`](https://help.github.com/articles/set-up-git/): For source control
-1. [`dep`](https://github.com/golang/dep): For managing external Go dependencies. - Please Install dep v0.5.0 or greater.
+1. [`dep`](https://github.com/golang/dep): For managing external Go dependencies. - Please install dep v0.5.0 or greater.
 1. [`ko`](https://github.com/google/ko): For development. `ko` version v0.1 or higher is required for `dashboard` to work correctly.
 1. [Node.js & npm](https://nodejs.org/): For building and running the frontend locally. See `engines` in [package.json](./package.json) for versions used.
 1. [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/): For interacting with your kube cluster.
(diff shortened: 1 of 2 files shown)


vtereso:
There is no open issue for this and it is not critical/necessary. However, I believe that `make format` is a bit easier than `go fmt ./cmd/... ./pkg/...`.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+	git reset --soft $(RESET_TARGET)
+	git commit -m "$(COMMIT)"
+	git push --force origin $(BRANCH)
+endif
+
+format: ## Runs go fmt on all non-vendor go packages
+	go fmt $(PKGS)
+
+test: ## Runs `go test -count=1 -race -v ./pkg/...`
+	go test -count=1 -race -v ./pkg/...
```
a-roberts: This Makefile look fine although will want to be documented (e.g. when do folks use it?), and how about an extra target for running the Docker tests (now we have Dockerfile_test and Dockerfile_race_test)?

a-roberts: I think it's best suited to DEVELOPMENT.md

I suggest `make test` runs our main Dockerfile_test and if you do `make race test` it runs both our main Dockerfile_test and the Dockerfile_race_test (so they're both independent targets, but you can combine them if you want)

A "final check before we merge" would involve `make race test` for example, so we get all the coverage. Lemme know what you think.

vtereso: Both of the dockerized files have nearly the same execution time
`time docker build -f Dockerfile_race_test .`=`real	2m2.667s`
`time docker build -f Dockerfile_test .`=`real	2m3.154s`
A final check (one that handles race conditions) makes sense, but is there a particular reason we don't consolidate to one of these files?
I always prefer to keep things as lean as possible, same regards for the Makefile. We could have `make test`, which would be undockerized and `make docker/test` will run our racefile? I made some changes.

a-roberts: `A final check (one that handles race conditions) makes sense, but is there a particular reason we don't consolidate to one of these files?`

Can't run the race tests in the Alpine image

Yeah I like `make test` undockerized and `make docker/test` as running *both* the Dockerfiles - we don't want to lose test coverage (one is Alpine which we "ship on" but w/o race checking, one is Stretch which we don't "ship on" but lets us do the race checks)

I say `ship on` in quotes as we're gonna end up releasing images eventually and they're gonna be Alpine for running on our target platforms I bet

Makefile
```diff
@@ -0,0 +1,39 @@
...
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+ifeq ($(COMMIT),)
+squash:
+	@echo "COMMIT=<COMMIT> not provided. Exiting..."
+	@exit 1
+else
+
+# Assuming PR is ready, everything added as intended, no conflicts
+squash: docker/test format ## Runs dockerized test, formats, and force pushes. Override RESET_TARGET='upstream/master' if necessary
```
a-roberts: This is pretty scary, I'd honestly not use this simply because it's involving a force push. I'd much rather make a copy of the user's current branch and force push the changes to that. A user can then, if that worked, force push from their copied branch to the main one. Then there's no chance of disaster

Makefile
```diff
@@ -0,0 +1,39 @@
+RESET_TARGET := upstream/master
+override PKGS := $(shell go list ./... | grep -v /vendor/)
+override BRANCH := $(shell sh -c "git branch | sed -n 's|* \(.*\)|\1|p'")
+override STAGED_GO_FILES := $(shell git diff --name-only --cached | grep -E ".*.go")
+
+.PHONY: squash format fmt dep test docker/test
+
+help:
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```
a-roberts: A comment would be useful here. Guessing it prints every target including the comment in this file, but oh nelly it's a beast.

DEVELOPMENT.md
```diff
@@ -100,15 +97,21 @@ Currently you must install the Tekton dashboard into the same namespace you wish
 
 ## Iterating
 
-While iterating on the project, you may need to:
+While iterating on the project, the following are common:
```
a-roberts: The following what? Operations? Actions?

Makefile
vtereso: Yeah, I pulled it from online since it's pretty common procedure with self documentation make files to avoid having to constantly update some help table

DEVELOPMENT.md
vtereso: Will revert.

Makefile
vtereso: Will add a comment

mnuttall: that's an awesome one-liner : )

Makefile
vtereso: Removing pushing logic. Commit squashing only on test success

Makefile
```diff
@@ -0,0 +1,40 @@
...
+	go fmt $(PKGS)
+	@echo "Adding back all files that have been formatted"
+	@echo $(STAGED_GO_FILES) | xargs -I {} git add {}
+
+fmt: format ## Alias for `make format`
+
+dep: ## Runs `dep ensure -v`
+	dep ensure -v
+
+test: ## Runs `go test -count=1 -race -v ./pkg/...`
```
mnuttall: The alias for `test` is useful. The `git` commands I'm less sure about. We do need people to squash their commits but I'm uneasy about telling folks how to do that. I could let this in if the git commands were removed from the Makefile. Otherwise I'm not sure this PR is getting in; it introduces as many problems as it solves.

mnuttall:
Hi @vtereso we're still unhappy about `make squash`. Rather than have this PR hang around forever I'm going to close it for now.

Issue #121, opened 2019-05-05
Query parameters for the task runs api not working as expected


steveodonovan:
Query parameters for the task runs api not working as expected

# Expected Behavior
- Filtering should be possible for any taskRun
- filtering should be by task name not task run name

# Actual Behavior
- fails with error for not found case, filters by run name in the successful case

# Steps to Reproduce the Problem
For the exception
1. run the query with a name which will not be found

# Additional Info

```
{"level":"debug","msg":"In getAllTaskRuns - namespace: default, parameters: name: pipeline0-task"}
2019/05/05 14:45:43 http: panic serving 127.0.0.1:51390: runtime error: index out of range
goroutine 7188 [running]:
net/http.(*conn).serve.func1(0xc00009cb40)
	/usr/local/Cellar/go/1.12/libexec/src/net/http/server.go:1769 +0xc9
panic(0x116b800, 0x1f53d00)
	/usr/local/Cellar/go/1.12/libexec/src/runtime/panic.go:522 +0x1b5
github.com/tektoncd/dashboard/pkg/endpoints.Resource.getAllTaskRuns(0x1474fc0, 0xc000319a40, 0x14bdba0, 0xc00020cb00, 0xc0001df140, 0xc00023a070)
	/Users/stephen/cloud/go/src/github.com/tektoncd/dashboard/pkg/endpoints/pipeline.go:408 +0x698
github.com/tektoncd/dashboard/vendor/github.com/emicklei/go-restful.(*Container).dispatch(0xc0002b3e60, 0x147ec40, 0xc0004cc2a0, 0xc000732100)
	/Users/stephen/cloud/go/src/github.com/tektoncd/dashboard/vendor/github.com/emicklei/go-restful/container.go:288 +0xa3b
net/http.HandlerFunc.ServeHTTP(0xc000341210, 0x147ec40, 0xc0004cc2a0, 0xc000732100)
	/usr/local/Cellar/go/1.12/libexec/src/net/http/server.go:1995 +0x44
net/http.(*ServeMux).ServeHTTP(0xc00031c340, 0x147ec40, 0xc0004cc2a0, 0xc000732100)
	/usr/local/Cellar/go/1.12/libexec/src/net/http/server.go:2375 +0x1d6
github.com/tektoncd/dashboard/vendor/github.com/emicklei/go-restful.(*Container).ServeHTTP(0xc0002b3e60, 0x147ec40, 0xc0004cc2a0, 0xc000732100)
[...]

PR #122, opened 2019-05-05
Task runs and pipeline runs by name not working as expected

steveodonovan:
https://github.com/tektoncd/dashboard/issues/121

# Changes

Fix name filtering on task runs. The api at the moment cant handle the empty response and is filtering by the wrong field.

# Submitter Checklist

review them:

for more details._]

@a-roberts Could I get a review please ? Thanks

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -231,6 +231,7 @@ Returns HTTP code 404 if an error occurred getting the Pipeline
 __PipelineRuns__
 ```
 GET /v1/namespaces/<namespace>/pipelinerun
+Get all Tekton PipelineRuns, also supports '?name=<pipeline-name>' querying
 Get all Tekton PipelineRuns, also supports '?repository=https://gitserver/foo/bar' querying
 Returns HTTP code 200 and a list of PipelineRuns, optionally matching the above query, in the given namespace
 Returns HTTP code 404 if an error occurred getting the PipelineRun list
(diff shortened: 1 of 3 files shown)


a-roberts:
@vtereso could you take a look please? I noticed you added the initial code for this [here](https://github.com/tektoncd/dashboard/pull/57) and we had an internal (not shared issue, before we open sourced) issue that was initially worked too.

Surely we want both the ability to filter on TaskRun and PipelineRunName and then Task and Pipeline name, at a quick glance it looks like this replaces one function with another?

steveodonovan:
@a-roberts @vtereso  Just to clarify, we already have run by run name, it uses the uri over query parameters. 

`GET /v1/namespaces/<namespace>/pipelinerun/<pipelinerun-name>`

ditto for task runs.

see https://github.com/tektoncd/dashboard/blob/master/DEVELOPMENT.md

vtereso:
Good changes, my mistake checking against the runCRD rather than their references.

a-roberts:
Discussed these changes with Stephen over Slack, so we're using the query param to view a certain Task or Pipeline for a TaskRun or PipelineRun now, and we use the path param to view a certain TaskRun or PipelineRun.

We'll need modifications to our API docs (DEVELOPMENT.md then).

One aside (not to do with this PR but noticed it when looking at the whole file diff to get some context) 

This check is gross and confusing - I don't like these "negative" tests when we can just check the first thing returned is what we want.

https://github.com/tektoncd/dashboard/blob/fc5dab84aff3d4ed388f197934dcbf26008329e9/pkg/endpoints/pipeline_test.go#L157

steveodonovan:
@a-roberts good to go. Thanks

PR #123, opened 2019-05-06
Add service, remove port-forward

vtereso:
# Changes
Modify deployment.yaml to include service with clusterIP that is accessed through `kubectl proxy`
Updated DEVELOPMENT.md

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -4,9 +4,8 @@
 
 Many of the instructions here replicate what's in the [tektoncd/pipeline development guide](https://github.com/tektoncd/pipeline/blob/master/DEVELOPMENT.md), with a couple of caveats currently.
 
-1. This project currently does not support being built with `ko`
-2. This project has not yet been tested with GKE
-3. The instructions here pertain to building and deploying the backend that lets us interact with Tekton resources, as well as running the frontend locally for development. Instructions for deploying the frontend with the backend are to follow.
+- This project has not yet been tested with GKE
+- The instructions here pertain to building and deploying the backend that lets us interact with Tekton resources, as well as running the frontend locally for development.
 
 We would love to accomplish these tasks and to update this document, contributions welcome!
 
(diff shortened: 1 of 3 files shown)


vtereso:
I am not entirely sure that we want this, but opening as an option.

steveodonovan:
Once we have a real release with an artefact we will probably want a service defined but it should probably use clusterIP over node port. Also it can be combined into a single config similar to the kubernetes dashboard [see here](https://raw.githubusercontent.com/kubernetes/dashboard/v1.10.1/src/deploy/recommended/kubernetes-dashboard.yaml)

The the standard way to access it would be using kube proxy over port forwarding (again similar to the kubernetes dashboard instructions)

mnuttall:
@skaegi I'd appreciate your thoughts on this one. I'm not happy setting a NodePort to any specific value due to the risk of port collision. As @steveodonovan says, we should be following the pattern of the kubernetes dashboard. https://raw.githubusercontent.com/kubernetes/dashboard/v1.10.1/src/deploy/recommended/kubernetes-dashboard.yaml doesn't use a NodePort. I'd rather not take this, and wait for `kubectl proxy` support as per https://kubernetes.io/docs/tasks/access-application-cluster/web-ui-dashboard/.

steveodonovan:
Just had to do this for something else. Here are the changes needed.

Add this to the bottom of the deployment yml (maybe rename the deployment yml to be tekton-dashboard.yml, not sure on that one)

```
---
# ------------------- Dashboard Service ------------------- #
apiVersion: v1
kind: Service
metadata:
  name: tekton-dashboard-service
  labels:
    app: tekton-dashboard-service
spec:
  ports:
    - name: http
      port: 9097
      targetPort: 9097
  selector:
    app: tekton-dashboard
```


url to access the dashboard, update the doc to reflect this in place of port-forward.

http://localhost:8001/api/v1/namespaces/default/services/http:tekton-dashboard-service:9097/proxy/#/pipelines

mnuttall:
@vtereso please can you check out @steveodonovan 's comments above about how to get `kubectl proxy` working? I don't know that we want to steal port 9097 - let's talk about that - but can you please confirm that the approach works, and if so rework the PR accordingly? Thank you!

Issue #91, opened 2019-04-26
Dashboard should display real time updates to status of resources


AlanGreene:
For example:
- when viewing a list of PipelineRuns:
  - new PipelineRuns should be added to the list dynamically
  - the status of the displayed PipelinesRuns should be updated in real time
- when viewing a single PipelineRun:
  - TaskRun / step status should be updated in realtime
  - logs should be updated in realtime
- etc.

The back end should provide a websocket connection which publishes events to the front end.

Code exists today providing a dedicated websocket service for pipelineruns, and I have a [test branch](https://github.com/AlanGreene/dashboard/commit/ec22e183adb641c949f4d5314e319163f99b3fb9) using that to dynamically update the PipelineRun list view in the front end.

The back end should provide a single websocket service through which all relevant events are sent, allowing the front end to maintain a single connection, rather than managing multiple parallel connections for various purposes.

vtereso:
Currently there is a websocket that ONLY delivers Create/Update/Delete information on PipelineRuns, but it will be simple enough to send all the data through a single socket. What are all the events that we wish to send?

AlanGreene:
Yes that's the one I've been using for testing/experiments so far and it's working great 👍 

We should probably approach this as a series of steps. I'd suggest:
1. move the existing PipelineRuns websocket from `/v1/websocket/pipelineruns` to `/v1/websocket`
1. enable front end support for real time updates of PipelineRuns (code mostly complete, just needs a little cleanup and tests) using that websocket
1. discuss / prioritise additional requirements for events that will be sent over that same single socket:
    - logs
    - TaskRuns
    - Pipelines
    - Tasks
    - PipelineResources
    - etc.

I don't think we have a full list of events we'd like / need just yet. As more details of the front end design begin to emerge next week we can start having more informed discussions around user expectations.

I'm open to suggestions on any / all of the points above.

vtereso:
Have we completed 1/2?

AlanGreene:
I don't think there's been any work done on this. We should discuss 'filtering / subscription to subset of event types' before we do anything in this area as we probably don't want the client dealing with a total firehose of all events from the cluster...

PR #139, opened 2019-05-10
Add a namespaces dropdown component

ncskier:
# Changes

I added a Namespaces dropdown component that can be reused many places in our future UI.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/NamespacesDropdown/NamespacesDropdown.js
+++ b/src/containers/NamespacesDropdown/NamespacesDropdown.js
@@ -1,0 +1,40 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from 'react';
+import { connect } from 'react-redux';
+import { Dropdown, DropdownSkeleton } from 'carbon-components-react';
+
+import { getNamespaces, isFetchingNamespaces } from '../../reducers';
+
+const NamespacesDropdown = ({ loading, ...dropdownProps }) => {
+  if (loading) {
+    return <DropdownSkeleton {...dropdownProps} />;
+  }
+  return <Dropdown {...dropdownProps} />;
+};
+
+NamespacesDropdown.defaultProps = {
+  label: 'Namespace',
+  items: [],
+  loading: true
+};
+
+function mapStateToProps(state) {
+  return {
+    items: getNamespaces(state),
+    loading: isFetchingNamespaces(state)
+  };
+}
+
+export default connect(mapStateToProps)(NamespacesDropdown);
--- a/src/containers/NamespacesDropdown/NamespacesDropdown.stories.js
+++ b/src/containers/NamespacesDropdown/NamespacesDropdown.stories.js
@@ -1,0 +1,63 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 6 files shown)


src/containers/NamespacesDropdown/NamespacesDropdown.js
```diff
@@ -0,0 +1,54 @@
...
+import React from 'react';
+import { connect } from 'react-redux';
+import { Dropdown } from 'carbon-components-react';
+
+import { fetchNamespaces } from '../../actions/namespaces';
+import { getNamespaces } from '../../reducers';
+
+class NamespacesDropdown extends React.Component {
+  componentDidMount() {
+    this.props.fetchNamespaces();
```
AlanGreene: namespaces should always be available in the dashboard so no need to request them here

src/containers/NamespacesDropdown/NamespacesDropdown.js
```diff
@@ -0,0 +1,54 @@
...
+      <div>
+        <Dropdown {...dropdownProps} />
+      </div>
+    );
+  }
+}
+
+NamespacesDropdown.defaultProps = {
+  label: 'Namespace',
+  items: ['loading...']
```
AlanGreene: I'd either leave this as an empty array or pre-populate with 'default' rather than 'loading...' since this value would actually be selected.

Alternatively, and probably better, expose the `isFetchingNamespaces` selector in reducers/index.js and use that to implement a real loading state using `DropdownSkeleton` or similar.

ncskier: Thanks Alan, I didn't know about the `DropdownSkeleton`. That is exactly what I wanted to do for loading, but didn't know what I could use to indicate the loading state 😄

src/containers/NamespacesDropdown/NamespacesDropdown.js
ncskier: How frequently are the namespaces updated? Can I assume that the namespaces will be up-to-date when the component mounts?

AlanGreene: Namespaces are only retrieved on page load at the moment, but once we address https://github.com/tektoncd/dashboard/issues/91 they could be kept up to date as changes are made.

Do we expect to see namespaces created / deleted frequently?

ncskier: No, I don't think so. On page loads will definitely be good enough for this.

ncskier:
Hi @AlanGreene I added some tests and a storybook entry. Let me know what you think.

src/containers/NamespacesDropdown/NamespacesDropdown.stories.js
```diff
@@ -0,0 +1,66 @@
+/*
```
AlanGreene: You'll also need to update the storybook config so this gets loaded: https://github.com/tektoncd/dashboard/blob/15b946eb3334095a6cb52e70c9860e40363afc51/.storybook/config.js#L41

```diff
-const req = require.context('../src/components', true, /\.stories.js$/);
+const req = require.context('../src', true, /\.stories.js$/);
```

src/containers/NamespacesDropdown/NamespacesDropdown.stories.js
```diff
@@ -0,0 +1,66 @@
...
+*/
+
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import { Provider } from 'react-redux';
+import configureStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+
+// import store from '../../store';
```
AlanGreene: remove

src/containers/NamespacesDropdown/NamespacesDropdown.stories.js
```diff
@@ -0,0 +1,66 @@
...
+import NamespacesDropdown from './NamespacesDropdown';
+
+const props = {
+  items: ['default', 'kube-public', 'kube-system', 'tekton-pipelines'],
```
AlanGreene: `items` isn't needed here as the component will read the values from the store

src/containers/NamespacesDropdown/NamespacesDropdown.test.js
```diff
@@ -0,0 +1,111 @@
+// /*
```
AlanGreene: double comments should be removed

i.e. `// /*` -> `/*`

ncskier:
Thanks for the review @AlanGreene. Just made those changes.

PR #171, opened 2019-05-17
Add a ServiceAccounts dropdown component

ncskier:
# Changes

Adding a ServiceAccounts dropdown component that can be reused many places in our future UI.
This PR adds a new `serviceaccounts` endpoint to the dashboard with accompanying documentation in `DEVELOPMENT.md`. Backend and frontend test files are included as well as a storybook file.

I also updated the namespaces dropdown to have a default label "Select Namespace" to match the service accounts dropdown default label "Select ServiceAccount". The label will appear when nothing is selected, and as a prop it can be overwritten.

Thanks @jessm12 for the initial work on this component.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/actions/serviceAccounts.js
+++ b/src/actions/serviceAccounts.js
@@ -1,0 +1,37 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { getServiceAccounts } from '../api';
+import { getSelectedNamespace } from '../reducers';
+
+export function fetchServiceAccountsSuccess(data) {
+  return {
+    type: 'SERVICE_ACCOUNTS_FETCH_SUCCESS',
+    data
+  };
+}
+
+export function fetchServiceAccounts({ namespace } = {}) {
+  return async (dispatch, getState) => {
+    dispatch({ type: 'SERVICE_ACCOUNTS_FETCH_REQUEST' });
+    let serviceAccounts;
+    try {
+      const requestedNamespace = namespace || getSelectedNamespace(getState());
+      serviceAccounts = await getServiceAccounts(requestedNamespace);
+      dispatch(fetchServiceAccountsSuccess(serviceAccounts));
+    } catch (error) {
+      dispatch({ type: 'SERVICE_ACCOUNTS_FETCH_FAILURE', error });
+    }
+    return serviceAccounts;
+  };
+}
--- a/src/containers/ServiceAccountsDropdown/ServiceAccountsDropdown.js
+++ b/src/containers/ServiceAccountsDropdown/ServiceAccountsDropdown.js
@@ -1,0 +1,99 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from 'react';
+import { connect } from 'react-redux';
+
+import {
+  getServiceAccounts,
+  isFetchingServiceAccounts,
+  getSelectedNamespace
+} from '../../reducers';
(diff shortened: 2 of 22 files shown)


src/actions/serviceAccounts.js
```diff
@@ -0,0 +1,38 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { getServiceAccounts } from '../api';
+import { getSelectedNamespace } from '../reducers';
+
+export function fetchServiceAccountsSuccess(data, namespace) {
```
AlanGreene: We've made some changes recently to the action creators and reducers. These changes should also be made for the new serviceaccounts functionality, both for consistency, and to ensure correct behaviour.

- changes to support loading all namespaces https://github.com/tektoncd/dashboard/pull/170
  - remove `namespace` from the success action creators
  - update reducers to read namespace value from the resource metadata
- allow overriding the selected namespace https://github.com/tektoncd/dashboard/pull/164
  - update 'fetch' action creators to accept namespace as a parameter (for use by extensions)
- expose action creators to extensions (where appropriate)
  - https://github.com/tektoncd/dashboard/blob/master/src/containers/Extension/actions.js

ncskier: Thanks for the feedback Alan. I'll look over these and make the changes.

ncskier:
@AlanGreene thanks for the review. I've updated the action creators and reducers to support loading all namespaces. I also added the fetch serviceaccounts action creator to extensions.

Let me know if there's anything else I should change 🙂

src/containers/ServiceAccountsDropdown/ServiceAccountsDropdown.js
```diff
@@ -0,0 +1,67 @@
...
+      this.props.fetchServiceAccounts();
+    }
+  }
+
+  render() {
+    const { loading, ...dropdownProps } = this.props;
+    if (loading) {
+      return <DropdownSkeleton {...dropdownProps} />;
+    }
+    return <Dropdown {...dropdownProps} />;
```
AlanGreene: The Dropdown component truncates long options but does not display a tooltip. To allow users to correctly identify the options we made some additions to the NamespacesDropdown to add a `title` attribute on the options. That could should be refactored for reuse here.

https://github.com/tektoncd/dashboard/blob/fc9259d40e958dfe2d3f6bd8307c58a1aafaed10/src/containers/NamespacesDropdown/NamespacesDropdown.js#L22

see the `itemToElement` and `itemToString` props passed to the Dropdown component

ncskier: Ok, will do

ncskier:
Thanks Alan, I added the `title` attribute just like with the `NamespacesDropdown`

ncskier:
I updated the tests, so they cover the ServiceAccount Dropdown's `componentDidUpdate` function. @AlanGreene please let me know if there's anything else I can do for this PR 🙂

src/containers/ServiceAccountsDropdown/ServiceAccountsDropdown.test.js
```diff
@@ -0,0 +1,277 @@
...
+  green: '',
+  blue: ''
+};
+
+const initialTextRegExp = new RegExp('select serviceaccount', 'i');
+
+const middleware = [thunk];
+const mockStore = configureStore(middleware);
+
+jest
```
AlanGreene: `spyOn` and other setup should be called from inside an `it` to allow for proper cleanup between tests

if there's common code you want to execute in all of the tests you can use a `beforeEach`

src/containers/ServiceAccountsDropdown/ServiceAccountsDropdown.js
```diff
@@ -0,0 +1,119 @@
...
+import { Dropdown, DropdownSkeleton } from 'carbon-components-react';
+
+import {
+  getServiceAccounts,
+  isFetchingServiceAccounts,
+  getSelectedNamespace
+} from '../../reducers';
+import { fetchServiceAccounts } from '../../actions/serviceAccounts';
+
+const itemToElement = ({ text }) => {
```
AlanGreene: I could have been clearer in my previous comment, when I mentioned reuse I meant that the core of the namespaces dropdown should be refactored to extract the common pieces as a presentational component with the native tooltips for truncated text etc. Then both the namespaces dropdown and this new service accounts dropdown would build on top of that.

Ideally:
- Dropdown (the new presentational component) would
  - provide the loading state using DropdownSkeleton
  - provide these `itemToElement` and `itemToString` implementations by default to the Carbon component
  - automatically transform the `items` prop using `itemStringToObject`

- The `render` function of ServiceAccountsDropdown could look something like this:

```js
  render() {
    return (
      <Dropdown
        {...this.props}
        selectedItem={this.state.selectedItem}
        onChange={this.onChange}
      />
    );
  }
```
where `Dropdown` is the new component I described above

ncskier: Ah, that sounds like the right way to do it 😅 I'll make the change.

ncskier:
Thank you for the feedback @AlanGreene. I have addressed your two change requests, and the common tooltip dropdown code is now in a presentational `TooltipDropdown` component. Let me know what you think 🙂

PR #180, opened 2019-05-23
Issue 129 Pipeline_test.go refactor and related improvements

vtereso:
# Changes
Refactor pipeline_test.go and related components in pipeline.go

test-utils.go:
- New `dummyServer()` (*httptest.Server) convention over over dummyRestfulRequest/dummyRestfulResponse (Allow the router to do the work for us rather than mocking the functionality)

pipeline_test.go:
- Use loops to create resources rather than referencing variables for single CRD instances
- Using `dummyServer()` (*httptest.Server) convention
- Consolidate update/creation logic of pipelineRuns tests
- Other improvements

pipeline.go:
- Use uuid rather than int cast of time.Now() to ensure resources made within same second have unique names (don't fail due to collisions)
- Move getTaskRunLog core functionality into function (to be used for getPipelineRunLog)
- Change pipelineRunLog from plaintext to object and new creation

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/endpoints/pipeline_test.go
+++ b/pkg/endpoints/pipeline_test.go
@@ -14,750 +14,689 @@ limitations under the License.
 package endpoints
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
-	"io"
-	"net/http/httptest"
+	"net/http"
+	"reflect"
+	"strconv"
 	"strings"
 	"testing"
 
-	restful "github.com/emicklei/go-restful"
 	v1alpha1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Namespaces are specified on K8s CRD initialization (not client creation) to assert DeepEqual
+
 // Task test
 func TestTask(t *testing.T) {
-
-	r := dummyResource()
-
-	task1 := v1alpha1.Task{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "Task1",
-		},
-		Spec: v1alpha1.TaskSpec{},
-	}
-	task2 := v1alpha1.Task{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "Task2",
-		},
-		Spec: v1alpha1.TaskSpec{},
-	}
-	task3 := v1alpha1.Task{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "Task3",
-		},
-		Spec: v1alpha1.TaskSpec{},
+	server, r := dummyServer()
+	namespace := "ns1"
+	_, err := r.K8sClient.CoreV1().Namespaces().Create(&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}})
+	if err != nil {
+		t.Fatalf("Error creating namespace '%s': %v\n", namespace, err)
 	}
 
-	// Add sample tasks
-	_, err := r.PipelineClient.TektonV1alpha1().Tasks("ns1").Create(&task1)
-	if err != nil {
-		t.Logf("testtask error: %s", err)
+	tasks := []v1alpha1.Task{
+		v1alpha1.Task{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "Task1",
+				Namespace: namespace,
+			},
+			Spec: v1alpha1.TaskSpec{},
+		},
+		v1alpha1.Task{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "Task2",
+				Namespace: namespace,
+			},
+			Spec: v1alpha1.TaskSpec{},
+		},
+		v1alpha1.Task{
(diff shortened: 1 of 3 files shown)


vtereso:
Integration/unit tests are showing failures even though there are none?

vtereso:
In may be easier to just look at the [before](https://github.com/tektoncd/dashboard/blob/fc9259d40e958dfe2d3f6bd8307c58a1aafaed10/pkg/endpoints/pipeline_test.go) and [after](https://github.com/tektoncd/dashboard/blob/3697e14d02683b7a4bbe53dce0b466e23d10b7f0/pkg/endpoints/pipeline_test.go) of pipeline_test.go rather than the change sets

pkg/endpoints/pipeline_test.go
```diff
@@ -14,1408 +14,1039 @@ limitations under the License.
...
+				if response.StatusCode != expectedStatusCode {
+					t.Errorf("Response code %d did not equal expected %d\n", response.StatusCode, expectedStatusCode)
+				}
+			}
+		}
 	}
-
-	// Add sample TaskRuns
-	_, err := r.PipelineClient.TektonV1alpha1().TaskRuns("ns1").Create(&TaskRun1)
+	// Test updatePipelineRun function: bad request
```
mnuttall: This method currently does several different tests. Can they be split up into separate test methods?

pkg/endpoints/pipeline_test.go
```diff
@@ -14,1408 +14,1039 @@ limitations under the License.
...
-			Name: "Pod3",
-		},
-		Spec: corev1.PodSpec{
-			Containers:     []corev1.Container{Container7},
-			InitContainers: []corev1.Container{},
-		},
+	for _, pipelineRun := range pipelineRuns {
+		_, err := r.PipelineClient.TektonV1alpha1().PipelineRuns(namespace).Create(&pipelineRun)
+		if err != nil {
+			t.Errorf("Error creating pipelineRun '%s': %v\n", pipelineRun.Name, err)
```
mnuttall: I've taken to using t.Fatalf() because it stops the test run at that point, making it easier to use logs to work out what the first point of failure was. If a test has all ready failed, there's really no point in continuing. Would such a change fit into this overhaul?

pkg/endpoints/pipeline_test.go
```diff
@@ -14,1408 +14,1039 @@ limitations under the License.
...
-	httpReq := dummyHTTPRequest("GET", "http://wwww.dummy.com:8383/v1/namespaces/ns1/taskrunlogs/TaskRun1", nil)
-	req := dummyRestfulRequest(httpReq, "ns1", "TaskRun1")
-	httpWriter := httptest.NewRecorder()
-	resp := dummyRestfulResponse(httpWriter)
-
-	r.getTaskRunLog(req, resp)
-	var taskRunLog TaskRunLog
-	if err := json.NewDecoder(httpWriter.Body).Decode(&taskRunLog); err != nil {
-		t.Error(err)
+	// Test createPipelineRun function: good requests with/without pipelineResources
```
mnuttall: do tests with and without pipelineResources belong in separate test methods?

pkg/endpoints/pipeline_test.go
vtereso: There are definitely certain places I could definitely be using fatal. For example, if creating the prerequisite CRD for the test fails (shouldn't happen, but still), I could return fatal. However, the create/update pipelineRun logic I grouped together because before it seems hard to grasp since it was broken up and there was a lot of overlap (in the set up). Even prior, getOne/getAll functionality was grouped together and I thought it made sense that testing an individual function should be scope to an individual test. Everything could be replaced to fatal, but it would not allow you to get the response for the rest of the functionality (e.g. `getX passed for all scenarios except Y` vs `getX failed early`). In either scenario, a fail is a fail and it shouldn't be approved, but if you prefer that style I could replaceAll?

pkg/endpoints/pipeline_test.go
vtereso: Most of the ~300-400 lines cut down (not that we have to) were due to the setup being the same since the same method is being tested. In this way, you only have to look at one area rather that making sure all your changes are in the distributed tests for the single endpoint. I can break it back to how it was before if you prefer, but personally I find this way a bit easier. I understand the createPipelineRun good request for pipelineResources was a bit bloated, but besides that we didn't break up getOne/getAll in other locations so I went one step further.

pkg/endpoints/pipeline_test.go
vtereso: Addressed by below two responses

pkg/endpoints/pipeline_test.go
mnuttall: it's fine, I hadn't realised that you'd spent the time to merge them together.

pkg/endpoints/pipeline_test.go
mnuttall: no fair enough, let's take it as-is.

mnuttall:
Happy to take this PR as-is but I'd like a green build first. 
/retest

mnuttall:
/lgtm
Thanks @vtereso !

PR #182, opened 2019-05-24
Add panelling to Import Tekton resources from repository

jessm12:
# Changes

Add Pipeline0 and Pipeline0 task yaml definitions that are applied at install (ko apply) time, this pipeline applies all resources under the Pipeline's specified directory at the provided repository. Add the panel and associated tests to allow the user to input a repository URL and ServiceAccount and trigger a PipelineRun of Pipeline0 from this data. 

Below is a screenshot of the panel:

Screen Shot 2019-05-24 at 15 10 49

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/config/pipeline0-task.yaml
+++ b/config/pipeline0-task.yaml
@@ -1,0 +1,25 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: pipeline0-task
+spec:
+  inputs:
+    resources:
+    - name: git-source
+      type: git
+    params:
+    - name: pathToResourceFiles
+      description: The path to the resource files to apply
+      default: /workspace/git-source
+    - name: apply-directory
+      description: The directory from which resources are to be applied
+      default: ""
+  steps:
+  - name: kubectl-apply
+    image: lachlanevenson/k8s-kubectl
+    command:
+    - kubectl
+    args:
+    - apply
+    - -f
+    - ${inputs.params.pathToResourceFiles}/${inputs.params.apply-directory}
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -128,9 +128,9 @@ export function getTaskRunLog(name, namespace) {
   return get(uri);
 }
 
-export function createPipelineRun(payload) {
-  const uri = `${apiRoot}/`;
-  return post(uri, payload);
+export function createPipelineRun(params, namespace) {
+  const uri = getAPI('pipelineruns', { namespace });
+  return post(uri, params);
 }
 
 export function getCredentials(namespace) {
--- a/src/containers/ImportResources/ImportResources.js
+++ b/src/containers/ImportResources/ImportResources.js
@@ -1,0 +1,177 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React, { Component } from 'react';
+
+import {
+  ToastNotification,
+  TextInput,
+  Button,
+  Form
+} from 'carbon-components-react';
+import { connect } from 'react-redux';
+import { Link } from 'react-router-dom';
+
+import './ImportResources.scss';
+
+import { createPipelineRun } from '../../api';
+import ServiceAccountsDropdown from '../ServiceAccountsDropdown';
+import { getSelectedNamespace } from '../../reducers';
+
+export class ImportResources extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      repositoryURL: '',
+      serviceAccount: '',
+      submitSuccess: false,
+      invalidInput: false,
+      logsURL: '',
+      directory: ''
+    };
+  }
+
+  handleServiceAccount = data => {
+    this.setState({
+      serviceAccount: data.selectedItem.text
+    });
+  };
+
+  handleTextInput = event => {
+    const identifier = event.target.name;
+    const inputValue = event.target.value;
+    this.setState({
+      [identifier]: inputValue,
+      invalidInput: false
+    });
+  };
+
+  handleSubmit = event => {
+    const { namespace } = this.props;
+    const pipelinename = 'pipeline0';
+    const gitresourcename = 'git-source';
+    const gitcommit = 'master';
+    const applydirectory = this.state.directory;
+    const repourl = this.state.repositoryURL;
+    const serviceaccount = this.state.serviceAccount;
+    const payload = {
+      pipelinename,
+      serviceaccount,
+      gitresourcename,
+      gitcommit,
+      repourl,
+      applydirectory
+    };
+
+    if (repourl === '') {
+      this.setState({
+        invalidInput: true
+      });
+      return;
+    }
+
+    this.setState({
+      invalidInput: false
+    });
+
+    const promise = createPipelineRun(payload, namespace);
+    promise
+      .then(headers => {
+        const logsURL = headers.get('Content-Location');
+        const pipelineRunName = logsURL.substring(logsURL.lastIndexOf('/') + 1);
+        const finalURL = '/pipelines/pipeline0/runs/'.concat(pipelineRunName);
+        this.setState({
+          logsURL: finalURL,
+          submitSuccess: true,
+          invalidInput: false
+        });
+      })
+      .catch(error => {
+        const statusCode = error.response.status;
+        switch (statusCode) {
+          case 500:
+            this.setState({
+              submitSuccess: false,
+              invalidInput: true
+            });
+            break;
+          default:
+        }
+      });
+    event.preventDefault();
+  };
+
+  render() {
+    return (
+      <div className="outer">
+        <h1 className="ImportHeader">
+          Import Tekton resources from repository
+        </h1>
+        <Form>
+          <TextInput
+            required
+            type="URL"
+            invalid={this.state.invalidInput}
+            invalidText="Please submit a valid URL"
+            helperText="The location of the YAML definitions to be applied (Git URL's supported)"
+            labelText="Repository URL"
+            placeholder="Enter repository URL"
+            data-testid="repository-url-field"
+            name="repositoryURL"
+            value={this.state.repositoryURL}
+            onChange={this.handleTextInput}
+          />
+          <TextInput
+            helperText="The path from which resources will be applied at the
+          specified repository"
+            labelText="Repository directory (optional)"
+            placeholder="Enter repository directory"
+            data-testid="directory-field"
+            name="directory"
+            value={this.state.directory}
+            onChange={this.handleTextInput}
+          />
+          <ServiceAccountsDropdown
+            helperText="The SA that the PipelineRun applying resources will
+          run under"
+            titleText="Service Account (optional)"
+            className="saDropdown"
+            onChange={this.handleServiceAccount}
+          />
+          <Button kind="primary" onClick={this.handleSubmit}>
+            Import and Apply
+          </Button>
+          {this.state.submitSuccess && (
+            <ToastNotification
+              kind="success"
+              title="Triggered PipelineRun to apply Tekton resources"
+              subtitle=""
+              caption={
+                <Link to={this.state.logsURL}>View status of this run</Link>
+              }
+            />
+          )}
+        </Form>
+      </div>
+    );
+  }
+}
+
+/* istanbul ignore next */
+function mapStateToProps(state) {
+  return {
+    namespace: getSelectedNamespace(state)
+  };
+}
+
+export default connect(mapStateToProps)(ImportResources);
--- a/src/containers/ImportResources/ImportResources.scss
+++ b/src/containers/ImportResources/ImportResources.scss
@@ -1,0 +1,46 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+@import '~carbon-components/scss/globals/scss/vars';
+
+.ImportHeader {
+  margin-left: 2.5rem;
+  padding-top: 2rem;
+  font-size: 1.5rem;
+}
+
+.outer {
+  background-color: white;
+  min-height: 46rem;
+  .bx--dropdown.bx--list-box {
+    margin: 0rem;
+    margin-bottom: 2.8rem;
+    width: 30%;
+    border-bottom: 1px solid $ui-04;
+  }
+  .bx--toast-notification {
+    margin: 2.8rem;
+  }
+  .bx--dropdown__wrapper {
+    margin-left: 2.8rem;
+  }
+  .bx--btn {
+    margin-left: 2.8rem;
+  }
+  .bx--form-item {
+    margin: 2.8rem;
+  }
+  .bx--text-input__field-wrapper {
+    width: 50%;
+  }
+}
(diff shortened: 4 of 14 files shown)


package-lock.json
```diff
@@ -7319,7 +7319,8 @@
         "ansi-regex": {
```
AlanGreene: There shouldn't be any change to the package-lock as this PR isn't modifying dependencies.

src/api/index.js
```diff
@@ -128,9 +128,9 @@ export function getTaskRunLog(name, namespace) {
   return get(uri);
 }
 
-export function createPipelineRun(payload) {
-  const uri = `${apiRoot}/`;
-  return post(uri, payload);
+export function createPipelineRun({ id, ...rest }, namespace) {
```
AlanGreene: I don't see where the `id` would be provided, this should probably be removed.

There's also no need to split the params, you're not doing anything with them individually here, you're just passing them all through so you can do it directly:

```js
export function createPipelineRun(params, namespace) {
  const uri = getAPI('pipelineruns', { namespace });
  return post(uri, params);
}
```

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,221 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React, { Component } from 'react';
+
+// import Link from 'react-router-dom';
```
AlanGreene: remove

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,221 @@
...
+import {
+  ToastNotification,
+  TextInput,
+  FormLabel,
+  Button,
+  Tooltip
+} from 'carbon-components-react';
+import { connect } from 'react-redux';
+
+import '../../components/Definitions/Definitions.scss';
```
AlanGreene: Components shouldn't be loading styles directly from other components.

If we need to extract common styles or rewrite things to apply globally from index.scss we should do that.

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,221 @@
...
+    const payload = {
+      pipelinename,
+      serviceaccount,
+      gitresourcename,
+      gitcommit,
+      repourl,
+      applydirectory
+    };
+
+    console.log(payload);
```
AlanGreene: remove

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,221 @@
...
+    });
+    event.preventDefault();
+  }
+
+  render() {
+    this.handleServiceAccount = this.handleServiceAccount.bind(this);
+    this.handleTextInput = this.handleTextInput.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+
+    return (
```
AlanGreene: See comment above about Carbon style / layout guidelines for forms

src/containers/ImportResources/ImportResources.scss
```diff
@@ -0,0 +1,69 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+.bx--text-input__field-wrapper {
+  width: 100%;
+}
+
+.bx--label {
+  font-size: 1rem !important;
```
AlanGreene: Avoid using `!important` as it makes it very difficult to override if needed.

src/containers/ImportResources/ImportResources.scss
```diff
@@ -0,0 +1,69 @@
...
+}
+
+.bx--form-requirement {
+  margin-bottom: 0 !important
+}
+
+.saDropdown {
+  margin: 0% !important;
+  width: 100% !important;
+  border-bottom: 1px solid #8c8c8c !important;
```
AlanGreene: use design tokens where possible instead of hard-coding colours etc.
https://www.carbondesignsystem.com/guidelines/color/usage

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,221 @@
...
+          });
+          break;
+        default:
+      }
+    });
+    event.preventDefault();
+  }
+
+  render() {
+    this.handleServiceAccount = this.handleServiceAccount.bind(this);
```
AlanGreene: `bind` returns a new function, so you're replacing these on every render. A better place for this would be the constructor, however even that's not necessary. Instead you can bind the functions in-place by writing them as follows:

```js
class ImportResources extends Component {
  state = {
    ...
  };

  handleServiceAccount = data => {
    this.setState(...);
  }
}
```

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,221 @@
...
+      repourl,
+      applydirectory
+    };
+
+    console.log(payload);
+
+    const promise = createPipelineRun(payload, namespace);
+    promise.then(headers => {
+      const logsURL = headers.get('Content-Location');
+      const trimmedLogsURL = logsURL.substring(logsURL.lastIndexOf('/') + 1);
```
AlanGreene: Is this a pipeline run name? The variable name could reflect this to make it easier to follow the code.

We'll also need to be careful when constructing client URLs as we're currently reviewing changes in the approach / structure to align more closely with the kubernetes API structure. Anywhere we hardcode assumptions about the client URLs will need to be updated if we make that change.

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,221 @@
...
+      const finalURL = '#/pipelines/pipeline0/runs/'.concat(trimmedLogsURL);
+      this.setState({
```
AlanGreene: this should be combined with the `setState` below

src/containers/ImportResources/ImportResources.js
jessm12: I made this change and the handle code is working all good however my tests are now failing with `Cannot spy the handleTextInput property because it is not a function
; undefined given instead` does my code look ok to you and how can I fix this please?

AlanGreene: Ideally the test wouldn't need to spy on this function and you could just assert on some user-visible change in the component, e.g. value in the input has updated.

If however this is not possible, you'll have to go back to the old method of binding:

1. in the constructor:
`this.handleTextInput = this.handleTextInput.bind(this);`

2. handleTextInput
```
  handleTextInput(event) {
    const identifier = event.target.name;
    const inputValue = event.target.value;
    this.setState({
      [identifier]: inputValue
    });
  }
```
3. in the test you can then `spyOn(ImportResource.prototype, 'handleTextInput')` as you have already

src/containers/ImportResources/ImportResources.scss
jessm12: I had tried originally to use the design tokens but they didn't seem to be recognised, how do I use them/is there something else I should do for them to be recognised?

AlanGreene: `@import '~carbon-components/scss/globals/scss/vars';` at the top of the file you want to use them in

src/containers/ImportResources/ImportResources.js
jessm12: I updated the test to check for purely the change in text on the screen but I can't figure out now why it is not working, please could you take a look?

jessm12: @AlanGreene I am getting `Unable to find an element with the text: /Invalid URL here/i.` even though the html that is printed after this failure seems to show that the value has been updated correctly for that field, shouldn't this mean that the `getByText` for this value should succeed?

src/containers/ImportResources/ImportResources.scss
```diff
@@ -0,0 +1,56 @@
...
+*/
+
+@import '~carbon-components/scss/globals/scss/vars';
+
+.bx--text-input__field-wrapper {
+  width: 100%;
+}
+
+.saDropdown {
+  margin: 0rem !important;
```
AlanGreene: please remove `!important`, if you need to override existing styles use a more specific selector e.g. `.someParentElement .saDropdown`

units can be omitted for `0`

src/containers/ImportResources/ImportResources.test.js
```diff
@@ -0,0 +1,85 @@
...
+    } 
+  });
+
+  it('Valid data submit displays success notification ', async () => {
+    const headerKey = 'Content-Location'
+    const headers = {
+      get(headerKey) {
+        return 'fake-tekton-pipeline-run';
+      }
+     }
```
AlanGreene: indentation

src/containers/ImportResources/ImportResources.test.js
```diff
@@ -0,0 +1,85 @@
...
+    
+    jest.spyOn(API, 'createPipelineRun').mockImplementation(() => Promise.resolve(headers));
+    const { queryByText, getByTestId, getByText } = render(<Provider store={store} ><ImportResources.ImportResources /></Provider>);
+
+    const repoURLField = getByTestId('repository-url-field');
+    await fireEvent.change(repoURLField, { target: { value: 'URL' } });
```
AlanGreene: `fireEvent` is synchronous, no need for async/await

jessm12: if I remove all async/await's in my tests they start failing, is this unexpected behaviour?

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,181 @@
...
+  render() {
+    return (
+      <div className="Outer">
+        <h1 className="ImportHeader">
+          Import Tekton resources from repository
+        </h1>
+        <TextInput
+          required
+          invalid={this.state.required}
+          helperText="The location of the YAML definitions to be applied"
```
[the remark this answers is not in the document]
jessm12: The Pipeline uses a resource of type git, we have only been using GitHub URL's so I can specify that it must be a GitHub URL here

jessm12:
@AlanGreene in terms of feedback on an invalid URL I am not sure what kind of checks we should be doing and where/how to go about them 🤔 the API does some basic checking at the moment for slashes and this is when and error is returned but it is not a full check for a valid URL

AlanGreene:
We should display something useful to the user, e.g. 'unable to load repo' / 'repo not found' or if we have any useful info in the API response. Right now it seems like nothing is happening when clicking the button which is not good.

src/containers/ImportResources/ImportResources.test.js
AlanGreene: instead of using async/await (which is introducing a slight delay and just happens to be enough for the component to update) we should wait for the update to happen, something like this:

```js
  it('Valid data submit displays success notification ', () => {
    const headers = {
      get() {
        return 'fake-tekton-pipeline-run';
      }
    };

    jest
      .spyOn(API, 'createPipelineRun')
      .mockImplementation(() => Promise.resolve(headers));
    const { getByTestId, getByText } = render(
      <Provider store={store}>
        <ImportResources.ImportResources />
      </Provider>
    );

    const repoURLField = getByTestId('repository-url-field');
    fireEvent.change(repoURLField, { target: { value: 'URL' } });
    fireEvent.click(getByText('Import and Apply'));
    waitForElement(() =>
      getByText(/Triggered PipelineRun to apply Tekton resources/i)
    );
  });
```

verified locally that this passes consistently


I also noticed quite a few lint errors in that file, it'd be a good idea to `npm run lint` and address any that remain. `npm run lint:fix` will automatically fix any whitespace / layout related issues which usually account for the majority. I'll also take a look at why this isn't being picked up in the PR builds.

jessm12:
http://example.com reports invalid input for me when I try to submit as should I think any URL that isn't of the form `http://example.com/something/something`, Brandon has just mentioned to me that there is a `type` field that can be set to URL for some additional validation, would this be sufficient?

src/containers/ImportResources/ImportResources.test.js
jessm12: ok thanks Alan! I think it is that none of the test files are being linted, as when I am doing npm start locally it is doing all the correct linting for me in all files but not in the test files

jessm12: I am having some trouble with this linting error:
``` 
const headerKey = 'Content-Location';
const headers = {
  get(headerKey) {
    return 'fake-tekton-pipeline-run';
  }
};
```

it is reporting that header key is defined and not used when it is used here in my code
```
  35:9   error  'headerKey' is assigned a value but never used      no-unused-vars
  39:11  error  'headerKey' is already declared in the upper scope  no-shadow
  39:11  error  'headerKey' is defined but never used               no-unused-vars
```

AlanGreene: `headerKey` isn't used in that snippet so you could just have:

```js
const headers = {
  get() {
    return 'fake-tekton-pipeline-run';
  }
};
```

That'll resolve the linting errors and still does the exact same thing.

jessm12:
This is ready for another review when you have a chance to take a look please @AlanGreene 👼 

I think the only part of your comments that I haven't addressed here is setting the dropdown to be deactivated or to display some message when there are no ServiceAccounts, I think this should be done in the component so I can do this work separately from this issue

ncskier:
@AlanGreene @jessm12 I am addressing the empty dropdown message in PR https://github.com/tektoncd/dashboard/pull/201, which can go in after Jess's PR (I expect there will be a few merge conflicts)

config/pipeline0-task.yaml
```diff
@@ -0,0 +1,25 @@
...
+    params:
+    - name: pathToResourceFiles
+      description: The path to the resource files to apply
+      default: /workspace/git-source
+    - name: apply-directory
+      description: The directory from which resources are to be applied
+      default: ""
+  steps:
+  - name: kubectl-apply
+    image: lachlanevenson/k8s-kubectl
```
AlanGreene: Should this be pinned to a specific version?

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,176 @@
...
+          <Button kind="primary" onClick={this.handleSubmit}>
+            Import and Apply
+          </Button>
+          {this.state.submitSuccess && (
+            <ToastNotification
+              kind="success"
+              title="Triggered PipelineRun to apply Tekton resources"
+              subtitle=""
+              caption={
+                <a href={this.state.logsURL}> View status of this run </a>
```
AlanGreene: This still needs to be addressed https://github.com/tektoncd/dashboard/pull/182#discussion_r288507779

src/containers/ImportResources/ImportResources.scss
```diff
@@ -0,0 +1,53 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+@import '~carbon-components/scss/globals/scss/vars';
+
+.bx--text-input__field-wrapper {
```
AlanGreene: These `.bx-*` styles at the top level will affect the entire application, is that intended? If so they should be extracted to a _global.scss or similar that's imported from App.scss

config/pipeline0-task.yaml
jessm12: It should always be fine just to use the latest image here as all we are doing is a `kubectl apply`

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,174 @@
...
+            labelText="Repository URL"
+            placeholder="Enter repository URL"
+            data-testid="repository-url-field"
+            name="repositoryURL"
+            value={this.state.repositoryURL}
+            onChange={this.handleTextInput}
+          />
+          <TextInput
+            helperText="The path from which resources will be applied at the
+          specifed repository"
```
AlanGreene: https://github.com/tektoncd/dashboard/pull/182#discussion_r288507389

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,174 @@
...
+          />
+          <Button kind="primary" onClick={this.handleSubmit}>
+            Import and Apply
+          </Button>
+          {this.state.submitSuccess && (
+            <ToastNotification
+              kind="success"
+              title="Triggered PipelineRun to apply Tekton resources"
+              subtitle=""
+              caption={<a href={this.state.logsURL}>View status of this run</a>}
```
AlanGreene: https://github.com/tektoncd/dashboard/pull/182#discussion_r288507779

src/containers/ImportResources/ImportResources.js
```diff
@@ -0,0 +1,177 @@
...
+          <Button kind="primary" onClick={this.handleSubmit}>
+            Import and Apply
+          </Button>
+          {this.state.submitSuccess && (
+            <ToastNotification
+              kind="success"
+              title="Triggered PipelineRun to apply Tekton resources"
+              subtitle=""
+              caption={
+                <Link href={this.state.logsURL}>View status of this run</Link>
```
AlanGreene: `<Link to=...`

jessm12:
I think this is hopefully 🤞 ready to go now @AlanGreene 😅

Issue #67, opened 2019-04-23
Surface "tekton" tagged credentials in the dashboard


skaegi:
For resources in the cluster like credentials that are tagged as associated with Tekton should be surfaced in the UI. Design coming...

mnuttall:
Git credentials in particular need a `tekton.dev/git-n: https://github.com` type annotation. 

Docker secrets that are not of type kubernetes.io/dockerconfigjson also need annotating. 

- New panelling should allow users to create secrets (user/pass or token-based). 
- Docker secrets should be of type kubernetes.io/dockerconfigjson in the first instance. 
- Service accounts should be patched with Docker secrets.

mnuttall:
@dibbles is helping to coordinate design efforts in this area so please work with him to pick up the latest thinking.

carlos-logro:
I'm going to tackle this issue, can't assign it to me though!

carlos-logro:
#185

PR #188, opened 2019-05-24
issue67 - secrets management - add new secret modal - merge order: 1st

carlos-logro:
issue: #67

# Changes

This PR is focused on the add new secret modal which consists on a parent container & a couple of stateless components. Secret Modal receives as props the open variable responsible for showing or hiding the modal, namespaces array which is fetched when the container mounts & handleNew handler function that is used to close the modal when the user either clicks on the close button, X icon or outside the modal. In it state it stores all the variables used throughout the form  & an array that stores each form field that is currently invalid. The container is set to initial state when component is unmounted. It has a couple of handlers: 
1. handleSubmit is called when the user clicks the submit button in the modal & basically validates the data & if everything looks good, it fires the createSecret action with the postData & desired namespace as arguments & then closes the modal.
2. handleChange just validates the current input fields being changed (except the urls) & update its state & might even store the fields name if it was invalid. 
3. handleAnnotationChange validates the urls & maintains the correct format of the annotations state variable. 
4. handleAdd & handleRemove either add more url fields or remove them & will only work when the service account field is filled & valid. 

I decided to divide the form into three components & the reason behind this was primarily to avoid clutter, but also, if in the future some other type of secrets can be created where some other fields are required & others are not then it makes it easier to render by adding a little inline logic. Each component is passed its necessary state variables, handleChange handler & invalidFields array. Each field in the component is disabled expect the first one & will be enabled once the field before it is filled & validated. Last two components include a disabled prop which is used to know if the first field in those respective components should render disabled or not. Lastly, s worth mentioning that each field inputs, aside from some simple attributes it also has an invalid attribute that is true when that field’s name is stored in the invalidFields array)
1. UniversalFields consist of the secret name, desired namespace & accessTo which can either be docker or git. 
2. BasicAuthFields consist of username, password & service account. 
3. Annotations just consists of the urls which may be more than one.

When testing the container the only test case featured is if it renders blank & for each of the components i test to see whether it correctly renders blank, disabled & invalid fields . 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/components/SecretsModal/Annotations.js
+++ b/src/components/SecretsModal/Annotations.js
@@ -1,0 +1,69 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from 'react';
+import { TextInput } from 'carbon-components-react';
+import Add from '@carbon/icons-react/lib/add--alt/24';
+import Remove from '@carbon/icons-react/lib/misuse/24';
+import './SecretsModal.scss';
+
+const Annotations = props => {
+  const {
+    annotations,
+    handleChange,
+    handleAdd,
+    handleRemove,
+    invalidFields
+  } = props;
+
+  const annotationFields = annotations.map((annotation, index) => {
+    return (
+      <div className="annotationRow" key={`annotationRow${annotation.id}`}>
+        <TextInput
+          id={`annotation-label${index}`}
+          labelText=""
+          aria-label={`Annotation Label#${index}. This is the tag Tekton uses for its resources.`}
+          value={annotation.label}
+          placeholder="tekton.dev/{source}-0"
+          onChange={handleChange}
+          invalid={invalidFields.indexOf(`annotation-label${index}`) > -1}
+          autoComplete="off"
+        />
+        <div className="colon">:</div>
+        <TextInput
+          id={`annotation-value${index}`}
+          labelText=""
+          aria-label={`Annotation Value#${index}. This is the url for the given Tekton resource.`}
+          value={annotation.value}
+          placeholder="https://domain.com"
+          onChange={handleChange}
+          invalid={invalidFields.indexOf(`annotation-value${index}`) > -1}
+          autoComplete="off"
+        />
+      </div>
+    );
+  });
+
+  return (
+    <div className="annotations">
+      <div className="labelAndButtons">
+        <p className="label">Server URL:</p>
+        <Remove className="removeIcon" onClick={handleRemove} />
+        <Add className="addIcon" onClick={handleAdd} />
+      </div>
+      {annotationFields}
+    </div>
+  );
+};
+
+export default Annotations;
--- a/src/components/SecretsModal/Annotations.test.js
+++ b/src/components/SecretsModal/Annotations.test.js
@@ -1,0 +1,75 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from 'react';
+import { render } from 'react-testing-library';
+import Annotations from './Annotations';
+
+it('Annotations shows blank fields', () => {
+  const props = {
+    handleChange() {},
+    invalidFields: [],
+    annotations: [
+      { label: `tekton.dev/git-0`, value: '', id: 'aaaa' },
+      { label: `tekton.dev/git-1`, value: '', id: 'bbbb' },
+      { label: `tekton.dev/git-2`, value: '', id: 'cccc' }
+    ],
+    handleAdd() {},
+    handleRemove() {}
+  };
+  const { getAllByDisplayValue, getByText } = render(
+    <Annotations {...props} />
+  );
+
+  expect(getByText(/Server URL/i)).toBeTruthy();
+  expect(getAllByDisplayValue('').length).toEqual(3);
+  expect(getAllByDisplayValue(/tekton.dev\/git-/i).length).toEqual(3);
+});
+
+it('Annotations incorrect fields', () => {
+  const props = {
+    handleChange() {},
+    invalidFields: [
+      'annotation-label1',
+      'annotation-label2',
+      'annotation-value2'
+    ],
+    annotations: [
+      { label: `tekton.dev/git-0`, value: 'https://domain.com', id: 'aaaa' },
+      { label: `tekton.dev/git-1`, value: '', id: 'bbbb' },
+      { label: `tekton.dev/git-2`, value: 'https://domain.com', id: 'cccc' },
+      { label: `tekton.dev/git-3`, value: 'https://domain.com', id: 'dddd' }
+    ],
+    handleAdd() {},
+    handleRemove() {}
+  };
+  const { getByLabelText } = render(<Annotations {...props} />);
+
+  const annotationLabel0 = getByLabelText(/Annotation Label#0/i);
+  const annotationLabel1 = getByLabelText(/Annotation Label#1/i);
+  const annotationLabel2 = getByLabelText(/Annotation Label#2/i);
+  const annotationLabel3 = getByLabelText(/Annotation Label#3/i);
+  const annotationValue0 = getByLabelText(/Annotation Value#0/i);
+  const annotationValue1 = getByLabelText(/Annotation Value#1/i);
+  const annotationValue2 = getByLabelText(/Annotation Value#2/i);
+  const annotationValue3 = getByLabelText(/Annotation Value#3/i);
+
+  expect(annotationLabel0.getAttribute('data-invalid')).toBeFalsy();
+  expect(annotationLabel1.getAttribute('data-invalid')).toBeTruthy();
+  expect(annotationLabel2.getAttribute('data-invalid')).toBeTruthy();
+  expect(annotationLabel3.getAttribute('data-invalid')).toBeFalsy();
+  expect(annotationValue0.getAttribute('data-invalid')).toBeFalsy();
+  expect(annotationValue1.getAttribute('data-invalid')).toBeFalsy();
+  expect(annotationValue2.getAttribute('data-invalid')).toBeTruthy();
+  expect(annotationValue3.getAttribute('data-invalid')).toBeFalsy();
+});
--- a/src/components/SecretsModal/BasicAuthFields.js
+++ b/src/components/SecretsModal/BasicAuthFields.js
@@ -1,0 +1,62 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from 'react';
+import { TextInput } from 'carbon-components-react';
+import './SecretsModal.scss';
+
+const BasicAuthFields = props => {
+  const {
+    username,
+    password,
+    handleChange,
+    invalidFields,
+    serviceAccount
+  } = props;
+  return (
+    <>
+      <TextInput
+        id="username"
+        autoComplete="off"
+        placeholder="example@domain.com"
+        value={username}
+        labelText="Email:"
+        onChange={handleChange}
+        invalid={invalidFields.indexOf('username') > -1}
+      />
+      <TextInput
+        id="password"
+        autoComplete="off"
+        type="password"
+        value={password}
+        placeholder="********"
+        labelText="Password/Token:"
+        onChange={handleChange}
+        invalid={invalidFields.indexOf('password') > -1}
+      />
+      <TextInput
+        id="serviceAccount"
+        autoComplete="off"
+        type="serviceAccount"
+        value={serviceAccount}
+        placeholder="default"
+        labelText="Service Account:"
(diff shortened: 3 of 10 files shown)


carlos-logro:
/cc 

> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here (e.g. `I signed it!`) and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google.com/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Ftektoncd%2Fdashboard%2Fpull%2F188) for more info**.

I signed it!

src/components/SecretsModal/Annotations.js
```diff
@@ -0,0 +1,82 @@
...
+          labelText=""
+          key={`annotation${i}`}
+          value={annotations[i]}
+          placeholder=""
+          onChange={handleChange}
+          invalid={invalidFields.indexOf(`annotation${i}`) > -1}
+          autoComplete="off"
+          disabled={disabled}
+        />
+        <div key={`colon${i}`} className="colon">
```
AlanGreene: Do we need a separate div for a colon? Can we achieve the same thing via styles instead using `::after { content: ":"; }`?

src/components/SecretsModal/UniversalFields.js
```diff
@@ -0,0 +1,89 @@
...
+        placeholder="secret-name"
+        value={name}
+        labelText="Name:"
+        onChange={handleChange}
+        invalid={invalidFields.indexOf('name') > -1}
+        invalidText="Must be less than 563 characters, contain only lowercase alphanumeric character, . or -"
+        autoComplete="off"
+      />
+      <br />
+      <Select
```
AlanGreene: We have a NamespacesDropdown component to ensure consistency

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -0,0 +1,273 @@
...
+*/
+
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { Modal } from 'carbon-components-react';
+import UniversalFields from '../../components/SecretsModal/UniversalFields';
+import Annotations from '../../components/SecretsModal/Annotations';
+import BasicAuthFields from '../../components/SecretsModal/BasicAuthFields';
+import '../../components/SecretsModal/SecretsModal.scss';
+/* istanbul ignore next */
```
AlanGreene: shouldn't be needed here for imports

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -0,0 +1,273 @@
...
+import { createSecret } from '../../actions/secrets';
+import { fetchNamespaces } from '../../actions/namespaces';
+import { getNamespaces } from '../../reducers';
+
+/* istanbul ignore next */
+function validateK8sResources(value, id) {
```
AlanGreene: naming, maybe `isValid*` or similar?

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -0,0 +1,273 @@
...
+  if (value.trim() === '') return false;
+
+  const trimmed = value.trim();
```
AlanGreene: move above the if statement for reuse

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -0,0 +1,273 @@
...
+};
+
+export /* istanbul ignore next */ class SecretsModal extends Component {
+  constructor(props) {
+    super(props);
+    this.state = initialState;
+  }
+
+  componentDidMount() {
+    this.props.fetchNamespaces();
```
AlanGreene: shouldn't be necessary, namespaces are loaded globally by the nav and kept up to date via websocket events

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -0,0 +1,273 @@
...
+      name,
+      namespace,
+      username,
+      password,
+      annotations,
+      accessTo,
+      serviceAccount
+    } = this.state;
+
+    if (validateK8sResources(name, 'name')) postData.name = name;
```
AlanGreene: please use braces for these conditionals, it will help avoid future bugs if someone tries adding additional statements under these

src/components/SecretsModal/Annotations.js
```diff
@@ -0,0 +1,82 @@
...
+    invalidFields,
+    disabled
+  } = props;
+
+  const annotationFields = [];
+  for (let i = 0; i < annotations.length; i += 2) {
+    annotationFields.push(
+      <div className="annotationRow" key={i}>
+        <TextInput
+          id={`annotation${i}`}
```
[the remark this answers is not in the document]
carlos-logro: I'm using data-testid for testing that way I don't have to use `documents.getElementByID`

src/components/SecretsModal/Annotations.test.js
```diff
@@ -0,0 +1,87 @@
...
+import Annotations from './Annotations';
+
+it('Annotations shows blank fields', () => {
+  const props = {
+    disabled: true,
+    handleChange() {},
+    invalidFields: [],
+    annotations: [
+      `tekton.dev/git-0`,
+      '',
```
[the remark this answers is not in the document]
carlos-logro: the empty elements are the `url` portion of the annotation. testing to see if they output correctly.

src/components/SecretsModal/BasicAuthFields.js
```diff
@@ -0,0 +1,76 @@
...
+        data-testid="serviceAccount"
+        autoComplete="off"
+        type="serviceAccount"
+        value={serviceAccount}
+        placeholder="default"
+        labelText="Service Account:"
+        onChange={handleChange}
+        invalid={invalidFields.indexOf('serviceAccount') > -1}
+        disabled={
+          password.trim() === '' || invalidFields.indexOf('password') > -1
```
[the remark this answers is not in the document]
carlos-logro: if the username is invalid then the user won't be able to add a password which will consequently won't be able to add a service account. The inputs are disabled if the the input above it are invalid or empty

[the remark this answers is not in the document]
carlos-logro: if the username is invalid then the user won't be able to add a password which will consequently won't be able to add a service account. The inputs are disabled if the the input above it are invalid or empty

src/components/SecretsModal/Annotations.js
carlos-logro: ill try to see if it works that way. But the main reason behind the div was for accurate spacing between the other two inputs

src/components/SecretsModal/UniversalFields.js
carlos-logro: ill update it to use that dropdown

src/containers/SecretsModal/SecretsModal.js
carlos-logro: will do, left like this because of prettier 😂

src/components/SecretsModal/Annotations.test.js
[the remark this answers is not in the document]
AlanGreene: so these should have actual values that get rendered?

src/components/SecretsModal/BasicAuthFields.js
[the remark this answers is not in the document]
AlanGreene: consider:
1. enter valid username
2. enter valid password
3. delete username
4. service account dropdown still active despite invalid username

Relying on the user to interact with elements in a specific order without enforcing it is risky and opens up many additional cases that need to be handled. Normally that type of interaction is guided by a wizard / multi-page form to ensure validity at each step but that's probably overkill here.

Can we validate everything on submit or do we really need to process each input individually like this?

src/components/SecretsModal/Annotations.js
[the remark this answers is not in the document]
AlanGreene: Can we use `byLabelText` or one of the other queries on some user-facing attributes?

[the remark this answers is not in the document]
carlos-logro: if I add a label text each input element will display that same label text which will distort the whole design... can't we stick with data-testid?

src/components/SecretsModal/Annotations.test.js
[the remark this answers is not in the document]
carlos-logro: Screen Shot 2019-06-21 at 10 00 24 AM
they render disabled at first with the second input(url part) blank

src/components/SecretsModal/BasicAuthFields.js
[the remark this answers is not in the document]
carlos-logro: I see what you say & you're correct. I can change the code to disable all next inputs until the current one i correct or just remove all that logic & leave the validation at the submit part of the form.

src/components/SecretsModal/Annotations.js
```diff
@@ -0,0 +1,82 @@
...
+    handleChange,
+    handleAdd,
+    handleRemove,
+    invalidFields,
+    disabled
+  } = props;
+
+  const annotationFields = [];
+  for (let i = 0; i < annotations.length; i += 2) {
+    annotationFields.push(
```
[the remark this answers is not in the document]
carlos-logro: using map in this case is kinda difficult since I'm iterating every two elements in the array instead of one because at each iteration im creating one row of annotation composed of 2 text inputs which with the `i` im able to access the element needed for the respective input.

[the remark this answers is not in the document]
AlanGreene: Do we control this data structure? If so I'd recommend changing it so we have an array of objects instead, each object could have the label + value required

src/components/SecretsModal/Annotations.js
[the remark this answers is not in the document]
AlanGreene: Each input must have some sort of label / description (label, title, aria-label, aria-labelledby, etc.) for accessibility purposes, otherwise screen reader users would get a generic announcement similar to "input, type text to edit" and no indication what the field relates to. Whatever mechanism is used does not necessarily have to be visible on the screen, but may be useful to other users anyway.

src/components/SecretsModal/Annotations.test.js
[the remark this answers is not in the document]
AlanGreene: see comment above re. accessibility
as a user looking at this it's not clear to me what value I should put in the second field, and whether I can put some arbitrary value in the first field or must use an incrementing number for each etc.

src/containers/SecretsModal/SecretsModal.js
AlanGreene: Still quite a few conditionals missing braces, makes the code much harder to scan especially when combined with nesting. The easiest way to address this is to add the following rule to the linter and run `npm run lint:fix`

```diff
diff --git a/.eslintrc.js b/.eslintrc.js
index 62c900d..b89061e 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,7 @@ module.exports = {
       "react"
   ],
   "rules": {
+    "curly": ["error", "all"],
     "import/prefer-default-export": "off",
     "jsx-a11y/anchor-is-valid": "off",
     "no-case-declarations": "off",
```

src/components/SecretsModal/Annotations.js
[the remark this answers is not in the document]
carlos-logro: yeah I have it to be stored as an array of strings but I can change it to that.

src/components/SecretsModal/Annotations.js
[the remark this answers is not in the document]
carlos-logro: true, I'll add aria-labels & descriptions & see if I can test them with those instead of the testid

src/containers/SecretsModal/SecretsModal.js
carlos-logro: ok ill add it!!

src/components/SecretsModal/Annotations.js
```diff
@@ -0,0 +1,71 @@
...
+    annotations,
+    handleChange,
+    handleAdd,
+    handleRemove,
+    invalidFields
+  } = props;
+
+  const annotationFields = annotations.map((annotation, index) => {
+    return (
+      <div className="annotationRow" key={`annotationRow(${index.toString()})`}>
```
AlanGreene: `index` should not be used for the key, instead you should use some attribute of the data that's unique, the label for example

src/components/SecretsModal/Annotations.js
```diff
@@ -0,0 +1,71 @@
...
+        <TextInput
+          id={`annotation-label${index}`}
+          labelText=""
+          aria-label={`Annotation Label#${index}`}
+          aria-describedby="This is the tag Tekton uses for its sources."
```
AlanGreene: `aria-describedby` should be the id of an element containing the description.

Does 'Annotation Label#0' provide enough information to the user about the field's purpose and expected value?

src/components/SecretsModal/BasicAuthFields.js
```diff
@@ -0,0 +1,67 @@
...
+    password,
+    handleChange,
+    invalidFields,
+    serviceAccount
+  } = props;
+  return (
+    <>
+      <TextInput
+        id="username"
+        aria-label="Username"
```
AlanGreene: this input already has a label, there shouldn't be a need for an `aria-label`

src/components/SecretsModal/Annotations.js
carlos-logro: if the label or value for that annotation is used then when the user changes one of those fields it will unfocus after one character change which is quite annoying. ill see what I can do

src/components/SecretsModal/Annotations.js
carlos-logro: at the moment that doest provide that information... ill add more detail to it & eliminate the aria-describedby

src/components/SecretsModal/BasicAuthFields.js
```diff
@@ -0,0 +1,63 @@
...
+        autoComplete="off"
+        placeholder="example@domain.com"
+        value={username}
+        labelText="Email:"
+        onChange={handleChange}
+        invalid={invalidFields.indexOf('username') > -1}
+      />
+      <TextInput
+        id="password"
+        aria-describedby="This is the url for the given Tekton source."
```
AlanGreene: `aria-describedby` not needed here since the input already has a label
also, looks like this had the wrong value anyway

carlos-logro: ooopss! :sweat_smile:

PR #189, opened 2019-05-24
install dashboard on minishift/openshift

Gl4di4torRr:
# Changes
Adding service, route, image stream and build config to simplify the process to install the dashboard.

# Submitter Checklist

review them:

- [ x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ y] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

for more details._

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,13 @@ $ docker push <mydockername>/dashboard:<mytag>
 $ kubectl apply -f config
 ```
 
+## Install on Minishift
+
+1. Install [tektoncd-pipeline-operator](https://github.com/openshift/tektoncd-pipeline-operator#deploy-openshift-pipelines-operator-on-minikube-for-testing)
+2. [Checkout](https://github.com/tektoncd/dashboard/blob/master/DEVELOPMENT.md#checkout-your-fork) the repository
+3. Install deployment config `$oc process -f config/templates/deploy.yaml | oc apply -f-`
+4. Install build config `$oc process -f build.yaml | oc apply -f-`
+
 ## Accessing Dashboard
 The Dashboard has can be accessed through its ClusterIP Service by running `kubectl proxy`. Assuming default is the install namespace for the dashboard, you can access the web UI at `localhost:8001/api/v1/namespaces/default/services/tekton-dashboard:http/proxy/`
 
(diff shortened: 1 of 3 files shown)


config/tekton-dashboard-deployment.yaml
```diff
@@ -1,42 +1,120 @@
-apiVersion: apps/v1
-kind: Deployment
+apiVersion: v1
+kind: Template
```
mnuttall: Hi @Gl4di4torRr this file is currently being used as a Deployment for those applying the changes directly to their local environment, such as Docker Desktop, minikube etc. I'm concerned that changing this from `kind: Deployment` to `kind: Template` is going to break a number of existing users. Would it be possible to provide the Template as a separate config element, perhaps in its own file?

Gl4di4torRr: @mnuttall maybe there should be a split between defining specific object types and templates? I can created a folder `config/templates` and place my template files there and leave the deployment object alone for the sake of existing users.

config/tekton-dashboard.yaml
```diff
@@ -0,0 +1,59 @@
+# ------------------- Dashboard Deployment ------------------- #
```
mnuttall: I'm being a bit fussy now but the Deployment section of this file shouldn't be showing any changes. Is this just git getting confused or have you moved and recreated the file?

config/tekton-dashboard-deployment.yaml
mnuttall: Thank you!

config/tekton-dashboard.yaml
mnuttall: Maybe I had an old copy of this - it's looking much cleaner now.

mnuttall:
I think we're good to go once you've got your conflicts resolved and changes squashed down to a single commit ... because Tekton uses Prow, we don't get the glorious 'squash and commit' button, and have to do it ourselves : (

Gl4di4torRr:
@mnuttall okay, i think i got everything cleaned up and squashed the way you want it.
Interesting fact about Prow - i did not know that. Was wondering why you were asking me to squash commits manually :)

mnuttall:
/approve
/lgtm
Thank you!

Issue #190, opened 2019-05-27
Add navigation to view task and pipeline definitions


steveodonovan:
# Expected Behavior

The dashboard should contain a view of pipeline and task resources

# Actual Behavior

This navigation does not currently exist

PR #191, opened 2019-05-27
Task and pipeline definition view

steveodonovan:
# Changes

https://github.com/tektoncd/dashboard/issues/190

add view for task and pipeline crds

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/CustomResourceDefinition/CustomResourceDefinition.js
+++ b/src/containers/CustomResourceDefinition/CustomResourceDefinition.js
@@ -1,0 +1,115 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import {
+  CodeSnippetSkeleton,
+  InlineNotification
+} from 'carbon-components-react';
+
+import jsYaml from 'js-yaml';
+
+import { fetchTask } from '../../actions/tasks';
+import { fetchPipeline } from '../../actions/pipelines';
+
+import {
+  getPipeline,
+  getSelectedNamespace,
+  getTask,
+  getTasksErrorMessage,
+  getPipelinesErrorMessage
+} from '../../reducers';
+
+export /* istanbul ignore next */ class CustomResourceDefinition extends Component {
+  state = {
+    loading: true
+  };
+
+  componentDidMount() {
+    const { match } = this.props;
+    const { name, type } = match.params;
+    this.fetch(type, name).then(() => this.setState({ loading: false }));
+  }
+
+  componentDidUpdate(prevProps) {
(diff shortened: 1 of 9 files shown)


src/containers/CustomResourceDefinition/CustomResourceDefinition.js
```diff
@@ -0,0 +1,114 @@
...
+    const { name, type } = match.params;
+    const { match: prevMatch, namespace: prevNamespace } = prevProps;
+    const { name: prevName, type: prevType } = prevMatch.params;
+    if (namespace !== prevNamespace || name !== prevName || type !== prevType) {
+      this.setState({ loading: true }); // eslint-disable-line
+      this.fetch(type, name).then(() => this.setState({ loading: false }));
+    }
+  }
+
+  fetch = async (type, name) => {
```
AlanGreene: `async` not needed here

src/containers/CustomResourceDefinition/CustomResourceDefinition.js
```diff
@@ -0,0 +1,114 @@
...
+    switch (type) {
+      case 'tasks':
+        return this.props.fetchTask({ name });
+      case 'pipelines':
+        return this.props.fetchPipeline({ name });
+      default:
+        return Promise.resolve();
```
AlanGreene: This is fine for now but when we remove the dependency on knowing the resource type we may need to revisit this for error handling.

src/containers/CustomResourceDefinition/CustomResourceDefinition.js
```diff
@@ -0,0 +1,114 @@
...
+      return (
+        <InlineNotification
+          kind="error"
+          title="Error loading resource"
+          subtitle={JSON.stringify(error)}
+        />
+      );
+    }
+    return (
+      <div className="bx--snippet--multi">
```
AlanGreene: We could use the `CodeSnippet` component here

src/containers/CustomResourceDefinition/CustomResourceDefinition.js
```diff
@@ -0,0 +1,114 @@
...
+    );
+  }
+}
+
+/* istanbul ignore next */
+function mapStateToProps(state, props) {
+  return {
+    namespace: getSelectedNamespace(state),
+    error: getTasksErrorMessage(state) || getPipelinesErrorMessage(state),
+    tasks: getTask(state, { name: props.match.params.name }),
```
AlanGreene: Can we check props.type and only load the relevant error / resource here? (avoids unnecessary re-renders)

src/containers/Pipelines/Pipelines.js
```diff
@@ -91,6 +94,11 @@ export /* istanbul ignore next */ class Pipelines extends Component {
                           {pipelineName}
                         </Link>
                       </StructuredListCell>
+                      <StructuredListCell>
+                        <Link to={`/pipelines/${pipelineName}`}>
```
AlanGreene: we should have a tooltip on this to identify its purpose, also helpful for accessibility

same for Tasks

src/containers/CustomResourceDefinition/CustomResourceDefinition.js
steveodonovan: Tried. Code snippet has the not optional and not configurable see more. It looks awful, although if you can see a way to sort the initial length it would be nice but it looked hardcoded to me.

AlanGreene: Had a look into the code and you're right, it's hard-coded to a max height of 255.
Let's stick with the current approach for now and we may consider either creating our own code box or contributing some enhancements to Carbon if we decide to revisit.

Issue #166, opened 2019-05-17
Add support for 'all namespaces' option


AlanGreene:
Current UI only displays resources from a single namespace at a time, controlled by the namespace dropdown in the side nav.

We should add an 'all namespaces' option to the dropdown so that a user can choose to view all resources of a given type (e.g. Pipeline, TaskRun, etc.) defined in the cluster regardless of namespace.

- Option 1:
  - front end makes an API call per namespace to get the requested resources
  - will need to handle loading / error states for multiple in-flight requests for a single resource type (not currently supported)
- Option 2:
  - back end API provides support for retrieving resources from all namespaces, perhaps using a reserved namespace value (e.g. '*') or by some other means
  - for example, looking at the approach taken by the kube API server (accessed via kube proxy below)
    - get pipelines for a particular namespace:
      - http://localhost:8001/apis/tekton.dev/v1alpha1/namespaces/default/pipelines
    - get pipelines for all namespaces (drop the `/namespaces/{name}`)
      - http://localhost:8001/apis/tekton.dev/v1alpha1/pipelines

Regardless of which option we pursue there are some front end changes that will need to be made:
- [x] 'success' actions should no longer include the namespace directly in the payload as results could come from multiple namespaces
- [x] reducers should read the namespace directly from the resource's metadata when processing

steveodonovan:
Backend api will be `.../v1/namespaces/*/<resource>` 

However one issue here `.../v1/namespaces/*/<resource>/<resourceName>`  will be invalid. A namespace must be provided when requesting a specific resource by name. The above will return an 404 and the message `an empty namespace may not be set when a resource name is provided` This is coming from the pipelineClient, not our application. Queryparameter filtering will still work though `/pipelineRuns?name=pipelinename` for example.

AlanGreene:
Updated description to include example of how this is handled by the kube API server.

vtereso:
The all namespaces option should be specific to get all CRD routes?

vtereso:
I favor anything except for the additional routes where we `(drop the /namespaces/{name})`

vtereso:
Updates @steveodonovan @AlanGreene ?

AlanGreene:
@vtereso we're going with the `*` approach for now to unblock the front end work and to avoid making multiple requests from the browser. https://github.com/tektoncd/dashboard/pull/174

ncskier:
[...]

PR #196, opened 2019-05-28
Update NamespacesDropdown to support 'All Namespaces' option

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/166

- Update TooltipDropdown to accept object items
- Update NamespacesDropdown to add support for an 'All Namespaces' option
- Update use of the NamespacesDropdown to use the item id for the selected item

This is not used yet but is required as part of the efforts to support displaying resources from all namespaces as well as updating the client URL to include namespace.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/components/TooltipDropdown/TooltipDropdown.js
+++ b/src/components/TooltipDropdown/TooltipDropdown.js
@@ -14,9 +14,9 @@ limitations under the License.
 import React from 'react';
 import { Dropdown, DropdownSkeleton } from 'carbon-components-react';
 
-const itemToElement = ({ text }) => {
+const itemToElement = ({ id, text }) => {
   return (
-    <div key={text} title={text}>
+    <div key={id} title={text}>
       {text}
     </div>
   );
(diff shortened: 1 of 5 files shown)


src/components/TooltipDropdown/TooltipDropdown.js
```diff
@@ -14,23 +14,29 @@ limitations under the License.
...
+    <div key={id} title={text}>
       {text}
     </div>
   );
 };
 
 const itemToString = ({ text }) => text;
 
-const itemStringToObject = text => ({ text });
+const itemToObject = item => {
```
steveodonovan: Would making all keys objects be better or overkill ?

src/containers/NamespacesDropdown/NamespacesDropdown.js
```diff
@@ -21,16 +21,33 @@ const NamespacesDropdown = props => {
...
 
+const allNamespacesLabel = 'All Namespaces';
+
 NamespacesDropdown.defaultProps = {
+  allNamespacesLabel,
   items: [],
   loading: true,
-  label: 'Select Namespace'
+  label: 'Select Namespace',
+  showAllNamespaces: false
```
steveodonovan: Why is this field needed ?

src/components/TooltipDropdown/TooltipDropdown.js
AlanGreene: `item` can be either a string or an object. If it's an object we just return it as-is, if it's a string we transform it to be an object with the expected shape.

For most of our current use cases using strings is simpler, as the display value and the 'id' are the same. This change gives us flexibility where we want to override the 'id' (such as with the '*' for all namespaces).

src/containers/NamespacesDropdown/NamespacesDropdown.js
AlanGreene: We only want to show 'All Namespaces' in the nav dropdown. Other places the namespace dropdown is used should not display this option.

PR #207, opened 2019-06-04
Reflect unexported value in tektoncd/pipeline

vtereso:
Reflect new value for container prefix in Tekton from recent [changes](https://github.com/tektoncd/pipeline/commit/3d87e2e18174f1d7b088bdd14065d5f06ef86462)
# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/endpoints/credentials_test.go
+++ b/pkg/endpoints/credentials_test.go
@@ -331,7 +331,7 @@ func readAllCredentialsTest(namespace string, expectCreds []credential, expectEr
 	}
 
 	// Verify against K8s client
-	if namespace == "*"{
+	if namespace == "*" {
 		namespace = ""
 	}
 	testCredentials(r.getK8sCredentials(namespace), expectCreds, t)
--- a/pkg/endpoints/pipeline.go
+++ b/pkg/endpoints/pipeline.go
@@ -96,10 +96,7 @@ type PipelineRunUpdateBody struct {
 type PipelineRunLog []TaskRunLog
 type TaskRunLog struct {
 	PodName string
-	// Containers correlating to Task step definitions
 	StepContainers []LogContainer
-	// Additional Containers correlating to Task e.g. PipelineResources
-	PodContainers  []LogContainer
 	InitContainers []LogContainer
 }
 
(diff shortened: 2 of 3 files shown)


vtereso:
> This also requires a change to the front end code. See
> 
> https://github.com/tektoncd/dashboard/blob/4ba7af99119e535bf6273d05e3fbc6684dd9cfb0/src/containers/Log/Log.js#L43
> and associated test

Sounds good. I am going to propose a PR against Tekton to see if they will approve exporting these variables. This way (although still unfortunate), the only break would be the on the UI side.

vtereso:
From the points mentioned [against exporting the variables](https://github.com/tektoncd/pipeline/pull/951), we don't need to be using the prefix so I have taken it out and simplified the logic. @AlanGreene take a look and let me know if this same can be done on the UI side.

AlanGreene:
@vtereso the front end code needs to know the full container name to retrieve the correct logs. We cannot do a substring match based on the step name as there could be multiple steps matching the same substring.

vtereso:
> @vtereso the front end code needs to know the full container name to retrieve the correct logs. We cannot do a substring match based on the step name as there could be multiple steps matching the same substring.

However IIUC, conflicting step names are unresolved through the static prefix? I do not know if Tekton enforces unique step names.

vtereso:
The logging structs are as follows:
```
type PipelineRunLog []TaskRunLog
type TaskRunLog struct {
	PodName string
	StepContainers []LogContainer
	InitContainers []LogContainer
}
```
I removed the podContainers since this is was only populated by the nop container. While addressing this, would it be advantageous to have the taskRun and pipelineRun names in the above?

AlanGreene:
Apologies for the delayed response. As you're aware there have been a number of changes / discussions across the projects on this topic and I had intended on following up once the dust settled but it slipped my mind.

As part of adding support for real time updates in the front end we have updated the log component to poll for log content. The taskrun logs API is too heavy for this purpose as we only need the content for a single step at a time. For this reason the front end has switched to using the pod logs API, specifying the relevant container for the selected step.

Once tektoncd/pipeline 0.5 is released we'll update the code on the front end to use the new `step-` prefix (previously `build-step-`). Unless we have a particular need for the taskrun log API elsewhere I'd suggest we remove it.

PR #213, opened 2019-06-05
Update client URLs to include namespace

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/166

A user should be able to bookmark / share links to individual resources
displayed in the dashboard. Before this change that was not always possible
as the URL did not include the selected namespace. If a namespace other
than 'default' was selected, the user would need to manually select the
correct namespace from the dropdown in the side navigation to view the
resource referenced by the URL.

Update the URL for single Tekton resources to include the namespace so
that a URL will point to the correct view regardless of which namespace
was selected at the time the URL was copied.

Also provide namespace scoped URLs for the list views.
Link URL and namespace dropdown so they're kept in sync.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/SideNav/SideNav.js
+++ b/src/containers/SideNav/SideNav.js
@@ -1,0 +1,160 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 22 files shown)


src/containers/SideNav/SideNav.js
```diff
@@ -0,0 +1,147 @@
+import React, { Component } from 'react';
```
steveodonovan: CR

AlanGreene:
@steveodonovan thanks for the review, fixed the missing CR notice and squashed. Should be good to go.

PR #214, opened 2019-06-05
Handle PipelineRun that failed to start

jessm12:
# Changes

When a PipelineRun fails to start due to an error, for example it has a missing PipelineResource, render a message to say that the details could not be retrieved due to an error and display this error on the PipelineRun details panel as opposed to the current behaviour of displaying a blank screen
# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/PipelineRun/PipelineRun.js
+++ b/src/containers/PipelineRun/PipelineRun.js
@@ -82,13 +82,12 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
     const {
       status: { taskRuns: taskRunsStatus }
     } = pipelineRun;
-    const { message, status } = getStatus(pipelineRun);
-    if (status === 'False' && !taskRunsStatus) {
-      throw message;
-    }
-    const taskRunNames = Object.keys(taskRunsStatus);
-
-    return { pipelineRun, taskRunNames };
+    const { message, status, reason } = getStatus(pipelineRun);
+    return {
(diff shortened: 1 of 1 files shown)


jessm12:
Thanks @AlanGreene 😄 something similar to this is what I had wanted to do but wasn't sure how to go about it! I will work from your code now

jessm12:
please take a look @AlanGreene 🙂 

Screenshot 2019-06-06 at 11 22 55

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -165,7 +164,36 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
...
+          <RunHeader
+            lastTransitionTime={lastTransitionTime}
+            loading={loading}
+            runName={pipelineRunName}
+            reason="Error"
+            status={pipelineRunStatus}
+          />
+          <InlineNotification
+            kind="error"
+            title="Unable to retrieve PipelineRun logs"
```
AlanGreene: This title is a little misleading, the actual error described is that the PipelineRun couldn't be created.

Maybe we should go with something more generic like 'Unable to load PipelineRun details' as there are likely other scenarios that can result in the same state. We could also include `pipelineRunError.reason` for additional info.

jessm12: Sounds good I will make that change 👍

Issue #195, opened 2019-05-28
Carbon Design System Modal is not working properly with the dashboard UI Shell


ncskier:
# Expected Behavior
Here is the codesandbox link to a primitive example of what is expected from the Modal component: https://codesandbox.io/embed/carboncomponentsreact-7zzqn
![Screen Shot 2019-05-28 at 2 56 41 PM](https://user-images.githubusercontent.com/16961496/58504538-d4466880-8158-11e9-91b6-6d443b97e73d.png)

# Actual Behavior
There are a few things that went wrong when I tried using the Modal component in the dashboard code: 

#### 1. The header and side nav areas are not grayed out, and can be clicked without dismissing the modal view.
![Screen Shot 2019-05-28 at 2 46 13 PM](https://user-images.githubusercontent.com/16961496/58503964-a90f4980-8157-11e9-9066-39fc56830b26.png)

#### 3. When scrolling up/down the page, the modal view moves with the page. Instead, the modal view should have a fixed position, and only the content behind it should scroll.

# Steps to Reproduce the Problem

For reference, here is my branch where I mimicked the codesandbox code in the dashboard: https://github.com/ncskier/dashboard/tree/sandbox. I added a page at `http://localhost:8000/#/sandbox`.

# Additional Info
FYI @AlanGreene

AlanGreene:
I can reproduce the same issue in a codesandbox once I load the styles for the UI Shell header + content: https://codesandbox.io/s/carboncomponentsreact-vxg18

I'd suggest checking the Carbon repo / slack to see if this is a known issue or if there's a suggested solution.

carlos-logro:
After doing some testing here is what I have concluded:

1. The header and side nav areas are not grayed out, and the only way to make this happen is if the modal is opened within the App or index.html file. 

2. The modal is centered in the entire Content element when it should be centered in the viewport. To fix this i had to specify a viewport height by either 88.6vh or calc(100% - 7rem) 
*7rem being the height of the header*

3. When scrolling up/down the page, the modal view moves with the page. Like in the 1st answer, the only way to avoid this is by opening the modal from the App or index.html file

asked in `#carbon-react` to see what they suggest on doing for the 3rd one.

carlos-logro:
[...]

PR #217, opened 2019-06-06
issue195 - removed transform value so modals display properly

carlos-logro:
issue: #195 

# Changes

Before this change we had to hardcode styles so modals fit properly when displayed but a weird bug was happening that it was not fixed in the dashboard & disappeared when one was scrolling through the page. This was easily fixed by removing the transform value.

### DIFF
--- a/src/components/App/App.scss
+++ b/src/components/App/App.scss
@@ -68,5 +68,6 @@ main {
 
   .bx--content {
     margin-top: 7rem;
+    transform: none;
   }
 }


carlos-logro:
@steveodonovan @AlanGreene how can i make the integration-test pass?? its just one css line

AlanGreene:
@carlos-logro wait a while and retest, looks like the test infra is short on resources at the moment. You probably just hit it at a particularly busy time.

AlanGreene:
/lgtm
/approve

If this becomes a problem (janky scrolling / animation for example) we can revisit but all looks smooth now.

PR #220, opened 2019-06-06
Add namespaces dropdown to import resources view

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/166#issuecomment-499649129

The import resources view was implemented without a namespace input.
Instead it relied on the globally selected namespace. When the
'all namespaces' option was introduced recently this lead to the
scenario where importing would fail if 'all namespaces' was selected
in the global nav.

Add a namespaces dropdown to the import page, defaulting to the
selected namespace from global nav (if not 'all namespaces'). This
value will be used for the API call instead of the value from the nav.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/ImportResources/ImportResources.js
+++ b/src/containers/ImportResources/ImportResources.js
@@ -22,24 +22,31 @@ import {
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
 
-import './ImportResources.scss';
-
+import { ALL_NAMESPACES } from '../../constants';
 import { createPipelineRun } from '../../api';
-import ServiceAccountsDropdown from '../ServiceAccountsDropdown';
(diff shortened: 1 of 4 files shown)


src/containers/ImportResources/ImportResources.js
```diff
@@ -57,32 +64,38 @@ export class ImportResources extends Component {
...
+      applydirectory,
       gitcommit,
+      gitresourcename,
+      pipelinename,
       repourl,
-      applydirectory
+      serviceaccount
     };
 
+    let hasError = false;
```
steveodonovan: get rid of has error and put 2 returns instead ?

AlanGreene: updated to use a single setState + return

PR #270, opened 2019-06-19
Add type prop to PipelineResourcesDropdown

ncskier:
# Changes

This adds a `type` prop to the `PipelineResourcesDropdown` so that it will only display resources of one type; for example, `type="git"` will only display `git` resources in the dropdown.
# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/PipelineResourcesDropdown/PipelineResourcesDropdown.js
+++ b/src/containers/PipelineResourcesDropdown/PipelineResourcesDropdown.js
@@ -37,11 +37,15 @@ class PipelineResourcesDropdown extends React.Component {
   }
 
   render() {
-    const { namespace, ...rest } = this.props;
-    const emptyText =
-      namespace === ALL_NAMESPACES
-        ? `No Pipeline Resources found`
-        : `No Pipeline Resources found in the '${namespace}' namespace`;
+    const { namespace, type, ...rest } = this.props;
+    let emptyText = `No Pipeline Resources found`;
+    if (type && namespace !== ALL_NAMESPACES) {
+      emptyText = `No Pipeline Resources found of type '${type}' in the '${namespace}' namespace`;
+    } else if (type) {
+      emptyText = `No Pipeline Resources found of type '${type}'`;
+    } else if (namespace !== ALL_NAMESPACES) {
+      emptyText = `No Pipeline Resources found in the '${namespace}' namespace`;
+    }
     return <TooltipDropdown {...rest} emptyText={emptyText} />;
   }
 }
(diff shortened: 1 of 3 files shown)


src/containers/PipelineResourcesDropdown/PipelineResourcesDropdown.js
```diff
@@ -37,11 +37,14 @@ class PipelineResourcesDropdown extends React.Component {
   }
 
   render() {
-    const { namespace, ...rest } = this.props;
-    const emptyText =
-      namespace === ALL_NAMESPACES
-        ? `No Pipeline Resources found`
-        : `No Pipeline Resources found in the '${namespace}' namespace`;
+    const { namespace, type, ...rest } = this.props;
+    let emptyText = `No Pipeline Resources found`;
```
AlanGreene: We'll need multiple strings here as concatenation won't work for translation

src/containers/PipelineResourcesDropdown/PipelineResourcesDropdown.js
```diff
@@ -55,9 +58,17 @@ PipelineResourcesDropdown.defaultProps = {
...
+  const pipelineResourcesReducer = (acc, pipelineResource) => {
+    if (!type || type === pipelineResource.spec.type) {
+      acc.push(pipelineResource.metadata.name);
+    }
+    return acc;
+  };
   return {
-    items: getPipelineResources(state, { namespace }).map(
-      pipelineResource => pipelineResource.metadata.name
+    items: getPipelineResources(state, { namespace }).reduce(
```
AlanGreene: We could use `filter` here rather than `reduce`, something like:
`getPipelineResource(...).filter(resource => (!type || type === resource.spec.type))`

src/containers/PipelineResourcesDropdown/PipelineResourcesDropdown.js
ncskier: Bummer, so I'll have to go through all the combinations of those two booleans?

src/containers/PipelineResourcesDropdown/PipelineResourcesDropdown.js
ncskier: Ok, I'll have to do `filter` to `map` though to end with a string.

ncskier:
Thank you for the review @AlanGreene. I have made changes addressing your requests. Let me know what you think 🙂

src/containers/PipelineResourcesDropdown/PipelineResourcesDropdown.js
AlanGreene: We could simplify the strings, omit the type and rely on the label / other context to make it clear to the user. Otherwise yes, we'd need one string per combo.

PR #274, opened 2019-06-19
correcting minor mistake in previous secrets table update

carlos-logro:
# Changes

Mistakenly switched the id of the last table cell (delete button) to use the namespace & name as value but the correct value should be just the name (like it was originally). This is because when the user clicks to delete the secret, the name of that secret is retrieved from that id & used in the process of deleting that selected secret.


AlanGreene:
@carlos-logro won't you also need the namespace to ensure the correct secret is deleted?

carlos-logro:
@AlanGreene now that I think of it you're probably right... ill go ahead & make the necessary changes

PR #279, opened 2019-06-21
Add revision label, ensure labels in tests

vtereso:
# Changes
Add revision label to pipelineRuns with related test
Consolidate label generation into functions

# Submitter Checklist

review them:

for more details._


dibbles:
I'm guessing you would have a second PR to make similar labeling in the experimental repo, for pipelineruns created from the webhook automation?

In either case, I'm not sure this will actually resolve the issue you are trying to resolve - as the "revision" could be specified as a commit id - and what you are trying to resolve is filtering on a given branch name?

So I might not be able to find all builds on a certain branch as there is nothing on the label that identifies the branch?  

I think what you have tried to do is the generic approach based on standard tekton resources which is fine for a generic filter where the revision in the resource was xyz - but I think in the webhook case the revision would be a commit id too.

Probably need to have a chat to work out what your approach is regarding the problem you are solving as a whole.

PR #282, opened 2019-06-25
PipelineRun creation form for arbitrary pipelines

ncskier:
# Changes

This PR updates the `CreatePipelineRun` container to display the necessary fields for creating a PipelineRun for an arbitrary Pipeline. The form dynamically shows inputs for Params and PipelineResources based on the selected Pipeline's spec.

The tests have also been updated.

Since the dashboard's backend API for creating arbitrary PipelineRuns is not merged yet, this PR still does not have a button to display the CreatePipelineRun container.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/CreatePipelineRun/CreatePipelineRun.js
+++ b/src/containers/CreatePipelineRun/CreatePipelineRun.js
@@ -14,497 +14,419 @@ limitations under the License.
 import React from 'react';
 import { connect } from 'react-redux';
 import {
-  Button,
-  ComposedModal,
   Form,
   FormGroup,
   InlineNotification,
-  ModalBody,
-  ModalFooter,
-  ModalHeader,
-  TextInput,
-  Toggle
+  Modal,
+  TextInput
 } from 'carbon-components-react';
 import {
   NamespacesDropdown,
+  PipelineResourcesDropdown,
   PipelinesDropdown,
   ServiceAccountsDropdown
 } from '..';
-import { getSelectedNamespace } from '../../reducers';
+import { getPipeline } from '../../reducers';
 import { createPipelineRun } from '../../api';
+import { getStore } from '../../store/index';
 
 import './CreatePipelineRun.scss';
 import { ALL_NAMESPACES } from '../../constants';
 
-const resourceNameRegExp = /^[-.a-z0-9]{1,253}$/;
-const resourceNameInvalidText =
-  'Must consist of only lower case alphanumeric characters, -, and . with at most 253 characters';
-const noSpacesRegExp = /^\S*$/;
-const noSpacesInvalidText = 'Must contain no spaces';
-const urlRegExp = /^https?:\/\/.+/;
-const urlInvalidText = 'Must be a valid URL';
+const NAMESPACE = 'namespace';
+const PIPELINE_REF = 'pipelineRef';
+const RESOURCE_SPECS = 'resourceSpecs';
+const PARAM_SPECS = 'paramSpecs';
+const PIPELINE_ERROR = 'pipelineError';
+const DISABLED = 'disabled';
 
-const formValidation = {
-  namespace: {
-    required: true
-  },
-  pipeline: {
-    required: true
-  },
-  serviceAccount: {
-    required: true
-  },
-  gitName: {
-    required: false,
-    regexp: resourceNameRegExp
-  },
-  gitRepoURL: {
-    required: false,
-    regexp: urlRegExp
-  },
-  gitRevision: {
-    required: false,
-    regexp: noSpacesRegExp
-  },
-  imageName: {
-    required: false,
-    regexp: resourceNameRegExp
-  },
-  imageRegistryName: {
-    required: false,
-    regexp: noSpacesRegExp
-  },
-  imageRepoName: {
-    required: false,
-    regexp: noSpacesRegExp
-  },
-  helmPipeline: {
-    required: false
-  },
-  helmSecret: {
-    required: false,
-    regexp: resourceNameRegExp
+const parsePipelineInfo = (state, pipelineRef, namespace) => {
+  if (pipelineRef) {
+    let resourceSpecs;
+    let paramSpecs;
+    const pipeline = getPipeline(state, { name: pipelineRef, namespace });
+    if (pipeline) {
+      ({ resources: resourceSpecs, params: paramSpecs } = pipeline.spec);
+    }
+    const pipelineError = !pipeline;
+    return {
+      [RESOURCE_SPECS]: resourceSpecs,
+      [PARAM_SPECS]: paramSpecs,
+      [PIPELINE_ERROR]: pipelineError
+    };
   }
+  return {};
 };
 
-const initialState = {
-  namespace: {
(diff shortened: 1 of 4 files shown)


src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -14,497 +14,427 @@ limitations under the License.
...
-      <Form onSubmit={this.onSubmit}>
-        <ComposedModal
+      <Form>
+        <Modal
           className="create-pipelinerun"
-          onClose={this.onClose}
           open={open}
+          modalHeading="Create PipelineRun"
+          modalLabel={this.getPipelineInfo(PIPELINE_REF)}
+          primaryButtonText="Submit"
```
AlanGreene: the primary button text should reflect the action it will perform, in this case 'Create' or 'Create PipelineRun'

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -14,497 +14,427 @@ limitations under the License.
...
-                kind="error"
-                hideCloseButton
-                lowContrast
-                title="Unable to submit"
-                subtitle="Please fix the fields with errors"
-              />
-            )}
+          {this.getPipelineInfo(PIPELINE_ERROR) && (
+            <InlineNotification
+              kind="error"
```
AlanGreene: notifications may require the `lowContrast` prop to restore previous styling since the Carbon 10.3 update (switched to a high contrast dark theme by default for some components)

also shouldn't be necessary to specify the `subtitle` if it's being left blank

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -14,497 +14,427 @@ limitations under the License.
...
+                />
+              ))}
+            </FormGroup>
+          )}
+          <FormGroup legendText="Optional values">
             <ServiceAccountsDropdown
-              id="dropdown-service-account"
-              namespace={namespaceValue}
+              id="create-pipelinerun--sa-dropdown"
+              helperText="Optional"
```
AlanGreene: optional fields should be denoted as such via their labels, e.g. 'Service Account (optional)'
https://www.carbondesignsystem.com/components/form/usage#optional-vs.-required-fields

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -14,497 +14,427 @@ limitations under the License.
...
-    this.onInputChange = this.onInputChange.bind(this);
-    this.onHelmToggleChange = this.onHelmToggleChange.bind(this);
-    this.onSubmit = this.onSubmit.bind(this);
-    this.onClose = this.onClose.bind(this);
-    this.reset = this.reset.bind(this);
-    this.getPipeline = this.getPipeline.bind(this);
     this.checkFormValidation = this.checkFormValidation.bind(this);
-    this.checkValidation = this.checkValidation.bind(this);
-  }
+    this.getPipelineInfo = this.getPipelineInfo.bind(this);
```
AlanGreene: it would be nice to remove these calls to `bind`, but I know there are some issues with tests when trying to spy on class properties so ok with leaving them if needed

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -14,497 +14,427 @@ limitations under the License.
...
+    if (!valid) {
+      return;
+    }
 
-    const namespace = this.getNamespace();
-    const pipelineValue = this.getPipeline();
+    // Send API request to create PipelineRun
+    const pipelineRef = this.getPipelineInfo(PIPELINE_REF);
+    const namespace = this.getPipelineInfo(NAMESPACE);
     const payload = {
```
AlanGreene: constructing the payload should be the responsibility of the API
doing this in a follow up PR is fine

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -14,497 +14,427 @@ limitations under the License.
...
-          <ModalFooter>
-            <Button kind="secondary" onClick={this.onClose}>
-              Cancel
-            </Button>
-            <Button kind="primary" type="submit">
-              Submit
-            </Button>
-          </ModalFooter>
-        </ComposedModal>
+            <TextInput
```
AlanGreene: Should this be a NumberInput? What's the unit, seconds?

src/containers/CreatePipelineRun/CreatePipelineRun.js
ncskier: I can probably remove some of them. I always put `.bind` just to be safe, but I probably don't need to bind all of the functions? @AlanGreene can you please explain when I do and don't need to use `bind`?

src/containers/CreatePipelineRun/CreatePipelineRun.js
ncskier: Ok, I'll move this into the API call when I implement it in a follow up PR 👍

src/containers/CreatePipelineRun/CreatePipelineRun.js
ncskier: Weirdly enough, I think it's actually a string, because it gets parsed here https://golang.org/pkg/time/#ParseDuration (mentioned here: https://godoc.org/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1#PipelineRunSpec)

src/containers/CreatePipelineRun/CreatePipelineRun.js
AlanGreene: If the function is being called as a result of a user-event (e.g. click handler) you'll need to ensure it's bound to the component scope as by default it would be bound to the DOM node that was the target of the event (i.e. `this` would be a reference to the DOM node in the callback).

If the function is being called exclusively from other contexts where `this` is already the component, there's no need to bind.

Arrow functions (`() => { ... }`) (and class properties) and `bind` are just a few of the approaches to control scope in JS. You can read a little more about the background / reasoning for arrow functions here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_separate_this and then follow the rabbit hole as deep as you like on JS scope in general 😉

src/containers/CreatePipelineRun/CreatePipelineRun.js
AlanGreene: Ah I see, so a user could enter a value like "10m" for 10 minutes for example. We might need to provide a little more guidance in the UI for this but let's leave it as-is for now and we can revisit once we start using it ourselves.

src/containers/CreatePipelineRun/CreatePipelineRun.js
ncskier: Ok, I think I've minimized the number of `bind`s, and it will still work 😄

ncskier:
Thank you for the review @AlanGreene! I've made some changes based on your requests. Let me know how it looks! 🙂

PR #286, opened 2019-06-27
[Do not merge] Intentional test failure to test our test reporting on Prow

a-roberts:
We've noticed for the webhooks extension, when our Node.js tests fail the unit tests still report as having passed.

I've intentionally added a test failure here to see if the same problem impacts our dashboard.


a-roberts:
Investigation done - problem is in the `test-it.sh` script for the extension directory: we are capturing failure codes correctly here in the dashboard 😄

PR #315, opened 2019-07-04
Add rebuild button for PipelineRuns

a-roberts:
# Changes

Adds a rebuild button for PipelineRuns. Takes an existing PipelineRun, does a lookup to find its actual object (from the CRD), modifies the name and omits the resource version (we'll get a new one on creation, definitely unique and handles for us), and kicks off a new PipelineRun based on it.

Added tests too that checks the label convention is met - the Go tests are the good ones really.

**The notification now appears on the PipelineRun page, right at the top. Rebuild button is now a ghost, off to the right, looking very swish**. 

# Submitter Checklist

review them:

for more details._


AlanGreene:
Rather than having a specific API for rebuilding a PipelineRun given a name, could we not have the client create the PipelineRun with the details extracted from the one being rebuilt?

a-roberts:
> Rather than having a specific API for rebuilding a PipelineRun given a name, could we not have the client create the PipelineRun with the details extracted from the one being rebuilt?

I wanted to that but it's easier here to handle the need for a new name and resource version, which i can handle trivially through this API (otherwise we'd do that through the frontend - and I'd rather put anything non-trivial manipulating resources on the backend)

src/components/Rebuild/Rebuild.js
```diff
@@ -0,0 +1,115 @@
...
+    };
+  }
+
+  handleRebuild = event => {
+    if (event) {
+      event.preventDefault();
+    }
+
+    /* eslint no-useless-rename: "error" */
+    const pipelinerunname = this.props.runName;
```
jessm12: Consistent camel case please

pkg/endpoints/pipeline.go
```diff
@@ -757,3 +763,138 @@ func (r Resource) UpdatePipelineRun(request *restful.Request, response *restful.
...
+	if !endsInDashAndNumbers {
+		newName = fmt.Sprintf("%s-r-%s", newName, strconv.Itoa(time.Now().Nanosecond())[4:])
+	}
+
+	logging.Log.Debugf("Rebuilt PipelineRun name: %s", newName)
+
+	return newName
+}
+
+/* E.g. POST to http://localhost:9097/v1/namespaces/default/rebuild (provide name)
```
jessm12: Provide name how?

pkg/endpoints/pipeline.go
```diff
@@ -757,3 +763,138 @@ func (r Resource) UpdatePipelineRun(request *restful.Request, response *restful.
...
+		logging.Log.Debug("No name has been provided and a request has been made to rebuild with the name of an existing PipelineRun")
+		logging.Log.Debugf("Rebuilding PipelineRun: %s", requestData.PIPELINERUNNAME)
+		// A lookup will be made for the run, so no need to provide full data
+		// Method handles any error reporting through logs
+		rebuiltRun, err := r.rebuildImpl(nil, requestData.PIPELINERUNNAME, namespace)
+		if err != nil {
+			utils.RespondError(response, err, http.StatusInternalServerError)
+			return
+		} else {
+			// ALl is well, include the name of the new rebuilt run in the response
```
jessm12: typo

pkg/endpoints/pipeline.go
```diff
@@ -757,3 +763,138 @@ func (r Resource) UpdatePipelineRun(request *restful.Request, response *restful.
...
+			logging.Log.Debugf("Rebuilt ok, name is: %s", rebuiltRun.Name)
+			utils.WriteResponseLocation(request, response, rebuiltRun.Name)
+			return
+		}
+	}
+	// We should never get here - but if so, they've given us a request we can't handle
+	logging.Log.Error("didn't receive a request we could actually handle!")
```
jessm12: Maybe log the status message too

pkg/endpoints/pipeline_test.go
```diff
@@ -1167,3 +1167,137 @@ func TestPOSTCreatePipelineRuns(t *testing.T) {
...
+// Ensure it's identical apart from the name and resource version
+
+func TestRebuildWithExistingRunFromName(t *testing.T) {
+	server, r, namespace := testutils.DummyServer()
+	testPipelineNameToRun := "testPipeline1"
+
+	pipeline := createPipelineForTesting(testPipelineNameToRun, namespace, server, r, t)
+	pipelineRun := createPipelineRunForTesting(testPipelineNameToRun, namespace, server, r, t)
+
+	// Check we've got a new PipelineRun and its the same apart from name and resource version
```
jessm12: this comment confused me, isn't this supposed to be on the second creation only?

jessm12:
Screenshot 2019-07-04 at 17 21 53

this doesn't look quite right 🤔

a-roberts:
^^ I'll fix that, trying low contrast 😄  and moving to the right, then margins

a-roberts:
Working on displaying the notification on the PipelineRun page itself now and then I'm done with this

a-roberts:
@AlanGreene @ncskier could I get another review please?

image

And we get a toast notification now above, on that same page. I think my Go tests are decent but the React ones less so - but we're thinking we may change the PipelineRun code big time soon for the dash anyway. Open to improvement suggestions, not sure if I'm correct in passing the functions as props from the parent to child so I can set state accordingly.

👼

Rebuild.scss
```diff
@@ -0,0 +1,5 @@
+#rebuild-btn {
```
AlanGreene: this file should be removed

src/components/Rebuild/Rebuild.test.js
```diff
@@ -0,0 +1,53 @@
...
+  // No namespace provided here, payload as above
+  expect(rebuildMock).toHaveBeenCalledWith(undefined, expected);
+});
+
+it('rebuild button is ghost styled', async () => {
+  jest
+    .spyOn(API, 'rebuildPipelineRun')
+    .mockImplementation(() => Promise.resolve(headers));
+  renderWithRouter(<Rebuild {...props} runName="fake-pipeline-run" />);
+  const rebuildButton = document.getElementById('rebuild-btn');
```
AlanGreene: prefer `getByTestId` or similar functions from react-testing-library#render over using document.getElementById and other native DOM functions

src/components/Rebuild/Rebuild.test.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+};
+
+const headers = {
+  get() {
+    return 'fake-pipeline-run';
+  }
+};
+
+it('rebuild button creates API call with correct run name', async () => {
```
AlanGreene: doesn't look like `async` is needed here

src/components/Rebuild/Rebuild.scss
```diff
@@ -0,0 +1,11 @@
+.rebuild-btn {
+  position: absolute !important;
```
AlanGreene: avoid `!important`, instead prefer more specific selectors

`.bx--btn.rebuild-btn` or similar


also avoid `position: absolute` if possible as it can cause problems with responsive layouts, leading to overlapping content and other issues

src/components/Rebuild/Rebuild.js
```diff
@@ -0,0 +1,86 @@
...
+import './Rebuild.scss';
+
+export class Rebuild extends Component {
+  handleRebuild = event => {
+    if (event) {
+      event.preventDefault();
+    }
+
+    /* eslint no-useless-rename: "error" */
+    const pipelinerunname = this.props.runName;
```
AlanGreene: why are these eslint comments needed?

src/components/Rebuild/Rebuild.js
```diff
@@ -0,0 +1,86 @@
...
+    /* eslint no-useless-rename: "error" */
+    const pipelineName = this.props.pipelineRun.spec.pipelineRef.name;
+    const { namespace } = this.props.pipelineRun.metadata.namespace; // was namespaceToUse and OK
```
AlanGreene: this doesn't look right, `namespace.namespace`...

src/components/Rebuild/Rebuild.js
```diff
@@ -0,0 +1,86 @@
...
+    /* eslint no-useless-rename: "error" */
+    const payload = {
+      /* eslint no-useless-rename: "error" */
+      pipelinerunname
+    };
+
+    const promise = rebuildPipelineRun(namespace, payload);
+
+    promise
```
AlanGreene: no need for the `const` here, may have been leftover from previous code / debugging

```js
rebuildPipelineRun(namespace, payload)
  .then(header => {
```

src/components/Rebuild/Rebuild.js
```diff
@@ -0,0 +1,86 @@
...
+      .then(headers => {
+        const logsURL = headers.get('Content-Location');
+        const newPipelineRunName = logsURL.substring(
+          logsURL.lastIndexOf('/') + 1
+        );
+        const finalURL = `/namespaces/${namespace}/pipelines/${pipelineName}/runs/${newPipelineRunName}`;
+        this.props.setShowRebuildNotification(true);
```
AlanGreene: Rather than setting these individually can we just set a single object? Something like this:

```js
this.props.showRebuildNotification({
  message: 'Rebuilt run successfully',
  kind: 'success',
  logsURL
});
```

In PipelineRun you would `setState({ rebuildNotification })` with this object as the value.

src/components/Rebuild/Rebuild.js
```diff
@@ -0,0 +1,86 @@
...
+            this.props.setShowRebuildNotificationKind('error');
+            break;
+          default:
+        }
+      });
+  };
+
+  render() {
+    return (
+      <>
```
AlanGreene: no need for a fragment here as it's just returning a single button

src/components/Rebuild/Rebuild.js
```diff
@@ -0,0 +1,86 @@
...
+        <Button
+          data-testid="rebuild-btn"
+          id="rebuild-btn"
```
AlanGreene: see other comment regarding use of testid, `id` shouldn't be necessary

src/components/RunHeader/RunHeader.js
```diff
@@ -44,16 +44,23 @@ class RunHeader extends Component {
...
-                <div className="block-icon">
-                  {getStatusIcon({ reason, status })}
-                </div>
-                <div className="run-name" title={runName}>
-                  {runName}
-                </div>
-                <span className="status-label">{reason}</span>
-                <span className="time">{lastTransitionTime}</span>
-              </h1>
+              <div>
```
AlanGreene: Is the wrapping `div` here used to ensure the block extends the full width? If so it would be better to achieve this via explicit styles in the SCSS rather than adding the extra wrapper, this will make it easier to handle the responsive layout at smaller screen sizes.

src/components/RunHeader/RunHeader.js
```diff
@@ -44,16 +44,23 @@ class RunHeader extends Component {
...
+                  <div className="block-icon">
+                    {getStatusIcon({ reason, status })}
+                  </div>
+                  <div className="run-name" title={runName}>
+                    {runName}
+                  </div>
+                  <span className="status-label">{reason}</span>
+                  <span className="time">{lastTransitionTime}</span>
+                  <Rebuild
+                    {...this.props}
```
AlanGreene: Does the Rebuild button require all of the props from RunHeader or just a subset?

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -225,13 +272,52 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
...
+                to={showRebuildNotificationLogsURL}
+                onClick={ () => this.setShowRebuildNotification(false)}
+              >
+              View status of this rebuilt run
+              </Link>
+            }
+          />
+        )}
+
+        <div>
```
AlanGreene: what's the purpose of this additional `div` wrapper?

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -225,13 +272,52 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
...
     return (
       <>
-        <RunHeader
-          lastTransitionTime={lastTransitionTime}
-          loading={loading}
-          runName={pipelineRunName}
-          reason={pipelineRunReason}
-          status={pipelineRunStatus}
-        />
+        {/* prettier-ignore */ (showRebuildNotification && !showRebuildNotificationLogsURL) && (
```
AlanGreene: Prettier complaining about unnecessary parentheses? Let's just keep it happy 😉

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -225,13 +272,52 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
...
+          <ToastNotification
+            data-testid="rebuildsuccessnotification"
+            lowContrast
+            subtitle=""
+            title={showRebuildNotificationMessage}
+            kind={showRebuildNotificationKind}
+            caption={
+              <Link id="newpipelinerunlink" 
+                to={showRebuildNotificationLogsURL}
+                onClick={ () => this.setShowRebuildNotification(false)}
```
AlanGreene: if you're navigating away and this component is unmounted the state will be re-initialised next time you return anyway

src/components/Rebuild/Rebuild.js
a-roberts: Destructuring occurs otherwise so if we were to run `npm run lint:fix`, the variable name would become runName instead. I really do want it to be `pipelinerunname` as this is to be used in the JSON payload for the actual rebuild function

I'll add this as a comment in the code to make it obvious for maintainers too

src/containers/PipelineRun/PipelineRun.js
a-roberts: Hmm without this added, on the rebuilt PipelineRun this caption is presented

src/components/Rebuild/Rebuild.js
a-roberts: Yeah, fixed now, and my test code that was expecting it to be called with undefined and not the actual namespace, thanks

src/components/RunHeader/RunHeader.js
a-roberts: I'm passing 

```
    setShowRebuildNotification={this.setShowRebuildNotification}
          setShowRebuildNotificationKind={this.setShowRebuildNotificationKind}
          setShowRebuildNotificationLogsURL={this.setShowRebuildNotificationLogsURL}
          setShowRebuildNotificationMessage={this.setShowRebuildNotificationMessage}
```

to the RunHeader from PipelineRun which Rebuild needs to access, so unfortunately yeah 😢

src/components/Rebuild/Rebuild.js
AlanGreene: Rather than renaming it here you should be able to do it where it's used in the payload below and avoid these eslint comments. Normally the comments are used to allow something that would otherwise be a violation, it's odd to see it used this way.

src/containers/PipelineRun/PipelineRun.js
AlanGreene: my bad, since you're navigating from PipelineRun to another PipelineRun the component isn't unmounted just updated

src/containers/PipelineRun/PipelineRun.js
a-roberts: Prettier removes the brackets which I think changes the functionality as it doesn't evaluate both expressions, lemme check to confirm...

src/components/Rebuild/Rebuild.js
a-roberts: Great idea, trying

src/components/RunHeader/RunHeader.js
a-roberts: I'll run with this as well

src/containers/PipelineRun/PipelineRun.js
a-roberts: My bad, removed the rule, all is well

src/components/RunHeader/RunHeader.js
a-roberts: I think I added this so that it was all in one div to return as a jsx element and it wasn't to do with the block - but I don't even need it

Just built and tested, looks great without it anyway so I've removed it

src/components/Rebuild/Rebuild.js
a-roberts: Done, this was a nice refactor, thank you

a-roberts:
@AlanGreene ready for another review please, I think I've addressed all of your comments (e.g. the props refactor and selective CSS without using !important). I fixed some impl code and thus test code along the way too for the namespace reference.

src/components/Rebuild/Rebuild.scss
```diff
@@ -0,0 +1,6 @@
+//.rebuild-btn {
```
AlanGreene: remove

src/components/Rebuild/Rebuild.js
AlanGreene: I'd like to revisit these, but not a blocker for this PR.

src/components/RunHeader/RunHeader.js
```diff
@@ -53,6 +53,11 @@ class RunHeader extends Component {
                 </div>
                 <span className="status-label">{reason}</span>
                 <span className="time">{lastTransitionTime}</span>
+                <Rebuild
```
AlanGreene: RunHeader is reused for TaskRuns, and although it would be useful to have there it doesn't work at the moment so I'd suggest adding this only when `pipelineRun` is set.

a-roberts: On it, I'll add
```
      {pipelineRun && (
                  <Rebuild
                    {...this.props}
                    pipelineRun={pipelineRun}
                    runName={runName}
                  />
                )}
```
and retest

a-roberts:
@AlanGreene ready for another one please 😄

Issue #328, opened 2019-07-09
Missing permissions for Secrets


mnuttall:
The secret management panel, appearing soon, shows up the need for some extra permissions when running outside of a permissive environment such as Docker Desktop. 

- Creating secrets fails due to the lack of ability to 'patch' servicaccounts
- Deleting secrets fails due to the lack of ability to 'delete' secrets

These roles appear in two files; gcr-tekton-dashboard.yaml and dev-tekton-dashboard.yaml.

To reproduce the problem:
- Enable the 'Secrets' panel via SideNav.js
- Install on OpenShift
- Create and delete secrets

jessm12:
PR here https://github.com/tektoncd/dashboard/pull/355, tested with POST and DELETE calls and works 🙂

jessm12:
/close
Merged into master 🚀

PR #345, opened 2019-07-11
secrets paneling polished

carlos-logro:
# Changes

- Update table to match the other ones in the dashboard.

- Changed service account text field for the dropdown container.

- Made error messages more specific.

- Updated actions & reducers to delete one or more secrets.

- Updated delete modal to show all secrets selected that are about to be deleted.

- Separated handle change methods to avoid clutter. 

- Updated all the necessary tests.

### DIFF
--- a/src/actions/secrets.js
+++ b/src/actions/secrets.js
@@ -38,24 +38,35 @@ export function fetchSecrets({ namespace } = {}) {
         secretsFormatted.push(object);
       });
       dispatch(fetchSecretsSuccess(secretsFormatted));
-    } catch (e) {
-      const error = new Error('Could not fetch secrets');
+    } catch (error) {
       dispatch({ type: 'SECRETS_FETCH_FAILURE', error });
     }
     return secrets;
   };
 }
 
-export function deleteSecret(name, namespace) {
+export function deleteSecret(secrets) {
   return async dispatch => {
     dispatch({ type: 'SECRET_DELETE_REQUEST' });
-    try {
-      await deleteCredential(name, namespace);
-      dispatch({ type: 'SECRET_DELETE_SUCCESS', name, namespace });
-    } catch (e) {
-      const error = new Error(`Could not delete secret "${name}"`);
-      dispatch({ type: 'SECRET_DELETE_FAILURE', error });
-    }
+    const deletePromises = secrets.map(secret => {
+      const { name, namespace } = secret;
+      const response = deleteCredential(name, namespace);
+      const timeout = new Promise((resolve, reject) => {
+        setTimeout(() => {
+          reject(new Error('An error occured deleting the secret(s).'));
+        }, 1000);
+      });
+      const deleteWithinTimePromise = Promise.race([response, timeout]);
+      return deleteWithinTimePromise;
+    });
+
+    Promise.all(deletePromises)
+      .then(() => {
+        dispatch({ type: 'SECRET_DELETE_SUCCESS', secrets });
+      })
+      .catch(error => {
+        dispatch({ type: 'SECRET_DELETE_FAILURE', error });
+      });
   };
 }
 
@@ -66,10 +77,7 @@ export function createSecret(postData, namespace) {
     try {
       await createCredential(postData, namespace);
       dispatch(fetchSecrets());
-    } catch (e) {
-      const error = new Error(
-        `Could not create secret "${postData.name}" in namespace ${namespace}`
-      );
+    } catch (error) {
       dispatch({ type: 'SECRET_CREATE_FAILURE', error });
     }
   };
--- a/src/components/SecretsDeleteModal/SecretsDeleteModal.js
+++ b/src/components/SecretsDeleteModal/SecretsDeleteModal.js
@@ -16,7 +16,7 @@ import { Modal } from 'carbon-components-react';
 import './SecretsDeleteModal.scss';
 
 const SecretsDeleteModal = props => {
-  const { open, id, handleClick, handleDelete } = props;
+  const { open, toBeDeleted, handleClick, handleDelete } = props;
 
   return (
     <Modal
@@ -29,9 +29,13 @@ const SecretsDeleteModal = props => {
       onRequestSubmit={handleDelete}
       onRequestClose={handleClick}
     >
-      <p>
-        Are you sure you want to delete the secret <strong>{id}</strong>?
-      </p>
+      <p>Are you sure you want to delete these secrets?</p>
+      <ul>
+        {toBeDeleted.map(secret => {
+          const { name, namespace } = secret;
+          return <li key={`${name}:${namespace}`}>{name}</li>;
+        })}
+      </ul>
     </Modal>
   );
 };
--- a/src/components/SecretsDeleteModal/SecretsDeleteModal.test.js
+++ b/src/components/SecretsDeleteModal/SecretsDeleteModal.test.js
@@ -15,15 +15,35 @@ import React from 'react';
 import { fireEvent, render } from 'react-testing-library';
 import SecretsDeleteModal from './SecretsDeleteModal';
 
-it('SecretsDeleteModal renders with passed secret id', () => {
+it('SecretsDeleteModal renders with one passed secret', () => {
   const props = {
     open: true,
-    id: 'dummySecret',
+    toBeDeleted: [{ name: 'secret-name', namespace: 'default' }],
     handleClick() {},
     handleDelete() {}
   };
   const { queryByText } = render(<SecretsDeleteModal {...props} />);
-  expect(queryByText('dummySecret')).toBeTruthy();
+  expect(queryByText('secret-name')).toBeTruthy();
+  expect(queryByText('Cancel')).toBeTruthy();
+  expect(queryByText('Delete')).toBeTruthy();
+  expect(queryByText('Delete Secret')).toBeTruthy();
+});
+
+it('SecretsDeleteModal renders with multiple passed secrets', () => {
+  const props = {
+    open: true,
+    toBeDeleted: [
+      { name: 'secret-name', namespace: 'default' },
+      { name: 'other-secret', namespace: 'default' },
+      { name: 'another-one', namespace: 'default' }
+    ],
+    handleClick() {},
+    handleDelete() {}
+  };
+  const { queryByText } = render(<SecretsDeleteModal {...props} />);
+  expect(queryByText('secret-name')).toBeTruthy();
+  expect(queryByText('other-secret')).toBeTruthy();
+  expect(queryByText('another-one')).toBeTruthy();
   expect(queryByText('Cancel')).toBeTruthy();
   expect(queryByText('Delete')).toBeTruthy();
   expect(queryByText('Delete Secret')).toBeTruthy();
@@ -34,7 +54,7 @@ it('Test SecretsDeleteModal click events', () => {
   const handleDelete = jest.fn();
   const props = {
     open: true,
-    id: 'dummySecret',
+    toBeDeleted: [{ name: 'secret-name', namespace: 'default' }],
     handleClick,
     handleDelete
   };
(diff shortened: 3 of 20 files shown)


src/actions/secrets.js
```diff
@@ -39,23 +39,34 @@ export function fetchSecrets({ namespace } = {}) {
       });
       dispatch(fetchSecretsSuccess(secretsFormatted));
     } catch (e) {
-      const error = new Error('Could not fetch secrets');
-      dispatch({ type: 'SECRETS_FETCH_FAILURE', error });
+      dispatch({ type: 'SECRETS_FETCH_FAILURE', error: e });
```
AlanGreene:
```diff
-} catch (e) {
+} catch (error) {
-dispatch({ type: 'SECRETS_FETCH_FAILURE', error: e });
+dispatch({ type: 'SECRETS_FETCH_FAILURE', error });
```

similar repeated a few times below

src/actions/secrets.js
```diff
@@ -39,23 +39,34 @@ export function fetchSecrets({ namespace } = {}) {
...
-      const error = new Error(`Could not delete secret "${name}"`);
-      dispatch({ type: 'SECRET_DELETE_FAILURE', error });
-    }
+    const deletePromises = secrets.map(secret => {
+      const { name, namespace } = secret;
+      const response = deleteCredential(name, namespace);
+      const timeout = new Promise((resolve, reject) => {
+        setTimeout(() => {
+          reject();
+        }, 500);
```
AlanGreene: seems quite low, might need to revisit if we start seeing flakiness

src/components/SecretsDeleteModal/SecretsDeleteModal.js
```diff
@@ -29,9 +29,17 @@ const SecretsDeleteModal = props => {
       onRequestSubmit={handleDelete}
       onRequestClose={handleClick}
     >
-      <p>
-        Are you sure you want to delete the secret <strong>{id}</strong>?
-      </p>
+      <p>Are you sure you want to delete the(se) secret(s)?</p>
```
AlanGreene: 'Are you sure you want to delete these secrets?'

still ok if there's only 1

src/components/SecretsDeleteModal/SecretsDeleteModal.js
```diff
@@ -29,9 +29,17 @@ const SecretsDeleteModal = props => {
...
+      <ul>
+        {toBeDeleted.map(secret => {
+          const { name, namespace } = secret;
+          return (
+            <li key={`${name}:${namespace}`}>
+              <strong>{secret.name}</strong>
```
AlanGreene: style the `li` rather than using `strong` here since `strong` implies importance / urgency which doesn't apply to these values

src/components/SecretsDeleteModal/SecretsDeleteModal.test.js
```diff
@@ -15,15 +15,15 @@ import React from 'react';
 import { fireEvent, render } from 'react-testing-library';
 import SecretsDeleteModal from './SecretsDeleteModal';
 
-it('SecretsDeleteModal renders with passed secret id', () => {
+it('SecretsDeleteModal renders with passed secrets', () => {
```
AlanGreene: should have a test with multiple secrets too

src/components/SecretsModal/UniversalFields.js
```diff
@@ -34,15 +36,24 @@ const UniversalFields = props => {
...
       <NamespacesDropdown
         id="namespace"
-        selectedItem={{ id: selectedNamespace, text: selectedNamespace }}
-        onChange={handleChange}
+        selectedItem={
+          selectedNamespace
+            ? {
+                id: selectedNamespace,
+                text: selectedNamespace,
+                component: 'namespace'
```
AlanGreene: what's this `component` used for?

src/components/SecretsModal/UniversalFields.js
```diff
@@ -54,11 +65,11 @@ const UniversalFields = props => {
           component: 'accessTo'
         }}
         items={[
-          { id: 'git', text: 'Git Server', component: 'accessTo' },
-          { id: 'docker', text: 'Docker Registry', component: 'accessTo' }
+          { id: 'git', text: 'Git Server' },
+          { id: 'docker', text: 'Docker Registry' }
```
AlanGreene: We should document that these secrets are only for resources that require the tekton-labelled secrets, and that plain kube secrets required by other resource types (e.g. cluster, pullRequest, storage) are not covered by this UI.

src/components/SecretsTable/SecretsTable.js
```diff
@@ -28,124 +39,160 @@ const {
...
     Object.keys(secret.annotations).forEach(function annotationSetup(key) {
       annotations += `${key}: ${secret.annotations[key]}\n`;
     });
-    return {
-      id: `${secret.namespace}:${secret.name}`,
-      secret: secret.name,
-      annotations,
-      add: <Close />
-    };
+    return selectedNamespace !== ALL_NAMESPACES
```
AlanGreene: try to avoid duplication, especially in data formatting, as any update is increasingly likely to cause bugs the more places the code needs to be changed to match

```js
const formattedSecret = {
  id: `${secret.namespace}:${secret.name}`,
  secret: secret.name,
  annotations
};

if (selectedNamespace === ALL_NAMESPACES) {
   formattedSecret.namespace = secret.namespace;
}

return formattedSecret;
```

src/components/SecretsTable/SecretsTable.js
```diff
@@ -28,124 +39,160 @@ const {
...
             )}
-          </Table>
-        </TableContainer>
+          </TableContainer>
+        )}
+      />
+      {secrets.length === 0 && selectedNamespace === ALL_NAMESPACES && (
+        <div className="noSecrets">
+          <p>
+            No secrets created under any namespace, click the button above to
```
AlanGreene: avoid references to layout / position (above, left, etc.) in user-facing text as this is not accessible to users of screen readers, and also tightly couples the text to the design

src/containers/Secrets/Secrets.js
```diff
@@ -60,19 +61,23 @@ export /* istanbul ignore next */ class Secrets extends Component {
   handleDeleteSecretClick = value => {
     this.setState({
       openDeleteSecret: true,
-      toBeDeleted: value
+      toBeDeleted: value.map(secret => ({
+        namespace: secret.id.split(':')[0],
```
AlanGreene: rather than doing this string manipulation these values should be provided as an object, I thought that was done in a previous PR

src/containers/Secrets/Secrets.js
```diff
@@ -60,19 +61,23 @@ export /* istanbul ignore next */ class Secrets extends Component {
...
   delete = () => {
-    const { name, namespace } = this.state.toBeDeleted;
-    this.props.deleteSecret(name, namespace);
+    this.props.deleteSecret(this.state.toBeDeleted);
     this.handleDeleteSecretToggle();
   };
 
   render() {
-    const { secrets, loading, error } = this.props;
+    // const { loading, error, selectedNamespace } = this.props;
```
AlanGreene: remove

src/reducers/secrets.test.js
```diff
@@ -61,9 +61,8 @@ it('SECRETS_FETCH_FAILURE', () => {
 });
 
 it('SECRET_DELETE_SUCCESS', () => {
```
AlanGreene: should also have test for deleting multiple secrets

src/components/SecretsTable/SecretsTable.js
carlos-logro: I see, is there some other way to let the user know that there are no secrets in the given namespace or all-namepaces? 

this is what i was trying to achieve:
Screen Shot 2019-07-15 at 2 54 06 PM

carlos-logro:
@AlanGreene made those changes you requested except for one because did not know how to approach it ATM but i did leave a question above regarding it. Also, I added the helper text for the 'Access To' dropdown, let me know if that is the correct text. Thank you!

src/components/SecretsTable/SecretsTable.js
AlanGreene: You can still tell the user that there are no secrets, the only part of the string that's problematic is the use of 'above'. You could replace it with 'click the Create Secret button to add one' or similar.

carlos-logro: ohh i understand now. awesome!

jessm12:
I noticed while working https://github.com/tektoncd/dashboard/issues/328 that some of the credentials endpoints are not documented (create, update and delete are missing) https://github.com/tektoncd/dashboard/blob/master/DEVELOPMENT.md#api-definitions.

Please as you add the panelling could you add those definitions too @carlos-logro? 😄

src/components/SecretsDeleteModal/SecretsDeleteModal.js
```diff
@@ -29,9 +29,13 @@ const SecretsDeleteModal = props => {
...
       onRequestClose={handleClick}
     >
-      <p>
-        Are you sure you want to delete the secret <strong>{id}</strong>?
-      </p>
+      <p>Are you sure you want to delete these secrets?</p>
+      <ul>
+        {toBeDeleted.map(secret => {
+          const { name, namespace } = secret;
+          return <li key={`${name}:${namespace}`}>{secret.name}</li>;
```
AlanGreene:
```diff
-return <li key={`${name}:${namespace}`}>{secret.name}</li>;
+return <li key={`${name}:${namespace}`}>{name}</li>;
```

src/components/SecretsModal/UniversalFields.js
```diff
@@ -34,31 +36,39 @@ const UniversalFields = props => {
...
+              }
+            : ''
+        }
+        onChange={handleChangeNamespace}
+        invalid={invalidFields.indexOf('namespace') > -1}
       />
       <Dropdown
         id="accessTo"
         titleText="Access To:"
+        helperText="These secrets are only for resources that are tekton-labelled, thus plain kube secrets required by other resource types (e.g. cluster, pullRequest, storage) are not covered by this UI."
```
AlanGreene: This is too much content for the `helperText`. When I said we should document this I meant in the README or similar, I should have made that clearer.

src/components/SecretsTable/SecretsTable.js
```diff
@@ -28,124 +39,162 @@ const {
...
+            <TableToolbar>
+              <TableBatchActions {...getBatchActionProps()}>
+                <TableBatchAction
+                  id="delete-btn"
+                  data-testid="deleteButton"
+                  renderIcon={Delete}
+                  onClick={() => {
+                    handleDelete(
+                      selectedRows.map(secret => ({
+                        namespace: secret.id.split(':')[0],
```
AlanGreene: again this secret should be an object containing the name and namespace values instead of doing string manipulation

carlos-logro:
@AlanGreene 

Updates are done. Is worth noting that for the invalid input text I only added the word `required` to those inputs because at the moment there is no rules on validating them except for white spaces. Also, in the annotations part I added a `<p>` element instead of populating the `invalidText` prop because when it appeared it interfered with the layout of those elements and since multiple annotations can be filled, it be weird having them say the same message. Secondly please verify if the Read.me was correctly documented. Lastly, unfortunately the `selectedRows` is an array rendered by the `DataTable` component where its values are derived by the `headers` & `rows` passed as props to this component which makes it difficult for it to contain the secrets info as an object. 

for example: 
```Javascript
headers = [
 {
    header: "Secret",
    key: "secret"
 },
 {
    header: "Namespace",
    key: "namespace"
 },
 {
    header: "Annotations",
    key: "annotations"
 }
]

rows = [
 {
    annotations: "tekton.dev/git-0: https://github.com",
    id: "default:example",
    namespace: "default",
    secret: "example
 }
]
```

results in this selectedRows:
```Javascript
[
 {
   cells: [
      {
        id: "default:example:secret", 
        value: "example", 
        isEditable: false, 
        isEditing: false, 
        isValid: true,
      },
      {
        id: "default:example:namespace", 
        value: "default", 
        isEditable: false, 
        isEditing: false, 
        isValid: true,
      },
      {
        id: "default:example:annotations", 
        value: "tekton.dev/git-0: https://github.com", 
        isEditable: false, 
        isEditing: false, 
        isValid: true,
      }
   ],
   id: "default:example"
   isExpanded: false
  isSelected: true
 }
]

```

AlanGreene:
@carlos-logro Just 2 items below then I'm happy to get this merged. Once this is in there are some other minor style issues we should address, let's sync up on Monday to work through them.

1. I'm still seeing intermittent (~1/3) issues in the UI when deleting multiple secrets, even when the API responses are successful. Can you increase the timeout?

2. Also thinking back on our discussion with design this week about secrets, access tokens, etc., might be better to leave out the readme change until that's fully resolved to avoid confusion.

PR #354, opened 2019-07-16
include namespace as prop for extension component

carlos-logro:
# Changes
With this PR merged, the extension repo will be able to use the selected namespace to filter out the webhook shown in the table.

this is the PR in extension that depends on this PR: https://github.com/tektoncd/experimental/pull/148


AlanGreene:
This is already available to extensions via the selectors provided.

PR #382, opened 2019-07-26
Add mechanism for including other k8s resources

steveodonovan:
# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -57,10 +57,13 @@ export function getKubeAPI(
   ].join('');
 }
 
-export function getTektonAPI(type, { name = '', namespace } = {}, queryParams) {
+export function getResourcesAPI(
+  { group, version, type, name = '', namespace },
+  queryParams
+) {
   return [
     apiRoot,
-    '/proxy/apis/tekton.dev/v1alpha1/',
+    `/proxy/apis/${group}/${version}/`,
     namespace && namespace !== ALL_NAMESPACES
       ? `namespaces/${encodeURIComponent(namespace)}/`
       : '',
@@ -69,6 +72,13 @@ export function getTektonAPI(type, { name = '', namespace } = {}, queryParams) {
     encodeURIComponent(name),
     queryParams ? `?${new URLSearchParams(queryParams).toString()}` : ''
   ].join('');
+}
+
+export function getTektonAPI(type, { name = '', namespace } = {}, queryParams) {
+  return getResourcesAPI(
+    { group: 'tekton.dev', version: 'v1alpha1', type, name, namespace },
+    queryParams
+  );
 }
 
 export function getExtensionBaseURL(name) {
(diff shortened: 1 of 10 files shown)


src/containers/ResourceList/ResourceList.js
```diff
@@ -0,0 +1,159 @@
...
+          hideCloseButton
+          lowContrast
+          title={`Error loading ${type}`}
+          subtitle={getErrorMessage(error)}
+        />
+      );
+    }
+
+    return (
+      <>
```
AlanGreene: no need for the fragment here since there's just a single root component being returned

src/containers/ResourceList/ResourceList.js
```diff
@@ -0,0 +1,159 @@
...
+                namespace,
+                creationTimestamp,
+                uid
+              } = resource.metadata;
+              return (
+                <StructuredListRow className="definition" key={uid}>
+                  <StructuredListCell>
+                    <Link
+                      exact
+                      to={`/${resource.apiVersion}/${
```
AlanGreene: should use router utils to build the URL

src/containers/SideNav/SideNav.js
```diff
@@ -155,11 +155,15 @@ export class SideNav extends Component {
...
           {extensions.length > 0 &&
-            extensions.map(({ displayName, name }) => (
+            extensions.map(({ displayName, name, type, apiVersion }) => (
               <SideNavMenuItem
                 element={NavLink}
                 icon={<span />}
-                to={urls.extensions.byName({ name })}
+                to={
+                  type === 'kubernetes-resource'
+                    ? `/${apiVersion}${this.getPath(`/${name}`)}`
```
AlanGreene: router utils to build this

src/api/index.js
```diff
@@ -265,17 +281,43 @@ export function deleteCredential(id, namespace) {
...
       source: getExtensionBundleURL(name, bundlelocation),
       url
     })
   );
+  return extensions.concat(
+    resourceExtensions
+      ? (resourceExtensions.items || []).map(({ spec }) => ({
+          displayName: spec.displayname,
+          name: spec.name,
+          type: spec.type,
```
AlanGreene: using `type` feels a little off, but `kind` could cause confusion... might revisit this before finalising and documenting it

src/api/index.js
```diff
@@ -57,10 +57,16 @@ export function getKubeAPI(
   ].join('');
 }
 
-export function getTektonAPI(type, { name = '', namespace } = {}, queryParams) {
+export function getResourcesAPI(
+  group,
```
AlanGreene: I think it's time to move all of these into the object with `name` and `namespace`

```js
export function getResourcesAPI(
  {
    group,
    name,
    namespace,
    type,
    version
  },
  queryParams
) {
```

no need for the default empty object since at least one of those values will always be required, also makes the consuming code easier to read (named params)

also means `getCustomResource` and `getCustomResources` can just pass them through, no need for them to be aware of the contents

```js
export function getCustomResources(...args) {
  const uri = getResourcesAPI(...args);
  return get(uri).then(checkData);
}
```

src/api/index.js
```diff
@@ -265,17 +275,44 @@ export function deleteCredential(id, namespace) {
...
+  const resourceExtensions = await get(resourceExtensionsUri);
+  extensions = (extensions || []).map(
     ({ bundlelocation, displayname, name, url }) => ({
       displayName: displayname,
       name,
       source: getExtensionBundleURL(name, bundlelocation),
       url
     })
   );
+  return extensions.concat(
```
AlanGreene: we can deconstruct the split and avoid doing it twice, also no need for ternary


```js
return extensions.concat(
  (resourceExtensions && resourceExtensions.items || []).map(({ spec }) => {
    const { displayname: displayName, name } = spec;
    const [apiGroup, apiVersion] = spec.apiVersion.split('/');
    return {
      displayName,
      name,
      apiGroup,
      apiVersion,
      extensionType: 'kubernetes-resource'
    };
  })
);
```

src/containers/CustomResourceDefinition/CustomResourceDefinition.js
```diff
@@ -72,12 +74,21 @@ export /* istanbul ignore next */ class CustomResourceDefinition extends Compone
...
       default:
-        return Promise.resolve();
+        return getCustomResource({
+          group,
+          version,
+          type,
+          namespace,
+          name
+        }).then(resource => {
+          this.setState({ loading: false, resource });
```
AlanGreene: no need to set `loading: false` here as it's already done in `componentDidMount/Update`

src/utils/router.js
```diff
@@ -106,6 +106,22 @@ export const paths = {
...
+  },
+  kubernetesResources: {
+    all() {
+      return '/:group/:version/:type';
+    },
+    byNamespace() {
+      return byNamespace({ path: '/:group/:version/:type' });
+    }
+  },
+  kubernetesResource: {
```
AlanGreene: Let's merge these into `kubernetesResources` above, no need for a separate object here. Same pattern as used for existing resources.

```
kubernetesResources: {
  all() ...
  byNamespace() ...
  cluster() ...
  byName() ...
}
```

Issue #306, opened 2019-07-03
Bad helper text on Import Tekton resources panel


dibbles:
# Expected Behavior

Help text on the "Namespace" field along the lines of "The namespace in which the the resources will be created" 

# Actual Behavior

Text currently reads: "The namespace that the pipelinerun applying resources will run under"

# Steps to Reproduce the Problem

1. Click on Import Tekton resources

# Additional Info

PR #404, opened 2019-08-06
Bad helper text on Import Tekton resources panel

carlos-logro:
issue: #306 

# Changes
Updated the helper text in the namespace field.


src/containers/ImportResources/ImportResources.js
```diff
@@ -169,7 +169,7 @@ export class ImportResources extends Component {
           />
           <NamespacesDropdown
             id="import-namespaces-dropdown"
-            helperText="The namespace that the PipelineRun applying resources will run under"
+            helperText="The namespace in which the the resources will be created"
```
AlanGreene: `the the`

Issue #400, opened 2019-08-05
Create secrets panel is too short (hard to select last SA)


a-roberts:
# Expected Behavior
It should be easy to pick any service account from the dropdown

# Actual Behavior
It's a nightmare to choose the last service account because the modal isn't tall enough, literally end up selecting the last one then using the down arrow. Here I'm trying to choose `tekton-webhooks-extension` as my service account from the dropdown - it's not visible.

![image](https://user-images.githubusercontent.com/12814972/62467383-8880f280-b78b-11e9-8977-8da76bae35bd.png)

Yeah there's a scroll bar to the right but I can't use that or I'm no longer selecting this dropdown, and that's not for this

# Steps to Reproduce the Problem

1. Try to pick the last service account when creating a Docker secret

PR #405, opened 2019-08-06
Create secrets panel is too short (hard to select last SA)

carlos-logro:
issue: #400 

# Changes
Managed to make the Service Account dropdown easier to access by adding a css style to it so it appears upward instead of downward.

### DIFF
--- a/src/components/SecretsModal/BasicAuthFields.js
+++ b/src/components/SecretsModal/BasicAuthFields.js
@@ -12,9 +12,8 @@ limitations under the License.
 */
 
 import React from 'react';
-import { TextInput } from 'carbon-components-react';
+import { Select, SelectItem, TextInput } from 'carbon-components-react';
 import './SecretsModal.scss';
-import ServiceAccountsDropdown from '../../containers/ServiceAccountsDropdown';
 
 const BasicAuthFields = props => {
   const {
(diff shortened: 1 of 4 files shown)


carlos-logro:
Problem:
Screen Shot 2019-08-06 at 1 47 29 PM

carlos-logro:
Solution: 
Screen Shot 2019-08-06 at 1 47 11 PM

AlanGreene:
If we used a [Select](http://react.carbondesignsystem.com/?path=/story/select--default) (technically more correct for use in a form) here instead of a Dropdown it shouldn't have the same issue as it uses the native browser select menu which should be able to extend over the edge of the dialog.

carlos-logro:
> If we used a [Select](http://react.carbondesignsystem.com/?path=/story/select--default) (technically more correct for use in a form) here instead of a Dropdown it shouldn't have the same issue as it uses the native browser select menu which should be able to extend over the edge of the dialog.

that means refactoring the service account dropdown component??

carlos-logro:
@AlanGreene 

didn't refactor the sa component but i did changed the one in the modal which meant using action & reducers to get the sa in the store!

Screen Shot 2019-08-07 at 3 51 19 PM
Screen Shot 2019-08-07 at 3 51 27 PM

src/components/SecretsModal/BasicAuthFields.js
```diff
@@ -24,8 +23,13 @@ const BasicAuthFields = props => {
     handleChangeTextInput,
     handleChangeServiceAccount,
     invalidFields,
-    serviceAccount
+    serviceAccounts
   } = props;
+
+  const saItems = serviceAccounts
+    .filter(sa => sa.metadata.namespace === namespace)
+    .map(sa => <SelectItem value={sa.metadata.name} text={sa.metadata.name} />);
```
AlanGreene: missing `key` prop

carlos-logro: ill add a style to hide it then.

carlos-logro:
@AlanGreene hid the chevron icon! When you can please give the PR another review, Thank you

Issue #321, opened 2019-07-08
Update front end create pipeline run API to consume kube API


AlanGreene:
Continuation of https://github.com/tektoncd/dashboard/issues/290

The front end code currently uses the REST APIs provided by the back end.
Now that #268 has been merged we should begin migrating the front end APIs to use the APIs provided by the kube API server where possible / appropriate.

This issue is to track the following APIs:
- [ ] createPipelineRun

Also:
- [ ] delete any unused backend code once these are migrated

## Notes

There's already a [createPipelineRunAtProxy](https://github.com/tektoncd/dashboard/blob/47ae481e7ba856d85496715e3014d5ef1ac53c5f/src/api/index.js#L133) API that's used for the manual create pipeline run UI.

The legacy `createPipelineRun` is used by the import resources UI. This is the one that should be migrated.

Once this happens the `AtProxy` API should be renamed.

vtereso:
I picked up this issue from Carlos, but didn't realize there was still front-end work to be done. To migrate to the proxy url, the git pipelineresouce used within [pipeline0](https://github.com/tektoncd/dashboard/blob/master/config/pipeline0.yaml), will need to be created ahead of time. Someone feel free to finish up the work I have done on [my branch](https://github.com/tektoncd/dashboard/pull/439).

PR #421, opened 2019-08-13
createPipelineRun consume KubeAPI - backend

carlos-logro:
issue: #321 

# Changes
Update createPipelineRun API function call to use the KubeAPI instead of soon to be deprecated API. 

Divided the PR in two parts so this one just contains primarily updated/deleted backend code & updated e2e test.


carlos-logro:
@AlanGreene this is off topic with the issue but while working on it I noticed the notification in the `Import Resources` page to be in the bottom of the page which I find really weird & inconsistent with the rest of the dashboard. 

Screen Shot 2019-08-13 at 1 59 00 PM

carlos-logro:
@AlanGreene as you can see the integration test is failing because the e2e test we currently run uses the old API & endpoints instead of the proxy. Don't think i'll be able to update the line where it port forwards instead of starting the proxy because we still haven't migrated the credentials part. Any suggestions?

AlanGreene:
@carlos-logro I'm not sure I follow the issue about port-forward vs. proxy, this shouldn't have any impact. Regardless of which method you use to expose the dashboard, we provide a `/proxy` endpoint that proxies the kube APIs. This is **not** related to `kubectl proxy`. You should be able to update the e2e test to use the proxied APIs without changing the port-forward setup.

test/e2e-common.sh
```diff
@@ -18,19 +18,15 @@
 
 source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/e2e-tests.sh
 
+# Adam's debugging as hello world is always pending rn!
```
a-roberts: Let's get rid 😄

test/e2e-tests.sh
```diff
@@ -29,70 +29,56 @@ header "Setting up environment"
...
+export APP_SERVICE_ACCOUNT="tekton-dashboard"
+export PIPELINE_NAME="simple-pipeline-insecure"
+export IMAGE_RESOURCE_NAME="docker-image"
+export GIT_RESOURCE_NAME="git-source"
+export GIT_COMMIT="master"
+export REPO_NAME="go-hello-world"
+export REPO_URL="https://github.com/ncskier/go-hello-world"
+export EXPECTED_RETURN_VALUE="Hello World!"
+export REGISTRY="gcr.io/${E2E_PROJECT_ID}/${E2E_BASE_NAME}-e2e-img"
+export TEKTON_PROXY_URL="http://localhost:9097/proxy/apis/tekton.dev/v1alpha1/namespaces/tekton-pipelines"
```
a-roberts: The proxy URL for me is http://localhost:8001/api/v1/namespaces/tekton-pipelines/services/tekton-dashboard:http/proxy, this is actually a port-forward URL (would mention because by the looks of it, I know it contains proxy, but the variable name says otherwise)

test/e2e-tests.sh
```diff
@@ -134,7 +121,7 @@ do
   if [ "$resp" != "" ]; then
 		echo "Response from pod is: $resp"
```
a-roberts: I know this isn't your code but while you're here could you fix the indenting on this line please 👼

test/e2e-tests.sh
carlos-logro: tried using this URL but the integration test failed....

carlos-logro:
@a-roberts fire up the lgtm tag !

Issue #326, opened 2019-07-09
Filter by label on the PipelineRuns panel


jessm12:
In order to achieve https://github.com/tektoncd/dashboard/issues/223 filtering of `PipelineRuns` we must allow label filters to be selected on the `PipelineRuns` table view. 

Relies on https://github.com/tektoncd/dashboard/issues/323 for `PipelineRuns` that are manually created through the dashboard UI and https://github.com/tektoncd/experimental/issues/140 for experimental extension triggered `PipelineRuns`

Under this issue:
- Filter the request to fetch `PipelineRuns` by label selector, displaying only `PipelineRuns` that have these labels

aunderw:
Here is a PDF of the search/filter flow:

[search-scenario-github.pdf](https://github.com/tektoncd/dashboard/files/3398221/search-scenario-github.pdf)

jessm12:
Maybe https://material-ui.com/components/chips/ could be an option to use here 🙂

AlanGreene:
@jessm12 Carbon provides http://react.carbondesignsystem.com/?path=/story/tag--filter but it looks like there's an issue with the storybook at the moment. I've posted in #carbon-react to let them know.

jessm12:
@AlanGreene I was looking for filtering in carbon but couldn't find it anywhere as the storybook is broken and it's not listed there properly like you say, thanks for raising that 😅is this the right filtering that you're referring to? https://www.carbondesignsystem.com/patterns/filtering/#selecting-filters

AlanGreene:
The UI samples provided in that pattern are more for predefined filters where a user simply selects from a list. Our approach requires additional input so I wouldn't worry too much about the UI presentation in that particular pattern, although some of the other recommendations are still valid (e.g. ability to reset / clear filters).

aunderw:
This is the same scenario I posted yesterday except that it includes example results as well as a more explicit example of how filters could be passed by the URL. 

[tekton-search-and-filter-results.pdf](https://github.com/tektoncd/dashboard/files/3402275/tekton-search-and-filter-results.pdf)

jessm12:
Thank you for those @aunderw 😄

jessm12:
https://github.com/tektoncd/dashboard/issues/37 @kimholmes has linked some additional designs here with comments 🙂Though these extra features for the `PipelineRuns` filtering have not been implemented under this issue there could definitely be scope to enhance this feature in the future

jessm12:
[...]

PR #427, opened 2019-08-15
Add filter by label functionality to PipelineRuns panel

jessm12:
# Changes
https://github.com/tektoncd/dashboard/issues/326
Includes:

- The UI implementation on the `PipelineRuns` panel to add and delete label based filters and display `PipelineRuns` that match these filters
- Styling of new elements and some movement of elements on the panel
- Tests for this functionality testing adding single and multiple filters as well as deleting them
- Updates to other tests affected by the addition of filters

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/kubernetes/dashboard/blob/master/LICENSE)
 
-Tekton Dashboard is a general purpose, web-based UI for Tekton Pipelines. It allows users to manage and view Tekton Pipeline and Task runs and the resources involved in their creation, execution, and completion.
+Tekton Dashboard is a general purpose, web-based UI for Tekton Pipelines. It allows users to manage and view Tekton PipelineRuns and TaskRuns and the resources involved in their creation, execution, and completion. It also allows filtering of PipelineRuns by label. 
 
 ![Dashboard UI workloads page](docs/dashboard-ui.png)
 
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -113,10 +113,20 @@ export function getPipeline({ name, namespace }) {
   return get(uri);
 }
 
-export function getPipelineRuns({ namespace, pipelineName } = {}) {
+export function getPipelineRuns({
+  namespace,
+  pipelineName,
+  filters = []
+} = {}) {
   let queryParams;
-  if (pipelineName) {
-    queryParams = { labelSelector: `tekton.dev/pipeline=${pipelineName}` };
+  if (filters.length > 0 || pipelineName) {
+    queryParams = { labelSelector: [] };
+    if (pipelineName) {
+      queryParams.labelSelector.push(`tekton.dev/pipeline=${pipelineName}`);
+    }
+    if (filters.length > 0) {
+      queryParams.labelSelector.push(filters);
+    }
   }
   const uri = getTektonAPI('pipelineruns', { namespace }, queryParams);
   return get(uri).then(checkData);
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -3,6 +3,14 @@ export const messages = {
     'dashboard.pipelineRuns.transitionTime': 'Last Transition Time',
     'dashboard.pipelineRuns.status': 'Status',
     'dashboard.pipelineRuns.error': 'Error loading PipelineRuns',
+    'dashboard.pipelineRuns.createSuccess': 'Successfully created PipelineRun',
+    'dashboard.pipelineRuns.invalidFilter':
+      'Filters must be of the format labelKey:labelValue and contain accepted label characters',
+    'dashboard.pipelineRuns.duplicateFilter': 'No duplicate filters allowed',
+    'dashboard.pipelineRuns.labelDocs':
+      'https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set',
+    'dashboard.pipelineRuns.searchPlaceholder':
+      'Input a label filter of the format labelKey:labelValue',
     'dashboard.pipelineRun.failed': 'Cannot load PipelineRun',
     'dashboard.pipelineRun.notFound': 'PipelineRun not found',
     'dashboard.pipelineRun.error': 'Error loading PipelineRun',
--- a/src/containers/PipelineRuns/PipelineRuns.js
+++ b/src/containers/PipelineRuns/PipelineRuns.js
@@ -17,8 +17,12 @@ import { Link } from 'react-router-dom';
 import { injectIntl } from 'react-intl';
 import {
   Button,
+  Form,
   InlineNotification,
-  StructuredListSkeleton
+  Link as ReactLink,
+  Search,
+  StructuredListSkeleton,
+  Tag
 } from 'carbon-components-react';
 import { PipelineRuns as PipelineRunsList } from '@tektoncd/dashboard-components';
 import { getErrorMessage, urls } from '@tektoncd/dashboard-utils';
@@ -41,8 +45,21 @@ import { cancelPipelineRun } from '../../api';
 
 const initialState = {
   showCreatePipelineRunModal: false,
-  createdPipelineRun: null
+  createdPipelineRun: null,
+  validFilter: {
+    isValid: true,
+    filterMessageId: '',
+    filterDefaultMessage: '',
+    url: '',
+    urlMessage: ''
+  }
 };
+
+function arrayUnique(arr) {
+  return arr.filter(function unique(item, index) {
+    return arr.indexOf(item) >= index;
+  });
+}
 
 export /* istanbul ignore next */ class PipelineRuns extends Component {
   constructor(props) {
@@ -59,13 +76,19 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
     this.fetchPipelineRuns();
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps, prevState) {
     const { match, namespace } = this.props;
+    const { filters } = this.state;
+    const { filters: prevFilters } = prevState;
     const { pipelineName } = match.params;
     const { match: prevMatch, namespace: prevNamespace } = prevProps;
     const { pipelineName: prevPipelineName } = prevMatch.params;
 
-    if (namespace !== prevNamespace || pipelineName !== prevPipelineName) {
+    if (
+      namespace !== prevNamespace ||
+      pipelineName !== prevPipelineName ||
+      filters !== prevFilters
+    ) {
       this.reset();
       this.fetchPipelineRuns();
     }
@@ -74,6 +97,100 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
   toggleModal = showCreatePipelineRunModal => {
     this.setState({ showCreatePipelineRunModal });
   };
+
+  handleChange = event => {
+    const inputValue = event.target.value;
+    this.setState({
+      currentFilterValue: inputValue
+    });
+  };
+
+  handleAddFilter = event => {
+    event.preventDefault();
+    const currentURL = this.props.match.url;
+    const { currentFilterValue = '' } = this.state;
+    const filterRegex = '([a-z0-9A-Z-_./]:[a-z0-9A-Z-_.],?)+';
+    const filterValue = currentFilterValue.replace(/\s/g, '');
+    if (!filterValue.match(filterRegex)) {
+      this.setState({
+        validFilter: {
+          isValid: false,
+          filterMessageId: 'dashboard.pipelineRuns.invalidFilter',
+          filterDefaultMessage:
+            'Filters must be of the format labelKey:labelValue and contain accepted label characters',
+          url:
+            'https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set',
+          urlMessage: 'See the Kubernetes Label documentation for valid syntax'
+        }
+      });
+      return;
+    }
+    const colonToEqualsFilters = filterValue.replace(/:/g, '=');
+    let currentFiltersArray = colonToEqualsFilters.split(',');
+    currentFiltersArray = arrayUnique(currentFiltersArray);
+    if (this.props.filters.includes(currentFiltersArray[0])) {
+      this.setState({
+        validFilter: {
+          isValid: false,
+          filterMessageId: 'dashboard.pipelineRuns.duplicateFilter',
+          filterDefaultMessage: 'No duplicate filters allowed',
+          url: '',
+          urlMessage: ''
+        }
+      });
+      return;
+    }
+    const newQueryParams = `?${new URLSearchParams({
+      labelSelector: this.props.filters.concat(currentFiltersArray)
+    }).toString()}`;
+    const browserURL = currentURL.concat(newQueryParams);
+    this.props.history.push(browserURL);
+    this.resetCurrentFilterValue();
+    this.fetchPipelineRuns();
+  };
+
+  handleDeleteFilter = filter => {
+    const currentQueryParams = new URLSearchParams(this.props.location.search);
+    const labelFilters = currentQueryParams.getAll('labelSelector');
+    const labelFiltersArray = labelFilters.toString().split(',');
+    const index = labelFiltersArray.indexOf(filter);
+    labelFiltersArray.splice(index, 1);
+    const currentURL = this.props.match.url;
+    if (labelFiltersArray.length === 0) {
+      this.props.history.push(currentURL);
+    } else {
+      const newQueryParams = `?${new URLSearchParams({
+        labelSelector: labelFiltersArray
+      }).toString()}`;
+      const browserURL = currentURL.concat(newQueryParams);
+      this.props.history.push(browserURL);
+    }
+  };
+
+  handleCloseFilterError = () => {
+    this.setState({
+      validFilter: {
+        isValid: true,
+        filterMessageId: '',
+        filterDefaultMessage: '',
+        url: '',
+        urlMessage: ''
+      }
+    });
+  };
+
+  resetCurrentFilterValue() {
+    this.setState({
+      validFilter: {
+        isValid: true,
+        filterMessageId: '',
+        filterDefaultMessage: '',
+        url: '',
+        urlMessage: ''
+      },
+      currentFilterValue: ''
+    });
+  }
 
   handleCreatePipelineRunSuccess(newPipelineRun) {
     const {
@@ -114,6 +231,14 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
       intl
     } = this.props;
     const { pipelineName } = match.params;
+    const { currentFilterValue, validFilter } = this.state;
+    const {
+      filterMessageId,
+      filterDefaultMessage,
+      isValid,
+      url,
+      urlMessage
+    } = validFilter;
 
     if (loading) {
       return <StructuredListSkeleton border />;
@@ -141,7 +266,10 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
         {this.state.createdPipelineRun && (
           <InlineNotification
             kind="success"
-            title="Successfully created PipelineRun"
+            title={intl.formatMessage({
+              id: 'dashboard.pipelineRuns.createSuccess',
+              defaultMessage: 'Successfully created PipelineRun'
+            })}
             subtitle={
               <Link to={this.state.createdPipelineRun.url}>
                 {this.state.createdPipelineRun.name}
@@ -150,21 +278,86 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
             lowContrast
           />
         )}
-        <Button
-          className="create-pipelinerun-button"
-          iconDescription="Create PipelineRun"
-          renderIcon={Add}
-          onClick={() => this.toggleModal(true)}
-        >
-          Create PipelineRun
-        </Button>
-        <CreatePipelineRun
-          open={this.state.showCreatePipelineRunModal}
-          onClose={() => this.toggleModal(false)}
-          onSuccess={this.handleCreatePipelineRunSuccess}
-          pipelineRef={pipelineName}
-          namespace={selectedNamespace}
-        />
+        {!isValid && (
+          <InlineNotification
+            lowContrast
+            kind="error"
+            title={intl.formatMessage({
+              id: filterMessageId,
+              defaultMessage: filterDefaultMessage
+            })}
+            subtitle=""
+            role="alert"
+            onCloseButtonClick={this.handleCloseFilterError}
+          >
+            <ReactLink id="labelDocsLink" href={url}>
+              {urlMessage}
+            </ReactLink>
+          </InlineNotification>
+        )}
+        <Form onSubmit={this.handleAddFilter} autoComplete="on">
+          <div className="search-bar">
+            <Search
+              placeHolderText={intl.formatMessage({
+                id: 'dashboard.pipelineRuns.searchPlaceholder',
+                defaultMessage:
+                  'Input a label filter of the format labelKey:labelValue'
+              })}
+              className="search"
+              labelText={intl.formatMessage({
+                id: 'dashboard.pipelineRuns.searchPlaceholder',
+                defaultMessage:
+                  'Input a label filter of the format labelKey:labelValue'
+              })}
+              onChange={this.handleChange}
+              value={currentFilterValue}
+              data-testid="filter-search-bar"
+              id="filter-search"
+              name="filter-search"
+              autocomplete="label"
+            />
+            <Button
+              className="add-filter-button"
+              iconDescription="Add filter"
+              renderIcon={Add}
+              kind="ghost"
+              type="submit"
+            >
+              Add filter
+            </Button>
+            <Button
+              className="create-pipelinerun-button"
+              iconDescription="Create PipelineRun"
+              renderIcon={Add}
+              type="button"
+              onClick={() => this.toggleModal(true)}
+            >
+              Create PipelineRun
+            </Button>
+          </div>
+        </Form>
+        <div className="filters">
+          <div className="filters-title">Filters:</div>
+          {this.props.filters.map(filter => (
+            <Tag
+              type="blue"
+              filter
+              onClick={() => this.handleDeleteFilter(filter)}
+              key={filter}
+            >
+              {filter.replace(/=/g, ':')}
+            </Tag>
+          ))}
+        </div>
+        {this.state.showCreatePipelineRunModal && (
+          <CreatePipelineRun
+            open={this.state.showCreatePipelineRunModal}
+            onClose={() => this.toggleModal(false)}
+            onSuccess={this.handleCreatePipelineRunSuccess}
+            pipelineRef={pipelineName}
+            namespace={selectedNamespace}
+          />
+        )}
         <PipelineRunsList
           pipelineName={pipelineName}
           selectedNamespace={selectedNamespace}
@@ -176,21 +369,39 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
   }
 }
 
+PipelineRuns.defaultProps = {
+  filters: []
+};
+
+export function fetchFilters(searchQuery) {
+  const queryParams = new URLSearchParams(searchQuery);
+  let filters = [];
+  queryParams.forEach(function filterValueSplit(value) {
+    filters = value.split(',');
+  });
+  return filters;
+}
+
 /* istanbul ignore next */
 function mapStateToProps(state, props) {
   const { namespace: namespaceParam, pipelineName } = props.match.params;
+  const filters = fetchFilters(props.location.search);
   const namespace = namespaceParam || getSelectedNamespace(state);
-
   return {
     error: getPipelineRunsErrorMessage(state),
     loading: isFetchingPipelineRuns(state),
     namespace,
+    filters,
     pipelineRuns: pipelineName
       ? getPipelineRunsByPipelineName(state, {
+          filters,
           name: props.match.params.pipelineName,
           namespace
         })
-      : getPipelineRuns(state, { namespace })
+      : getPipelineRuns(state, {
+          filters,
+          namespace
+        })
   };
 }
 
--- a/src/containers/PipelineRuns/PipelineRuns.test.js
+++ b/src/containers/PipelineRuns/PipelineRuns.test.js
@@ -1,0 +1,363 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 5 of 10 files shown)


jessm12:
/hold
I'm out tomorrow so will work on the test failures next week though if anyone has any other review comments before then please do add them 😄

src/utils/index.test.js
```diff
@@ -143,7 +143,7 @@ it('sortRunsByStartTime', () => {
     sort is stable on all modern browsers so
     input order is preserved for b and e
    */
-  const sortedRuns = [b, e, f, c, d, a];
+  const sortedRuns = [e, b, f, c, d, a];
```
AlanGreene: I don't see any change to the implementation so not sure why this change is required. This test should be failing as `b` and `e` are expected to remain in that order, see comment

src/api/index.js
```diff
@@ -113,10 +113,17 @@ export function getPipeline({ name, namespace }) {
...
 
-export function getPipelineRuns({ namespace, pipelineName } = {}) {
+function getFilterValues(value) {
+  return value.tag;
+}
+
+export function getPipelineRuns({ namespace, pipelineName, filters } = {}) {
   let queryParams;
+  const pipelineNameArray = [`tekton.dev/pipeline=${pipelineName}`];
+  const filterValues = filters.map(getFilterValues);
```
AlanGreene: the filters should be passed to the API function already formatted correctly

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -88,6 +95,51 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+  };
+
+  handleDeleteFilter = filter => {
+    const currentQueryParams = new URLSearchParams(this.props.location.search);
+    const labelFilters = currentQueryParams.getAll('labelSelector');
+    const labelFiltersArray = labelFilters.toString().split(',');
+    const index = labelFiltersArray.indexOf(filter);
+    labelFiltersArray.splice(index, 1);
+    const currentURL = this.props.match.url;
+    console.log(this.props);
```
AlanGreene: remove

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -162,21 +216,55 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+            onChange={this.handleChange}
+            value={this.state.currentFilterValue}
+            data-testid="filter-search-bar"
+          />
+          <Button
+            className="add-filter-button"
+            iconDescription="Add filter"
+            renderIcon={Add}
+            kind="ghost"
+            onClick={() =>
```
AlanGreene: `onClick={this.handleAddFilter}` since it doesn't take params anymore

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -162,21 +216,55 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+          >
+            Create PipelineRun
+          </Button>
+        </div>
+        <div className="filters">
+          <div className="filters-title">Filters:</div>
+          {this.props.filters.map(filter => (
+            <Tag
+              filter
+              disabled={false}
```
AlanGreene: remove, default is false

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -162,21 +216,55 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+            renderIcon={Add}
+            onClick={() => this.toggleModal(true)}
+          >
+            Create PipelineRun
+          </Button>
+        </div>
+        <div className="filters">
+          <div className="filters-title">Filters:</div>
+          {this.props.filters.map(filter => (
+            <Tag
```
AlanGreene: each component in a list generated from an iterable (`map`, `forEach`, etc.) must have a `key` prop that is unique among its siblings, ideally based on the data (i.e. some property of `filter`)

More detail, including effects of omitting key or using bad values: https://reactjs.org/docs/lists-and-keys.html

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -285,21 +373,41 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
   }
 }
 
+PipelineRuns.defaultProps = {
+  filters: []
+};
+
+export function fetchFilters(props) {
```
AlanGreene: should just accept the search string as param, it shouldn't need to access anything else from either props or location

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -285,21 +373,41 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+  return filters;
+}
+
 /* istanbul ignore next */
 function mapStateToProps(state, props) {
   const { namespace: namespaceParam, pipelineName } = props.match.params;
+  const filters = fetchFilters(props);
   const namespace = namespaceParam || getSelectedNamespace(state);
-
+  console.log(props);
```
AlanGreene: remove

src/reducers/index.js
```diff
@@ -144,16 +144,31 @@ export function isFetchingPipelineResources(state) {
...
+  { filters, namespace = getSelectedNamespace(state) } = {}
 ) {
-  return pipelineRunsSelectors.getPipelineRuns(state.pipelineRuns, namespace);
+  const runs = pipelineRunsSelectors.getPipelineRuns(
+    state.pipelineRuns,
+    namespace
+  );
+  return runs.filter(pipelineRun => {
+    return filters.every(filter => {
+      const keyValueFilter = filter.split(':');
```
AlanGreene: `const [filterKey, filterValue] = filter.split(':');`

src/utils/index.test.js
jessm12: I had only changed this as it was failing locally for me and the order of them was changing when I printed out the expected, I will change this back now 😅

src/api/index.js
jessm12: Turns out this formatting wasn't needed at all 😄

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -149,21 +200,53 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
-        <CreatePipelineRun
-          open={this.state.showCreatePipelineRunModal}
-          onClose={() => this.toggleModal(false)}
-          onSuccess={this.handleCreatePipelineRunSuccess}
-          pipelineRef={pipelineName}
-          namespace={selectedNamespace}
-        />
+        <div className="search-bar">
+          <Search
+            placeHolderText="Input a filter e.g. gitRepo:https://github.com/example/helloworld"
```
a-roberts: This all looks great, the only thing I'd add is something in the UI that mentions it's by label only - it's unclear atm despite the above example that this is a key value pair and that you can't, for example, just search on `success`, or by a certain pipeline name. I'd prefer this to be done via the UI and not through docs - perhaps 'Add label filter' instead of 'Add filter', or making use of the placeholder text

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -75,6 +82,49 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
   };
 
+  handleChange = event => {
+    const inputValue = event.target.value;
+    this.setState({
+      currentFilterValue: inputValue
+    });
+  };
+
+  handleAddFilter = () => {
```
a-roberts: We could add a small bit of input validation here too, if I don't put a `:` in my query then display something (a small error icon in the textfield?) suggesting that only key:value pairs are accepted here

src/containers/PipelineRuns/PipelineRuns.test.js
```diff
@@ -0,0 +1,330 @@
...
+    },
+    byId: {
+      'id-service-account-1': {
+        metadata: {
+          name: 'service-account-1',
+          namespace: 'namespace-1',
+          uid: 'id-service-account-1'
+        }
+      },
+      'id-service-account-2': {
```
a-roberts: Can remove these namespace 2 and service account 2 definitions can't we?

src/reducers/index.js
```diff
@@ -144,16 +144,31 @@ export function isFetchingPipelineResources(state) {
...
   state,
-  { namespace = getSelectedNamespace(state) } = {}
+  { filters, namespace = getSelectedNamespace(state) } = {}
 ) {
-  return pipelineRunsSelectors.getPipelineRuns(state.pipelineRuns, namespace);
+  const runs = pipelineRunsSelectors.getPipelineRuns(
+    state.pipelineRuns,
+    namespace
+  );
+  return runs.filter(pipelineRun => {
```
a-roberts: Let's get this documented then in our main readme

src/api/index.js
```diff
@@ -113,10 +113,12 @@ export function getPipeline({ name, namespace }) {
...
 }
 
-export function getPipelineRuns({ namespace, pipelineName } = {}) {
+export function getPipelineRuns({ namespace, pipelineName, filters } = {}) {
   let queryParams;
+  const pipelineNameArray = [`tekton.dev/pipeline=${pipelineName}`];
+  queryParams = { labelSelector: filters };
   if (pipelineName) {
-    queryParams = { labelSelector: `tekton.dev/pipeline=${pipelineName}` };
+    queryParams = { labelSelector: filters.concat(pipelineNameArray) };
```
steveodonovan: remove line 118

`queryParams.labelSelector.push(`tekton.dev/pipeline=${pipelineName}`)`

Is filters ever undefined ? Also down stream if filters is empty will we end up with trailing `?` on our api calls

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -41,7 +43,8 @@ import { cancelPipelineRun } from '../../api';
 
 const initialState = {
   showCreatePipelineRunModal: false,
-  createdPipelineRun: null
+  createdPipelineRun: null,
+  currentFilterValue: ''
```
steveodonovan: Why does it need to be initialised to empty string ?

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -59,12 +62,16 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps, prevState) {
     const { match, namespace } = this.props;
+    const { filters } = this.state;
+    const { filters: prevFilters } = prevState;
     const { pipelineName } = match.params;
     const { match: prevMatch, namespace: prevNamespace } = prevProps;
     const { pipelineName: prevPipelineName } = prevMatch.params;
-
+    if (filters !== prevFilters) {
```
steveodonovan: This shouldnt be its own conditional, add a branch to the one below

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -175,21 +258,39 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
   }
 }
 
+PipelineRuns.defaultProps = {
+  filters: []
+};
+
+export function fetchFilters(searchQuery) {
+  const queryParams = new URLSearchParams(searchQuery);
```
steveodonovan: Query params always defined ? Why would we have multiple query params with comma seperation ?

src/reducers/index.js
```diff
@@ -144,16 +144,31 @@ export function isFetchingPipelineResources(state) {
...
+    return filters.every(filter => {
+      const [filterKey, filterValue] = filter.split(':');
+      if (filterKey !== undefined && filterValue !== undefined) {
+        const value = pipelineRun.metadata.labels[filterKey];
+        if (value === filterValue) {
+          return true;
+        }
+      }
+      return false;
```
steveodonovan: This can be done in one line.

`return filterKey && filterValue && pipelineRun.metadata.labels[filterKey] == filterValue`

src/containers/PipelineRuns/PipelineRuns.js
jessm12: `props.search.location` is '' by default so it'll be always defined, if there are no params in the query then it'll just iterate over nothing and return no filters

jessm12:
Thanks for your comments @steveodonovan 😄 The structure of the query is `labelSelector=key:value, key:value, key:value`, this would also leave scope for potentially adding fieldSelector to this query as well

steveodonovan:
Did you sync up with @AlanGreene on that schema ? My web development experience is limited but I do know that `:` and `,` are unsafe characters to use in a url. Are we handling the encoding/decoding ?

jessm12:
Yes, the encoding and decoding is handled by the URLSearchParams, the URL is of the form http://localhost:8000/#/pipelineruns?labelSelector=foo%3Abar%2Cbaz%3Abam

steveodonovan:
@jessm12 I'm not seeing any of my feedback actually addressed, just the comments marked resolved. Missing a commit or is that intentional ?

jessm12:
Ah yes, haven't pushed yet! Just adding input validation now for filters and providing some feedback when it's incorrect input 😄

jessm12:
@steveodonovan have pushed the code now, please take a look! 😅

a-roberts:
/approve

I believe Jess has addressed Al's feedback and all of mine, if you think this is good as well @steveodonovan please give this the final say lgtm 😄

jessm12:
@steveodonovan where can I go to see that initial URL? I can't see it in the browser or network tab 😅I will address that linter warning 👍

steveodonovan:
Screenshot 2019-08-20 at 14 09 46


the network tab, the outbound http request to the back end.

Seeing a few console errors to be addressed too.

jessm12:
@steveodonovan I have addressed the initial URL empty query param and the remaining console errors for this change 🙂

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -75,6 +83,58 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
 
+  handleChange = event => {
+    const inputValue = event.target.value;
+    this.setState({
+      currentFilterValue: inputValue
+    });
+  };
+
+  handleAddFilter = () => {
+    this.setState({
```
steveodonovan: you might have some unexpected behaviour around the number of rer-enders happening here

src/reducers/index.js
```diff
@@ -144,16 +144,29 @@ export function isFetchingPipelineResources(state) {
...
+    namespace
+  );
+  return runs.filter(pipelineRun => {
+    return filters.every(filter => {
+      const [filterKey, filterValue] = filter.split(':');
+      if (filterKey && filterValue) {
+        const value = pipelineRun.metadata.labels[filterKey];
+        return value === filterValue;
+      }
+      return false;
```
steveodonovan: My earlier comment was not addressed.

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -149,21 +210,64 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+            role="alert"
+          />
+        )}
+        <div className="search-bar">
+          <Search
+            placeHolderText="Input a label filter - labelKey:labelValue"
+            className="search"
+            labelText="Search by label filter"
+            onChange={this.handleChange}
+            value={this.state.currentFilterValue}
```
steveodonovan: Shouldn't reference state directly, unwrap from the state at the top of the component

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -75,6 +83,58 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+    });
+  };
+
+  handleAddFilter = () => {
+    this.setState({
+      invalidFilter: false
+    });
+    const currentURL = this.props.match.url;
+    const { currentFilterValue = '' } = this.state;
+    if (!currentFilterValue.match('([a-z A-Z]+:[a-z A-Z]+,?)+')) {
```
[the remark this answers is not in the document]
jessm12: I want to be able to render an error on the page when an invalid filter is input so was using a component state variable and not in the store to achieve this as we have in many other components, I want to not allow the user to input that value at all. I'm not sure if I could achieve this with that method?

I will switch the logic and also add the number checking as you're right that wouldn't be covered by that pattern! 👍

src/containers/PipelineRuns/PipelineRuns.js
jessm12: hmm, It seemed to work fine in practice and in the tests, what behaviour do you think I need to address?

src/reducers/index.js
jessm12: We did reduce the amount of code by working from what you suggested though believe that the logic flow is easier to read this way with some separation of the checks

src/api/index.js
```diff
@@ -113,10 +113,16 @@ export function getPipeline({ name, namespace }) {
   return get(uri);
 }
 
-export function getPipelineRuns({ namespace, pipelineName } = {}) {
+export function getPipelineRuns({ namespace, pipelineName, filters } = {}) {
```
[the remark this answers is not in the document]
jessm12: Sorry I'm not completely sure what your question is referring to? 😅filters is defined as a default prop so is an empty array if not set I think?

src/containers/PipelineRuns/PipelineRuns.js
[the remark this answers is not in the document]
steveodonovan: Sorry, meant to say state, not store. Not having it in the state would be much better but I see the difficulty, problem is this flip flopping of this value in the state is going to cause re-renders which isn't ideal. It'll do for now I guess, @AlanGreene Might have an idea when hes back.

Also be sure to include anything else that might appear in there (afaik it can be almost anything right ?)

src/containers/PipelineRuns/PipelineRuns.js
jessm12: If I don't reset the state for the notification it stays their indefinitely even after the user inputs a new valid filter, I'm not sure of a way that I can improve this?

steveodonovan: Not needed sure, I was thinking https://github.com/tektoncd/dashboard/pull/427/files#diff-ee9adb63dcebb1625cd7ef8aa5d27cb4R111 or in resetCurrentFilter it could unset that value

src/reducers/index.js
steveodonovan: Totally disagree, if branches and multiple returns are definitely more complicated to read than a simple right to left boolean, it even looks like a regular sentence :sweat_smile: `return filterKey && filterValue && pipelineRun.metadata.labels[filterKey] == filterValue`

jessm12: fair enough I see what you're saying

src/containers/PipelineRuns/PipelineRuns.js
jessm12: ah got you, I'll go with the latter

src/api/index.js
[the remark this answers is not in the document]
jessm12: I think I understand what you meant if it was called from some other component then it may not be defined, I'll add the default 👍

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -149,21 +209,64 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
-          Create PipelineRun
-        </Button>
-        <CreatePipelineRun
-          open={this.state.showCreatePipelineRunModal}
-          onClose={() => this.toggleModal(false)}
-          onSuccess={this.handleCreatePipelineRunSuccess}
-          pipelineRef={pipelineName}
-          namespace={selectedNamespace}
-        />
+        {!this.state.validFilter && (
```
steveodonovan: Last nit , should be good to go after, unwrap this on line 176 along side currentFilterValue

steveodonovan:
Found a bug running it locally. If you add a filter which finds nothing and refresh the page you get the following

Screenshot 2019-08-22 at 09 29 41

jessm12:
@steveodonovan I think this should be ready to go now? 😄

steveodonovan:
I'm seeing some more unexpected behaviour. If you navigate to the page, apply filters, refresh then clear the filters on the refreshed page, the clearing of those filters doesn't work. Might seem niche but a more common use case for this would be sending a link to another user. They wouldn't be able to interact with he filters as expected.

jessm12:
@steveodonovan 🤔I'll take a look

steveodonovan:
I see why. https://github.com/tektoncd/dashboard/pull/427/files#diff-ee9adb63dcebb1625cd7ef8aa5d27cb4R157 We fetch with the filter. So if you have a filter that gives a subset of the pipeline runs, the clear that filter, the other piplineRuns are not in the store and we don't make a request to get the rest.

jessm12:
@steveodonovan yeah I see 😅I thought adding a this.fetchPipelineRuns() to the end of that method might work to fetch the rest but doesn't seem to, do you have any idea of the fix?

a-roberts:
Tested this just now, my suggestions:

1. Don’t allow duplicate filters
2. Enter key for "Add filter" please
3. Would be helpful to have ability to get previous filters you’ve added (history like on textfields? Is there a knob we can use for this component)?
4. For input it should be : and not =, replace found : with = for the proxy calls (do a get pod and grep on label, it's of the format foo:bar not foo=bar)
5. Did one invalid filter, closed notification, invalid filter again and it didn’t pop up

a-roberts:
Thanks @jessm12 for making the requested changes! Over to @AlanGreene and @steveodonovan once more for the last lgtm please 😄

steveodonovan:
It might be by design but the filter doesn't handle white space `tekton.dev/pipeline: 999-working-logs` is an example of the label applied by tekton on the resource and it has the white space. The white space before the pipeline name in the filter doesn't work here. 

Screenshot 2019-08-29 at 11 30 20


@a-roberts why the no duplicate filters ? It can stay, just curious.

Also I can apply duplicate filters without error. `tekton.dev/pipeline:999-working-logs,tekton.dev/pipeline:999-working-logs`

Screenshot 2019-08-29 at 11 29 31

a-roberts:
@steveodonovan good catch with the whitespace, as we can't have duplicate labels on the actual PipelineRun (key already existing) the only way you'd input it here is really through user error, so I'd rather we prevent it here

jessm12:
hmm, oops I think I accidentally removed the whitespace check on subsequent changes 🤦‍♀not sure what's happening with the duplication as I thought I had that working and have tested, I'll look again

steveodonovan:
I was under the impression duplicate labels are applied with or conditions for some reason (just checked, i was wrong). In that case the duplicate detection needs to account the label key, not the full key value.

`tekton.dev/pipeline:999-woring-logs,tekton.dev/pipeline:99-working-logs,tekton.dev/pipeline:999-wor-logs`

Screenshot 2019-08-29 at 11 41 02

jessm12:
I've got you, I hadn't realised that was the way that the labelling worked either, will fix that

jessm12:
https://github.com/tektoncd/dashboard/pull/427/files#diff-ee9adb63dcebb1625cd7ef8aa5d27cb4R106 I'm not sure why but moving this whitespace replace to here has broken it, trying to fix 😅

jessm12:
@steveodonovan whitespace and `key:value` pairs that are duplicates and entered at once are now handled 👍

steveodonovan:
> So concretely: gitServer: foo, gitServer: bar, gitServer: baz is totally valid, but gitServer: foo, gitServer: foo should be prevented

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#api

Not sure how aligned you'd like to be with the kubernetes way but see the `set-based requirements` section. I'm fine with it as is but it works as long as we don't call the api server with the label selectors. I tried with multiple query parameters one after the other and got some strange behaviour. 

Still need to give this another (hopefully final) once over

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -150,21 +237,94 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+              id: 'dashboard.pipelineRuns.invalidFilter',
+              defaultMessage:
+                'Filters must be of the format labelKey:labelValue and use only accepted label characters'
+            })}
+            subtitle=""
+            role="alert"
+            onCloseButtonClick={this.handleCloseFilterError}
+          />
+        )}
+        {duplicateFilter && (
```
steveodonovan: Too much code duplication, anyway to combine these 2  ? A duplicate filter is an invalid filter. Should invalid filter be an array or error messages instead of a boolean maybe ?

src/containers/PipelineRuns/PipelineRuns.test.js
```diff
@@ -0,0 +1,285 @@
+/*
```
steveodonovan: Would tests be possible for some of the problem cases found during the review

steveodonovan:
If i add a filter using an `=` instead of the `:` I don't get an error. Instead I just get an empty list.

jessm12:
@steveodonovan I reliably get an error when adding a filter with '=' instead of ':'

Which example did you use to see that behaviour please? 

Screenshot 2019-08-29 at 16 37 34

steveodonovan:
My bad, was using a bad example :sweat_smile: `tekton.dev:pipeline=999-working-logs`

just the code feedback then

src/containers/PipelineRuns/PipelineRuns.test.js
jessm12: I've added a test to check duplicate filters are prevented and to check that an invalid filter such as `foo=bar` is also reported, are there any others you would suggest?

README.md
```diff
@@ -2,7 +2,7 @@
 
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/kubernetes/dashboard/blob/master/LICENSE)
 
-Tekton Dashboard is a general purpose, web-based UI for Tekton Pipelines. It allows users to manage and view Tekton Pipeline and Task runs and the resources involved in their creation, execution, and completion.
+Tekton Dashboard is a general purpose, web-based UI for Tekton Pipelines. It allows users to manage and view Tekton Pipeline and Task Runs and the resources involved in their creation, execution, and completion. It also allows filtering of Pipeline Runs by label.
```
a-roberts: TaskRuns, PipelineRuns (both being CRDs)

src/constants/index.js
```diff
@@ -3,6 +3,12 @@ export const messages = {
     'dashboard.pipelineRuns.transitionTime': 'Last Transition Time',
     'dashboard.pipelineRuns.status': 'Status',
     'dashboard.pipelineRuns.error': 'Error loading PipelineRuns',
+    'dashboard.pipelineRuns.createSuccess': 'Successfully created PipelineRun',
+    'dashboard.pipelineRuns.invalidFilter':
+      'Filters must be of the format labelKey:labelValue and use only accepted label characters',
```
a-roberts: and contain*, then include the regex please

src/constants/index.js
```diff
@@ -3,6 +3,12 @@ export const messages = {
...
+    'dashboard.pipelineRuns.duplicateFilter': 'No duplicate filters allowed',
+    'dashboard.pipelineRuns.searchPlaceholder':
+      'Input a label filter - labelKey:labelValue',
```
a-roberts: of the format instead of -

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -41,9 +44,20 @@ import { cancelPipelineRun } from '../../api';
 
 const initialState = {
   showCreatePipelineRunModal: false,
-  createdPipelineRun: null
+  createdPipelineRun: null,
+  validFilter: {
+    state: true,
```
a-roberts: state for a variable name here... how about isValid? why state?

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -75,6 +94,90 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+      currentFilterValue: inputValue
+    });
+  };
+
+  handleAddFilter = event => {
+    event.preventDefault();
+    const currentURL = this.props.match.url;
+    const { currentFilterValue = '' } = this.state;
+    const filterValue = currentFilterValue.replace(/\s/g, '');
+    if (!filterValue.match('([a-z0-9A-Z-_./]:[a-z0-9A-Z-_.],?)+')) {
```
a-roberts: I'd store this regex up top as a global const so we can easily refer to it later instead of just in this method

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -75,6 +94,90 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+      this.setState({
+        validFilter: {
+          state: false,
+          filterMessageId: 'dashboard.pipelineRuns.invalidFilter',
+          filterDefaultMessage:
+            'Filters must be of the format labelKey:labelValue and use only accepted label characters'
```
a-roberts: contain instead of use

Again would mention the regex, and what is state being false here again? It's invalidFilter right

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -150,21 +258,80 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
...
+            role="alert"
+            onCloseButtonClick={this.handleCloseFilterError}
+          />
+        )}
+        <Form onSubmit={this.handleAddFilter} autoComplete="on">
+          <div className="search-bar">
+            <Search
+              placeHolderText={intl.formatMessage({
+                id: 'dashboard.pipelineRuns.searchPlaceholder',
+                defaultMessage: 'Input a label filter - labelKey:labelValue'
```
a-roberts: Instead of - I'd mention `of the format`

src/containers/PipelineRuns/PipelineRuns.test.js
```diff
@@ -0,0 +1,363 @@
...
+  );
+
+  const filterValue = 'baz=bam';
+  const filterInputField = getByTestId('filter-search-bar');
+  fireEvent.change(filterInputField, { target: { value: filterValue } });
+  fireEvent.click(getByText('Add filter'));
+
+  expect(
+    queryByText(
+      /Filters must be of the format labelKey:labelValue and use only accepted label characters/i
```
a-roberts: Make sure to include the regex if you go about my suggestion to have as a global var that we can refer to in message

src/containers/PipelineRuns/PipelineRuns.js
jessm12: we decided on using positive values for readability so isValid = true / isValid = false hence an invalidFilter

src/containers/PipelineRuns/PipelineRuns.test.js
jessm12: going to add a link to the kube label syntax docs instead but will add as global var for easy access 🙂

a-roberts:
Thanks @jessm12, @steveodonovan for the extensive testing and @AlanGreene!

I've sat with Jess and we're satisfied with how this is working and displaying error messages as well now, so, assuming all the tests go ahead and pass...

/approve
/lgtm

/meow space

Issue #453, opened 2019-08-28
All strings displayed in the UI should be externalised


AlanGreene:
# Expected Behavior
It should be possible to customise / translate all strings displayed in the UI (excluding user data or error messages provided by the environment).

# Actual Behavior
Many strings are still hard-coded in the components instead of using `react-intl`.

# Additional Info
We have begun extracting strings as described, this process should be completed and all new strings added using `react-intl` to avoid building up additional debt.

Some of the components that have been migrated to @tektoncd/dashboard-components have been partially updated and need to be revisited to ensure no hardcoded strings remain.

We also need to ensure the process for adding new strings is clearly documented in https://github.com/tektoncd/dashboard/blob/master/DEVELOPMENT.md or other appropriate location. Tooling should also be added to avoid manual steps to build the externalised list of current strings.

### `@tektoncd/dashboard-components`

- [x] ~~Breadcrumbs~~ (likely to be removed in design update, 'Import Tekton Resources' is the only string that would need to be extracted otherwise)
- [x] CancelButton
- [x] ErrorBoundary
- [x] Header
- [x] KeyValueList
- [x] LabelFilter
- [x] Log
- [x] LogoutButton
- [x] PageErrorBoundary
- [x] PipelineResources
- [x] PipelineRun
- [x] PipelineRuns
- [x] Rerun
- [x] ResourceTable
- [x] RunDropdown
- [x] RunHeader
- [x] Spinner
- [x] Step
- [x] StepDefinition
  - Parameters, Input Resources, Output Resources
- [x] StepDetails
- [x] StepDetailsHeader
- [x] StepStatus
- [x] Tab
- [x] Table
- [x] Tabs
- [x] Task
- [x] TaskRuns
- [x] TaskSkeleton
- [x] TaskTree
- [x] Trigger
- [x] ViewYAML

### `src/components`

- [x] PipelineResourcesModal
  - [x] GitResourceFields
  - [x] UniversalFields
- [x] SecretsDeleteModal
- [x] SecretsModal
  - [x] BasicAuthFields
  - [x] UniversalFields
- [x] TextInput
- [x] TooltipDropdown

### `src/containers`

- [x] App
- [x] ClusterTasks
- [x] CreatePipelineRun
- [x] CustomResourceDefinition
- [x] EventListener
- [x] EventListeners
- [x] Extension
- [x] Extensions
- [x] ImportResources
- [x] NamespacesDropdown
- [x] PipelineResource
- [x] PipelineResources
[...]

PR #454, opened 2019-08-28
Add logout button for OpenShift/Minishift

a-roberts:
# Changes

Displays the following log out button if we're on OpenShift or Minishift (technically, if someone's installed the `Route` CRD; which they're only going to do if they're installing on one of these platforms).

I've also modified the openshift deployment yaml so that it's ko applyable.

The button appears as follows:
image

and on click we're sent immediately the OKD login page. Note that I initially had this implemented with a `Link to`, but this lead to a page re-render on click, visible skeletons that would never resolve, and forbidden errors in the console, unless a user then refreshed manually: which lead them to the login page.

Note that I've also modified Header to be a component and not a function.

Tested by manually deploying on both OpenShift/Minishift and Docker Desktop

Annoyingly I do get:

```
main.ea1af815e2b91e09dd4a.js:1 GET http://localhost:8001/api/v1/namespaces/tekton-pipelines/services/tekton-dashboard:http/proxy/proxy/apis/route.openshift.io/v1/ 404 (Not Found)
```

in my console on Docker Desktop, which I can't for the life of me figure out how to get rid of (even though it's a caught exception I believe)

# Submitter Checklist

review them:

for more details._

@AlanGreene @steveodonovan here's my first real attempt, could you please take a look and provide feedback? There are a couple eslint ignores too I'm not a fan of!

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -28,7 +28,8 @@ module.exports = api => {
     plugins: [
       '@babel/plugin-proposal-class-properties',
       '@babel/plugin-proposal-export-default-from',
-      '@babel/plugin-syntax-dynamic-import'
+      '@babel/plugin-syntax-dynamic-import',
+      '@babel/plugin-transform-runtime'
     ],
     env: {
       development: {
--- a/packages/components/src/components/Header/Header.js
+++ b/packages/components/src/components/Header/Header.js
@@ -20,7 +20,7 @@ import './Header.scss';
 
 import tektonBanner from '../../images/tekton-horizontal-color.png';
 
-export default function Header({ children }) {
+export default function Header({ children, logoutButton }) {
   return (
     <CarbonHeader aria-label="Tekton Dashboard" className="header">
       <Link to="/">
@@ -30,6 +30,7 @@ export default function Header({ children }) {
           src={tektonBanner}
         />
       </Link>
+      {logoutButton}
       {children && <div className="header-children">{children}</div>}
     </CarbonHeader>
   );
--- a/packages/components/src/components/LogoutButton/LogoutButton.test.js
+++ b/packages/components/src/components/LogoutButton/LogoutButton.test.js
@@ -1,0 +1,44 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from 'react';
+import { waitForElement } from 'react-testing-library';
+import { renderWithRouter } from '../../utils/test';
+import Logout from './LogoutButton';
+
+it('Header renders logout button if on OpenShift', async () => {
+  const mockedResponse = {
+    headers: { get: () => 'text/plain' },
+    ok: true,
+    status: 200,
+    text: () => Promise.resolve('')
+  };
+  const shouldDisplayLogoutTrueMock = jest
+    .fn()
+    .mockImplementation(() => mockedResponse);
+  const logoutButton = (
+    <Logout shouldDisplayLogout={shouldDisplayLogoutTrueMock} />
+  );
+  const { queryByText } = renderWithRouter(logoutButton);
+  await waitForElement(() => queryByText(/log out/i));
+});
+
+it('Header does not render logout button if not on OpenShift', async () => {
+  const shouldDisplayLogoutFalseMock = jest
+    .fn()
+    .mockImplementation(() => Promise.reject());
+  const { queryByText } = renderWithRouter(
+    <Logout shouldDisplayLogout={shouldDisplayLogoutFalseMock} />
+  );
+  expect(queryByText(/log out/i)).toBeFalsy();
+});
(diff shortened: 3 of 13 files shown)


a-roberts:
Seeing `ReferenceError: regeneratorRuntime is not defined` everywhere, sad times, all is well on my laptop. Maybe I need to add `npm run bootstrap` into the test script...

test/presubmit-tests.sh
```diff
@@ -73,6 +73,7 @@ function node_test() {
   local failed=0
   echo "Running node tests from $(pwd)"
   node_npm_install || failed=1
+  npm run bootstrap || failed=1
```
AlanGreene: bootstrap is already called in node_npm_install above

packages/components/src/components/Header/Header.js
```diff
@@ -11,26 +11,65 @@ See the License for the specific language governing permissions and
...
+        )}
+        {this.props.children && (
+          <div className="header-children">{this.props.children}</div>
+        )}
+      </CarbonHeader>
+    );
+  }
 }
+
+export default Header;
```
AlanGreene: We should make sure any new strings are externalised, docs to come (https://github.com/tektoncd/dashboard/issues/453).

Steps:
- `import { injectIntl } from 'react-intl';`
- `export default injectIntl(Header);`
- instead of the hardcoded 'Log out' string use a message object, e.g.:
  ```
  this.props.intl.formatMessage({
    id: 'dashboard.header.logOut',
    defaultMessage: 'Log out'
  })
  ```
- and add it to `src/constants.js` (we're working on tooling to avoid this manual step)

packages/components/src/components/Header/Header.js
```diff
@@ -11,26 +11,65 @@ See the License for the specific language governing permissions and
...
+    try {
+      const detectedOpenShift = await shouldDisplayLogout();
+      this.setState({
+        showLogout: detectedOpenShift
+      });
+    } catch (error) {} // eslint-disable-line
+  }
+
+  handleLogoutClick() { // eslint-disable-line
+    window.location.href = '../../oauth/sign_out';
```
AlanGreene: Is this something we control? i.e. can we guarantee this relative path will be correct?

babel.config.js
```diff
@@ -28,7 +28,8 @@ module.exports = api => {
     plugins: [
       '@babel/plugin-proposal-class-properties',
       '@babel/plugin-proposal-export-default-from',
-      '@babel/plugin-syntax-dynamic-import'
+      '@babel/plugin-syntax-dynamic-import',
+      '@babel/plugin-transform-runtime'
```
AlanGreene: what's this new babel plugin required for?

packages/components/src/components/Header/Header.js
```diff
@@ -11,26 +11,65 @@ See the License for the specific language governing permissions and
...
-import React from 'react';
+import React, { Component } from 'react';
 
 import { Link } from 'react-router-dom';
-import { Header as CarbonHeader } from 'carbon-components-react';
-
+import { Button, Header as CarbonHeader } from 'carbon-components-react';
+import LogoutIcon from '@carbon/icons-react/lib/logout/20';
 import './Header.scss';
+import { shouldDisplayLogout } from '../../../../../src/api';
```
AlanGreene: code in packages/* should not consume code from the dashboard core
instead, the API function required should be passed in as a prop when the component is used

this ensures we can re-use the components in other contexts outside the dashboard

babel.config.js
a-roberts: This is me trying to get past the test failures I'm seeing atm

packages/components/src/components/Header/Header.js
a-roberts: Thanks, lemme give this a go

packages/components/src/components/Header/Header.js
a-roberts: I wonder if I can use base url/oauth/sign_out, checking

packages/components/src/components/Header/Header.js
a-roberts: I'll do this, thanks

test/presubmit-tests.sh
a-roberts: Removing, was more for testing

packages/components/src/components/LogoutButton/LogoutButton.test.js
```diff
@@ -0,0 +1,44 @@
...
+});
+
+it('Header does not render logout button if not on OpenShift', async () => {
+  const shouldDisplayLogoutFalseMock = jest
+    .fn()
+    .mockImplementation(() => Promise.reject());
+  renderWithRouter(
+    <LogoutButton shouldDisplayLogout={shouldDisplayLogoutFalseMock} />
+  );
+  expect(document.getElementById('logout-btn-link')).toBeNull();
```
AlanGreene: try to avoid `document.*` where possible, prefer the query functions provided by react-testing-library, e.g. getByText / queryByTestId, etc.

also looks like these may have been left over from a previous version of the code, these ids aren't used in the component

src/api/index.js
```diff
@@ -331,3 +331,14 @@ export function getInstallProperties() {
   const uri = `${apiRoot}/v1/properties`;
   return get(uri);
 }
+
+export function shouldDisplayLogout() {
+  console.log('toad skin');
```
AlanGreene: 🐸

src/api/index.js
```diff
@@ -331,3 +331,14 @@ export function getInstallProperties() {
...
+  const routesUri = getResourcesAPI({
+    group: 'route.openshift.io',
+    version: 'v1'
+  });
+  const thePromise = get(routesUri, { Accept: 'text/plain' });
+  console.log(`ok: ${thePromise}`);
```
AlanGreene: let's remove this, and we can then `return get(...)`

src/api/index.js
a-roberts: 🤣 oops, now the world knows my most common debug statement

src/api/index.js
a-roberts: 😆 yeah was also for debugging, gone, thanks

PR #455, opened 2019-08-29
Add log fetching parameter to log component. Remove log container

steveodonovan:
Add log fetching as a parameter to the log component

# Submitter Checklist

review them:

for more details._


packages/components/src/components/Log/Log.test.js
```diff
@@ -16,21 +16,25 @@ import Log from './Log';
 import { renderWithIntl } from '../../utils/test';
 
 it('Log renders default content', () => {
-  const { queryByText } = renderWithIntl(<Log />);
+  const { queryByText } = renderWithIntl(<Log fetchLogs={() => undefined} />);
   expect(queryByText(/No log available/i)).toBeTruthy();
 });
 
-it('Log renders the provided content', () => {
```
steveodonovan: @AlanGreene The loading state is stopping me from having this and It really should be covered. Tried a few things but nothing worked. Any suggestions on how I can wait for the component to be past the loading state in the test ?

packages/components/src/components/Log/Log.js
```diff
@@ -25,21 +25,44 @@ const LogLine = ({ data, index, style }) => (
...
+    id: 'dashboard.pipelineRun.stepFailed',
+    defaultMessage: 'Step failed'
+  }
+};
+
+export class LogContainer extends Component {
+  state = { loading: true };
+
+  componentDidMount() {
+    this.loadLog().then(this.setState({ loading: false }));
```
AlanGreene: Let's move the state update into `loadLog` where it's adding the logs to state, so this would just become:

```
this.loadLog();
this.initPolling();
```

and in `loadLog`:

`this.setState({ loading: false, logs: logs ? logs.split('\n') : undefined });`

This makes sure we're covered if we add a manual refresh or similar in future.

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -260,11 +260,20 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
 
     const logContainer = (
       <Log
-        key={stepName}
-        stepName={stepName}
-        podName={taskRun.pod}
+        fetchLogs={async () => {
```
AlanGreene: can we extract this for re-use (PipelineRun, TaskRun, TaskRuns)? it would take the run and stepStatus as params

packages/components/src/components/Log/Log.test.js
AlanGreene: You can use `waitForElement` from `react-testing-library` to wait for the expected text. If it times out (5 seconds?) the test fails.

```diff
-expect(queryByText(/testing/i)).toBeTruthy();
+await waitForElement(() => getByText(/testing/i));
```
```

File: packages/components/src/components/Log/Log.js [hunk 4]
```diff
@@ -25,21 +25,44 @@ const LogLine = ({ data, index, style }) => (
   </div>
 );
 
-class Log extends Component {
-  getLogList() {
-    const { status, intl } = this.props;
+const trailers = {
+  Completed: {
```
AlanGreene: not sure if this will work (not using intl.formatMessage here, and passing these as variables later) when we add tooling to extract the strings automatically, but I'll keep an eye on it

packages/components/src/components/Log/Log.stories.js
```diff
@@ -15,17 +15,43 @@ import React from 'react';
...
 
-const long = Array.from({ length: 2000 }, (v, i) => `Line ${i + 1}`);
+const long = Array.from(
+  { length: 2000 },
+  (v, i) => `Line ${i + 1} \n`
+).toString();
```
AlanGreene: minor nit, this prefixes all lines except the first with `,` and looks odd in the storybook

instead:
```js
const long = Array.from({ length: 2000 }, (v, i) => `Line ${i + 1}\n`).join('');
```

PR #490, opened 2019-09-04
createPipelineRun consume KubeAPI - frontend

carlos-logro:
issue: #321 

# Changes
Update createPipelineRun API function call to use the KubeAPI instead of soon to be deprecated API. 

Divided the PR in two parts so this one just contains the frontend.

### DIFF
--- a/src/containers/ImportResources/ImportResources.js
+++ b/src/containers/ImportResources/ImportResources.js
@@ -22,8 +22,13 @@ import {
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
 import { ALL_NAMESPACES, urls } from '@tektoncd/dashboard-utils';
+import { getGitValues } from '../../utils';
 
-import { createPipelineRun, determineInstallNamespace } from '../../api';
+import {
+  createPipelineResource,
+  createPipelineRun,
+  determineInstallNamespace
+} from '../../api';
 import { getSelectedNamespace } from '../../reducers';
 import { NamespacesDropdown, ServiceAccountsDropdown } from '..';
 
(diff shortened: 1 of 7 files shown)


src/containers/ImportResources/ImportResources.js
```diff
@@ -93,47 +88,88 @@ export class ImportResources extends Component {
...
+            props
+              .then(properties => {
+                const logsURL = headers.get('Content-Location');
+                const pipelineRunName = logsURL.substring(
+                  logsURL.lastIndexOf('/') + 1
+                );
+
+                const finalURL = urls.pipelineRuns.byName({
+                  namespace: properties.InstallNamespace,
+                  pipelineName: 'pipeline0',
```
mnuttall: what's the significance of naming `pipeline0` in this section?

a-roberts:
For the backend PR I noticed the removal of definePipelineRun which included labelling PipelineRuns (e.g. adding the git repo info so we can filter on it and delete them by repo when deleting a webhook for example) - see https://github.com/tektoncd/dashboard/pull/421/files#diff-2a268056787157d73c4168498850de86L121

I don't see labelling code here done through the frontend, have I missed something or is that now done elsewhere? And I've merged my bug fix for ImportResources.js so there's a conflict to resolve please

carlos-logro:
@a-roberts everything is working normal now, all labels included in the payload before sending request & once a response is received the url for the pipeline-run is specified in the notification. That being said, I never experienced problems with pipeline0

a-roberts:
> @a-roberts everything is working normal now, all labels included in the payload before sending request & once a response is received the url for the pipeline-run is specified in the notification. That being said, I never experienced problems with pipeline0. Lastly, the createPipelineRun page should be updated to include these labels as well in the payload.

Thank you!

I'll go through my scenarios once more now with your latest code, then edit this with any findings for you to look over.

src/utils/index.js
```diff
@@ -58,3 +58,24 @@ export function isStale(resource, state, resourceIdField = 'uid') {
   const incomingVersion = parseInt(resource.metadata.resourceVersion, 10);
   return existingVersion > incomingVersion;
 }
+
+export function getGitValues(url) {
+  let copyUrl;
```
AlanGreene: This can be cleaned up a bit:

```js
let copyURL = url.toLowerCase().replace(/https?:\/\//, '');
```

Also, what's the reasoning behind including labels with empty string values if we can't extract repo info?

carlos-logro: I really didn't know what to do in that case... should I return an error if labels can't be extracted??

AlanGreene: if the labels can't be extracted we don't have a valid github URL, should we even get this far?

carlos-logro: Yeah, if the labels can't be extracted I just ended up returning a blank object & ultimately create the pipelineRun with just these two labels: `'tekton.dev/pipeline'` & `app`.

Issue #499, opened 2019-09-05
Ensure stale resources are not merged to the redux store


AlanGreene:
# Expected Behavior
The resources presented in the dashboard should be the latest versions available in the cluster.

# Actual Behavior
There's a possibility that a race condition between the REST APIs and websocket events could result in a resource in the redux store being overwritten with an older version resulting in stale data being presented to the user.

# Additional Info
The reducers should protect against this type of race condition by comparing the `metadata.resourceVersion` field of the existing and incoming resources.

Also related: https://github.com/tektoncd/dashboard/issues/383, https://github.com/tektoncd/dashboard/issues/385, https://github.com/tektoncd/dashboard/issues/397, and https://github.com/tektoncd/dashboard/issues/401

PR #500, opened 2019-09-05
Ignore stale resources from REST response and websocket payloads

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/499

Before merging resources into the store, ensure any stale
resources are ignored. Most reducers use the reducerCreators
so this serves as a logical location for the filtering.

Some reducers either don't support updates or do a full replace
on receiving a list response so they have been skipped.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/reducers/reducerCreators.js
+++ b/src/reducers/reducerCreators.js
@@ -15,7 +15,7 @@ import { combineReducers } from 'redux';
 import keyBy from 'lodash.keyby';
 import merge from 'lodash.merge';
 
-import { typeToPlural } from '../utils';
+import { isStale, typeToPlural } from '../utils';
 
 function createByIdReducer({ type }) {
   const typePlural = typeToPlural(type);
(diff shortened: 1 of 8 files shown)


src/reducers/reducerCreators.js
```diff
@@ -15,14 +15,17 @@ import { combineReducers } from 'redux';
...
-import { typeToPlural } from '../utils';
+import { isStale, typeToPlural } from '../utils';
 
 function createByIdReducer({ type }) {
   const typePlural = typeToPlural(type);
   return function byId(state = {}, action) {
     switch (action.type) {
       case `${type}Created`:
       case `${type}Updated`:
+        if (isStale(action.payload, state)) {
```
steveodonovan: Anyway this could be a case rather than a condition in updated. Could we also have a race condition on created events ?

AlanGreene: yes both could have the same race condition, this already handles both as it falls through from the 'Created' case

PR #508, opened 2019-09-06
Fix key prop on Log component

AlanGreene:
# Changes

Setting `key={stepName}` works to ensure the component is
re-mounted when the selected step changes. However, it is
possible for 2 steps in different Tasks to have the same name.

Update key to use a combination of the Task and step ids to
ensure uniqueness.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRun/PipelineRun.js
+++ b/packages/components/src/components/PipelineRun/PipelineRun.js
@@ -228,7 +228,11 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
     );
 
     const logContainer = (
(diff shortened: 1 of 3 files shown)


AlanGreene:
@steveodonovan I missed this case before. Using `stepName` is fine for switching between steps in the same TaskRun, but not guaranteed to be unique across multiple TaskRuns.

I spotted this on the demo-pipeline from tektoncd/pipeline's examples after resetting my cluster. It has a 'build-and-push' step in both the 'app' and 'web' TaskRuns.

AlanGreene:
TaskRun only displays a single TaskRun resource and already uses `stepName` so it's safe to leave as-is.

Issue #320, opened 2019-07-08
Update front end credentials APIs to consume kube APIs


AlanGreene:
Continuation of https://github.com/tektoncd/dashboard/issues/290

The front end code currently uses the REST APIs provided by the back end.
Now that #268 has been merged we should begin migrating the front end APIs to use the APIs provided by the kube API server where possible / appropriate.

This issue is to track the following APIs:
- [ ] createCredential
- [ ] deleteCredential
- [ ] getCredential
- [ ] getCredentials
- [ ] updateCredential

Also:
- [ ] delete any unused backend code once these are migrated

Megan-Wright:
PR Here: https://github.com/tektoncd/dashboard/pull/441

a-roberts:
Carolyn's going to partner up with Megan on this one, we're figuring out how to do the patching/unpatching of the service account through the Kube frontend API atm

a-roberts:
Merged so closing

PR #523, opened 2019-09-16
Use kube api for secret adding and deleting

CarolynMabbott:
# Changes

Uses the Kube Api to add and delete secrets
Patches service accounts when adding secrets using kube api
Gets up and updates secrets 
Unpatches multiple service accounts when deleting a secret 
Updated and added tests to test everything 

References #320 

 @Megan-Wright gets credit as well 😃 

# Submitter Checklist

review them:

for more details._


src/actions/secrets.test.js
```diff
@@ -24,35 +24,71 @@ import {
...
+    uid: 'ced1c79d-c991-11e9-a380-025000000001'
+  },
+  secrets: [
+    {
+      name: 'default-token-kbn7j'
+    }
+  ]
 };
 const response = { ok: true, status: 204 };
+// const response1 = { ok: true, status: 200 };
```
a-roberts: Can be cleaned up

src/actions/secrets.test.js
```diff
@@ -64,7 +100,7 @@ it('fetchSecretsSuccess', () => {
 
 it('fetchSecrets', async () => {
   const secrets = data;
-  const secretsFromatted = dataFormatted;
+  const secretsFormatted = dataFormatted;
```
a-roberts: 😸

src/api/comms.js
```diff
@@ -43,7 +51,7 @@ export function checkStatus(response = {}) {
   throw error;
 }
 
-export function request(uri, options = defaultOptions) {
+export async function request(uri, options = defaultOptions) {
```
a-roberts: Woah, lots are calling this function - is this definitely safe to make async?

src/api/comms.js
```diff
@@ -72,6 +80,48 @@ export function put(uri, body) {
   });
 }
 
+export function patchAddSecretBody(secretName) {
```
a-roberts: Method name's a little confusing, is this generateBodyForSecretPatching?

src/api/comms.js
```diff
@@ -72,6 +80,48 @@ export function put(uri, body) {
...
+      value: {
+        name: secretName
+      }
+    }
+  ];
+
+  return patchAddBody;
+}
+
+export function patchRemoveSecretBody(indexOfSecret) {
```
a-roberts: Why's it by index and not secret name? Can you add a comment to justify this? We could instead, for example, accept a secret name here and "internally" get its index through our own code so this is a nicer API for others to use

src/api/comms.test.js
```diff
@@ -32,6 +43,80 @@ describe('getHeaders', () => {
   });
 });
 
+describe('getPatchHeaders', () => {
```
a-roberts: Can you add a comment as to why these tests are needed?

src/api/comms.test.js
```diff
@@ -32,6 +43,80 @@ describe('getHeaders', () => {
...
+    };
+    const result = getPatchHeaders(customHeaders);
+    expect(result).toMatchObject(customHeaders);
+    expect(result).toMatchObject(getPatchHeaders());
+  });
+});
+
+describe('patchAddSecretBody', () => {
+  it('patchAddSecretBody', () => {
+    const secretName = 'Groot';
```
a-roberts: 👽 

Can you change the names in both the describe and the test name itself to be more... "x should do y"?

src/api/comms.test.js
```diff
@@ -32,6 +43,80 @@ describe('getHeaders', () => {
...
+        }
+      }
+    ];
+    const result = patchAddSecretBody(secretName);
+    expect(result).toMatchObject(secretResponse);
+    expect(result).toMatchObject(patchAddSecretBody(secretName));
+  });
+});
+
+describe('patchAddSecret', () => {
```
a-roberts: Same comment about names here

src/api/comms.test.js
```diff
@@ -32,6 +43,80 @@ describe('getHeaders', () => {
...
+    };
+    fetchMock.mock(uri, data);
+    return patchAddSecret(uri, data).then(response => {
+      expect(response).toEqual(data);
+      fetchMock.restore();
+    });
+  });
+});
+
+describe('patchRemoveSecretBody', () => {
```
a-roberts: Same comment about names

src/api/comms.test.js
```diff
@@ -32,6 +43,80 @@ describe('getHeaders', () => {
...
+  it('patchRemoveSecretBody', () => {
```
a-roberts: Not the easiest of tests to read and comprehend, can you add a comment here/appropriate test name so it's more clear/easier to maintain?

src/api/comms.test.js
```diff
@@ -32,6 +43,80 @@ describe('getHeaders', () => {
...
+        path: `serviceaccount/secrets/${indexOfSecret}`
+      }
+    ];
+    const result = patchRemoveSecretBody(1);
+    expect(result).toMatchObject(secretResponse);
+    expect(result).toMatchObject(patchRemoveSecretBody(indexOfSecret));
+  });
+});
+
+describe('patchRemoveSecret', () => {
```
a-roberts: Ditto about names

src/api/index.js
```diff
@@ -343,6 +358,82 @@ export function getInstallProperties() {
...
+) {
+  const uri = getKubeAPI('serviceaccounts', {
+    name: serviceAccountName,
+    namespace
+  });
+  const patch1 = await patchAddSecret(uri, secretName);
+  return patch1;
+}
+
+// Get list of service accounts where secrets patched to, returns service account in full
```
a-roberts: Err, bit of a mouthful to read and confusing - come up with a better method name please 😅

src/api/index.js
```diff
@@ -343,6 +358,82 @@ export function getInstallProperties() {
...
+  return secretServiceAccountList;
+}
+
+export function getIndexOfSecret(sa, secretName) {
+  let x = 0;
+
+  let indexOfSecret = -1;
+  sa.forEach(element => {
+    if (element.name === secretName) {
+      indexOfSecret = x;
```
a-roberts: Add a break after this?

src/api/index.js
```diff
@@ -343,6 +358,82 @@ export function getInstallProperties() {
...
+    saList.forEach(element => {
+      const saSecretList = element.secrets;
+      let secretFound = false;
+      saSecretList.forEach(element1 => {
+        if (element1.name === secretName) {
+          secretFound = true;
+        }
+      });
+      if (secretFound === true) {
+        secretServiceAccountList.push(element);
```
a-roberts: Why not add the push in the above if and omit this?

src/api/index.js
```diff
@@ -343,6 +358,82 @@ export function getInstallProperties() {
...
+
+  return indexOfSecret;
+}
+
+export async function getIndexAndRemove(sa, secretName, namespace) {
+  const indexOfSecret = await getIndexOfSecret(sa.secrets, secretName);
+
+  // Should never be -1 as means the secret is not found
+  if (indexOfSecret === -1) {
+    return 'Not Found';
```
a-roberts: Should this instead throw an error which we can catch and do stuff with later? Instead of returning a string

config_frontend/setupTests.js
```diff
@@ -15,4 +15,5 @@ import 'react-testing-library/cleanup-after-each';
 import fetchMock from 'fetch-mock';
 
 fetchMock.catch();
+fetchMock.config.warnOnFallback = false;
```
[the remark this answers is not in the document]
AlanGreene: This should be reverted. The reason for the warning is to identify tests where network requests are not being stubbed as they should. We should aim to clean these up rather than ignoring them as they may lead to flaky tests in some environments.

pkg/router/router.go
```diff
@@ -77,13 +77,17 @@ func registerKubeAPIProxy(r endpoints.Resource, container *restful.Container) {
 	proxy.Consumes(restful.MIME_JSON, "text/plain").
 		Produces(restful.MIME_JSON, "text/plain").
 		Path("/proxy")
+	proxy.Consumes(restful.MIME_JSON, "application/json-patch+json").
```
AlanGreene: can `"application/json-patch+json"` be added to the list above instead?
is it really needed for the `Produces` call too? Which APIs produce Patch JSON as output?

src/actions/secrets.js
```diff
@@ -50,6 +81,7 @@ export function deleteSecret(secrets) {
     dispatch({ type: 'SECRET_DELETE_REQUEST' });
     const deletePromises = secrets.map(secret => {
       const { name, namespace } = secret;
+      unpatchServiceAccount(name, namespace);
```
AlanGreene: should we be waiting for this before deleting the credential or is it safe to do both in parallel?

src/actions/secrets.js
```diff
@@ -80,6 +111,15 @@ export function createSecret(postData, namespace) {
     } catch (error) {
       dispatch({ type: 'SECRET_CREATE_FAILURE', error });
     }
+    const serviceAccount = await getServiceAccount(
+      postData.metadata.labels.serviceAccount,
+      namespace
+    );
+    patchServiceAccount(
```
AlanGreene: should we be waiting for / returning this?

src/api/index.js
```diff
@@ -343,6 +358,82 @@ export function getInstallProperties() {
   return get(uri);
 }
 
+
+export async function patchServiceAccount(
+  serviceAccountName,
```
AlanGreene: Can we use named parameters instead of positional ones? This is consistent with most of our existing APIs and generally tends to produce more readable and flexible code. So this function would accept an object as a parameter instead:

`... patchServiceAccount({ serviceAccountName, namespace, secretName }) { ...`


Same for `getServiceAccount` above

src/api/index.js
```diff
@@ -333,6 +340,14 @@ export function getNamespaces() {
   return get(uri).then(checkData);
 }
 
+export function getServiceAccount(name1, namespace1) {
```
```js
export function getServiceAccount({ name, namespace }) {
  const uri =getKubeAPI('serviceaccounts', { name, namespace });
  return get(uri);
}
```

src/api/index.js
```diff
@@ -343,6 +358,82 @@ export function getInstallProperties() {
...
+      if (secretFound === true) {
+        secretServiceAccountList.push(element);
+      }
+    });
+  }
+
+  return secretServiceAccountList;
+}
+
+export function getIndexOfSecret(sa, secretName) {
```
AlanGreene: first param here looks like it's an array of secrets, not a service account as the param name would suggest

also why are we using index here, seems odd. Is the order guaranteed? Should we not be using name / uid or something more stable?

if we are sticking with index, [findIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) is a native implementation

src/api/index.js
```diff
@@ -343,6 +358,82 @@ export function getInstallProperties() {
...
+      indexOfSecret = x;
+    }
+    x += 1;
+  });
+
+  return indexOfSecret;
+}
+
+export async function getIndexAndRemove(sa, secretName, namespace) {
+  const indexOfSecret = await getIndexOfSecret(sa.secrets, secretName);
```
AlanGreene: `getIndexOfSecret` is not an async function, so `await` is not needed here

src/components/SecretsTable/SecretsTable.test.js
```diff
@@ -15,7 +15,7 @@ import React from 'react';
 import { fireEvent, render } from 'react-testing-library';
 import SecretsTable from './SecretsTable';
 
-it('SecretsTable renders with no secrets', () => {
+it('SecretsTable renders with no secrets', async () => {
```
AlanGreene: I don't see any use of `await` here, `async` shouldn't be necessary

src/reducers/index.js
```diff
@@ -289,7 +289,10 @@ export function getSecrets(
   state,
   { namespace = getSelectedNamespace(state) } = {}
 ) {
-  return secretSelectors.getSecrets(state.secrets, namespace);
+  const secretsList = secretSelectors.getSecrets(state.secrets, namespace);
+  return secretsList.filter(secret => {
```
AlanGreene: rather than filter on the client we should request only the secrets of interest

we can filter by type on the API, just like we did on the back end using the fieldSelector, e.g.:

http://localhost:9097/proxy/api/v1/namespaces/default/secrets?fieldSelector=type=kubernetes.io/basic-auth

AlanGreene:
@carlos-logro fyi, some changes here to the secrets action creators. I don't think there'll be any conflict with your changes in https://github.com/tektoncd/dashboard/pull/510 but probably best if you both sync up on this to be sure it's all good.

config_frontend/setupTests.js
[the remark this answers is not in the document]
Megan-Wright: We have  reverted this change and both the functionality and tests are still working so we will leave it out.

pkg/router/router.go
Megan-Wright: We have added this to the list above, however we definitely need to keep both the consumes and produces parts as without these the patching does not work.

CarolynMabbott: Screenshot 2019-09-16 at 14 18 57

src/actions/secrets.js
```diff
@@ -11,9 +11,38 @@ See the License for the specific language governing permissions and
...
+  createCredential,
+  deleteCredential,
+  getCredentials,
+  getServiceAccount,
+  patchServiceAccount,
+  unpatchServiceAccount
+} from '../api';
 import { getSelectedNamespace } from '../reducers';
 
+export function getAPIRoot() {
```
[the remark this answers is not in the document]
Megan-Wright: We added these in for manually testing our code and forgot to remove them - have done this now.

src/actions/secrets.js
```diff
@@ -25,15 +54,17 @@ export function fetchSecrets({ namespace } = {}) {
   return async (dispatch, getState) => {
     dispatch({ type: 'SECRETS_FETCH_REQUEST' });
     let secrets;
+
     try {
       const selectedNamespace = namespace || getSelectedNamespace(getState());
       secrets = await getCredentials(selectedNamespace);
       const secretsFormatted = [];
-      secrets.forEach(secret => {
+      secrets.items.forEach(secret => {
```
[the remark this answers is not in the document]
Megan-Wright: The SecretsDisplayTable shows the correct information.
The KubeAPI changed the reply so we had to add .items to retain the format the function required.

src/actions/secrets.js
Megan-Wright: We must wait for getServiceAccount as it is needed for the patching.

src/actions/secrets.test.js
Megan-Wright: Cleaned up :)

src/actions/secrets.test.js
Megan-Wright: 🐶

src/api/comms.js
Megan-Wright: We have reverted it to a non-async function and the code and tests both work :)

src/api/comms.js
Megan-Wright: We've updated this to the name you suggested

src/api/comms.js
CarolynMabbott: The way to remove items through the KubeApi doesnt allow you to remove with a name, it must be index. We have the function indexOfSecret(sa, secretName) which gets the index of the secret name to remove

src/actions/secrets.js
AlanGreene: yes but I mean should we be waiting for the patch operation to complete before returning from this API function? or return the promise from the patch API so the consumer of `createSecret` can wait. As it stands, this function will return immediately once `patchServiceAccount` is called, and will not wait for it to complete, whether it succeeds / fails.

carlos-logro:
@AlanGreene just checked this & there shouldn't be no conflicts with PR #510. Other than that everything looks good for me aside from yours & Adam's comments 👍

src/api/comms.test.js
Megan-Wright: Done 😸

src/api/comms.test.js
Megan-Wright: Done 😸

src/api/comms.test.js
Megan-Wright: Done 😸

src/api/comms.test.js
Megan-Wright: Done 😸

src/api/comms.test.js
Megan-Wright: Done 😸

src/api/index.js
CarolynMabbott: Done 😸

src/api/index.js
CarolynMabbott: Have tried this with the code above however it failed when we run it. 
When we console.log() the name and namespace they both were undefined 

It worked when we did it as (name, namespace) instead of (name1, namespace1)

src/api/index.js
CarolynMabbott: Done 😸

src/api/index.js
AlanGreene: to use the code I provided you'd also need to update where it's called to pass in the params as an object

src/actions/secrets.js
CarolynMabbott: It's safe to do both these in parallel

src/api/index.js
CarolynMabbott: Done

src/api/index.js
Megan-Wright: The order is guaranteed as it does not change unless we add or delete secrets. Because of the way the patch command works in the KubeAPI, it is impossible to remove by name or anything more stable (we have tried this) and therefore we must remove by index. However this works fine as the correct index will always be able to be found.

src/reducers/index.js
CarolynMabbott: Added your code in, thank you

src/api/index.js
Megan-Wright: Done 😸

src/actions/secrets.js
Megan-Wright: done 😸

src/api/index.js
Megan-Wright: done 😸

pkg/router/router.go
AlanGreene: Looks like this change may have been reverted, `proxy.Consumes` still appears twice

src/actions/secrets.js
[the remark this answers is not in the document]
AlanGreene: Looks like this change may also have been reverted

src/actions/secrets.js
AlanGreene: I'm not seeing the change here

CarolynMabbott:
@AlanGreene  It was a rebase gone very wrong so lost most of the changes 
New PR is https://github.com/tektoncd/dashboard/pull/634

PR #532, opened 2019-09-17
Add pending state and function to generate pipeline run name

steveodonovan:
# Changes

Adds parameter to generate pipeline run name in list view. Adds pending state to pipeline runs without status.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRuns/PipelineRuns.js
+++ b/packages/components/src/components/PipelineRuns/PipelineRuns.js
@@ -36,6 +36,8 @@ import './PipelineRuns.scss';
 const PipelineRuns = ({
   cancelPipelineRun,
   createPipelineRunURL = urls.pipelineRuns.byName,
+  createPipelineRunDisplayName = ({ pipelineRunMetadata }) =>
+    pipelineRunMetadata.name,
   createPipelineRunsByPipelineURL = urls.pipelineRuns.byPipeline,
   intl,
   pipelineName,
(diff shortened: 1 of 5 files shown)


packages/components/src/components/PipelineRuns/PipelineRuns.js
```diff
@@ -36,6 +36,7 @@ import './PipelineRuns.scss';
 const PipelineRuns = ({
   cancelPipelineRun,
   createPipelineRunURL = urls.pipelineRuns.byName,
+  createPipelineRunName = ({ pipelineRunMetadata }) => pipelineRunMetadata.name,
```
AlanGreene: maybe `createPipelineRunDisplayName` since we're not actually modifying the name of the run?

packages/components/src/components/PipelineRuns/PipelineRuns.js
```diff
@@ -134,9 +136,9 @@ const PipelineRuns = ({
               data-status={status}
             >
               {getStatusIcon({ reason, status })}
-              {pipelineRun.status.conditions
+              {pipelineRun.status && pipelineRun.status.conditions
                 ? pipelineRun.status.conditions[0].message
-                : ''}
+                : 'Pending'}
```
AlanGreene: intl

packages/utils/src/utils/index.js
```diff
@@ -113,7 +113,8 @@ export function getStatus(resource) {
 
 export function isRunning(reason, status) {
   return (
-    status === 'Unknown' && (reason === 'Running' || reason === 'Building')
+    !status ||
```
AlanGreene: Should we not differentiate between 'pending' and 'running'? Showing a running icon could be a bit misleading if a run is just queued but hasn't been applied yet.

steveodonovan: Sounds good, Mind pointing me towards a suitable carbon icon ?

AlanGreene: do we need an icon at all for this state? I'll have a look though, give me a while

AlanGreene: https://www.carbondesignsystem.com/guidelines/iconography/library/

Since we're already using the 'checkmark' and 'error' icons from the status block we might consider 'undefined' or 'unknown'
image

However, I think 'time' might also work:
image

PR #564, opened 2019-09-26
Issue 552: Disable dropdowns if namespace is null

dibbles:
# Changes

Introduced an isDisabled() method which returns true if the namespace is set to "".  The Pipelines and Service Accounts dropdowns then use this method to check whether or not they can be accessed.

Test updated to include checking that dropdowns are disabled and then enabled on selction of namespace.

# Submitter Checklist

review them:

for more details._


a-roberts:
I've seen this on @dibbles's laptop, behaves reasonably and intuitively (prompting users to select a namespace first) so let's take this

/lgtm
/approve
/meow space

Issue #567, opened 2019-09-26
Rebuild button fails intermittently (code 500 for backend, no page load for frontend sometimes)


CarolynMabbott:
# Expected Behavior
I can rebuild a rebuild of a PipelineRun as many times as I want 

# Actual Behavior
I tried to rebuild a rebuild of a PipelineRun and it failed after the second rebuild 
I then tried to replicate this and it failed after 9 times 
It may be intermittent issue 

# Steps to Reproduce the Problem

1. Rebuild a rebuild of a PipelineRun
2. Issue may happen / may not happen 
3.

# Additional Info
Get this error in the UI / console 

![Screenshot 2019-09-26 at 15 15 14](https://user-images.githubusercontent.com/32342288/65696037-78411180-e070-11e9-9778-13cbf407f7c4.png)

Get this in the logs 
<img width="1440" alt="Screenshot 2019-09-26 at 15 15 54" src="https://user-images.githubusercontent.com/32342288/65696083-90b12c00-e070-11e9-9208-14af1baf2fa2.png">

a-roberts:
After five rebuilds of rebuilds I hit this

```
Failed to load resource: the server responded with a status of 404 (Not Found)
vendors.07998c1fadf87f5e0aec.js:91 TypeError: Cannot read property 'steps' of undefined
    at main.b8467c49b3ec32ae6748.js:1
    at Array.map (<anonymous>)
    at _.loadTaskRuns (main.b8467c49b3ec32ae6748.js:1)
    at _.value (main.b8467c49b3ec32ae6748.js:1)
    at ya (vendors.07998c1fadf87f5e0aec.js:91)
    at ma (vendors.07998c1fadf87f5e0aec.js:91)
    at Ys (vendors.07998c1fadf87f5e0aec.js:91)
    at Bs (vendors.07998c1fadf87f5e0aec.js:91)
    at Ls (vendors.07998c1fadf87f5e0aec.js:91)
    at As (vendors.07998c1fadf87f5e0aec.js:91)
Aa @ vendors.07998c1fadf87f5e0aec.js:91
main.b8467c49b3ec32ae6748.js:1 Uncaught (in promise) TypeError: Cannot read property 'steps' of undefined
    at main.b8467c49b3ec32ae6748.js:1
    at Array.map (<anonymous>)
    at _.loadTaskRuns (main.b8467c49b3ec32ae6748.js:1)
    at _.value (main.b8467c49b3ec32ae6748.js:1)
[...]

PR #588, opened 2019-09-30
Resiliency: function even with missing status & TaskRun info for PipelineRun page

a-roberts:
# Changes
For https://github.com/tektoncd/dashboard/issues/567

Without this change, every fifth(ish) rebuild would fail to render due to either being unable to find .steps from something undefined, or taskruns. With this change, add hardening (returning empty in some cases, or a small amount of data so we can render the page in others), and then rely on the websocket and new data coming in to update the components accordingly.

Tested with over 20 rebuilt runs in a row and noticing the components continue to be updated with new PipelineRun information accordingly.

Also includes a backend change to generate a more unique suffix (delegating to the k8s GenerateName feature).

# Submitter Checklist

review them:

Nope - sorry, desperate to get this in for 0.2.0! Thinking about adding...

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRun/PipelineRun.js
+++ b/packages/components/src/components/PipelineRun/PipelineRun.js
@@ -16,7 +16,7 @@ import {
   InlineNotification,
   StructuredListSkeleton
 } from 'carbon-components-react';
-import { injectIntl } from 'react-intl';
+import { FormattedMessage, injectIntl } from 'react-intl';
 import {
   getErrorMessage,
   getStatus,
@@ -42,6 +42,11 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
 
   loadPipelineRunData = () => {
     const { pipelineRun } = this.props;
+
+    if (!(pipelineRun.status && pipelineRun.status.taskRuns)) {
+      return { pipelineRun };
+    }
+
     const {
       status: { taskRuns: taskRunsStatus }
     } = pipelineRun;
@@ -56,14 +61,23 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
   };
 
   loadTaskRuns = pipelineRun => {
-    if (!pipelineRun) {
+    if (!pipelineRun || !pipelineRun.status || !pipelineRun.status.taskRuns) {
       return [];
     }
 
     const { tasks, taskRuns } = this.props;
+
+    if (!tasks || !taskRuns) {
+      return [];
+    }
+
     const {
       status: { taskRuns: taskRunDetails }
     } = pipelineRun;
+
+    if (!pipelineRun.status || !pipelineRun.status.taskRuns) {
+      return [];
+    }
 
     return taskRuns
       .map(taskRun => {
@@ -78,11 +92,35 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
           namespace: taskRunNamespace,
           annotations
         } = taskRun.metadata;
+
         const { reason, status: succeeded } = getStatus(taskRun);
+
         const { pipelineTaskName } = taskRunDetails[taskRunName] || {};
         const { params, resources: inputResources } = taskRun.spec.inputs;
         const { resources: outputResources } = taskRun.spec.outputs;
-        const steps = stepsStatus(task.spec.steps, taskRun.status.steps);
+
+        let steps = '';
+
+        if (!task.spec || !taskRun.status) {
+          steps = {
+            id: 0,
+            reason: 'unknown',
+            status: 'unknown'
+          };
+          let theRun = taskRun;
+          theRun = {
+            metadata: {
+              uid: '0'
+            },
+            status: {
+              podName: 'unknown'
+            }
+          };
+          taskRun = theRun; // eslint-disable-line no-param-reassign
+        } else {
+          steps = stepsStatus(task.spec.steps, taskRun.status.steps);
+        }
+
         return {
           id: taskRun.metadata.uid,
           pipelineTaskName,
@@ -185,6 +223,16 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
       );
     }
     const taskRuns = this.loadTaskRuns(pipelineRun, taskRunNames);
+
+    if (taskRuns.length === 0) {
+      return (
+        <FormattedMessage
+          id="dashboard.taskRun.noTaskRuns"
+          defaultMessage="No TaskRuns found for this PipelineRun yet…"
+        />
+      );
+    }
+
     const taskRun = selectedTaskRun(selectedTaskId, taskRuns) || {};
 
     const { definition, reason, status, stepName, stepStatus } = taskRunStep(
@@ -199,6 +247,7 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
         stepStatus={stepStatus}
       />
     );
+
     return (
       <>
         <RunHeader
--- a/packages/components/src/components/PipelineRun/PipelineRun.test.js
+++ b/packages/components/src/components/PipelineRun/PipelineRun.test.js
@@ -106,3 +106,54 @@ it('PipelineRunContainer handles error state', async () => {
   );
   await waitForElement(() => getByText('Error loading PipelineRun'));
 });
+
+// A scenario exists (typically with rebuild) where all PipelineRun data
+// is not yet available. Verify that the container still renders OK
+it('PipelineRunContainer handles no TaskRuns found yet', async () => {
+  const match = {
+    params: {
+      pipelineName: 'foo',
+      pipelineRunName: 'bar'
+    }
+  };
+
+  const middleware = [thunk];
+  const mockStore = configureStore(middleware);
+
+  const testStore = mockStore({
+    tasks: {
+      byNamespace: {}
+    },
+    taskRuns: {
+      // Nothing here
+      byId: {},
+      byNamespace: {},
+      errorMessage: null,
+      isFetching: false
+    },
+    namespaces: {
+      selected: 'default'
+    },
+    pipelineRuns: {
+      byId: {},
+      byNamespace: { default: {} },
+      errorMessage: null,
+      isFetching: false
+    }
+  });
+
+  const { getByText } = renderWithIntl(
+    <Provider store={testStore}>
+      <PipelineRun
+        match={match}
+        error={null}
+        fetchTaskRuns={() => Promise.resolve()}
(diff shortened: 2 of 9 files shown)


a-roberts:
Got a couple legit unit test failures so on the case

packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -47,6 +47,15 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
 
   loadPipelineRunData = () => {
     const { pipelineRun } = this.props;
+
+    if (!pipelineRun.status) {
+      return {};
+    }
+
+    if (!pipelineRun.status.taskRuns) {
```
AlanGreene: let's combine these:

```js
if (!(pipelineRun.status && pipelineRun.status.taskRuns)) {
```

can we return more than just an empty object? e.g. at least `{ pipelineRun }`

packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -65,11 +74,36 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
       return [];
     }
 
+    if (!pipelineRun.status) {
```
AlanGreene: let's combine these:

```js
if (!(pipelineRun && pipelineRun.status && pipelineRun.status.taskRuns && tasks && taskRuns)) {
  return [];
}
```

packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -65,11 +74,36 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
...
+
+    if (!taskRuns) {
+      return [];
+    }
+
     const {
       status: { taskRuns: taskRunDetails }
     } = pipelineRun;
 
+    if (!pipelineRun.status) {
```
AlanGreene: shouldn't these be caught by the checks above?

a-roberts:
Unit tests now fixed, displaying a lovely error message if it can't be loaded (verified by going to the PipelineRuns page for a non-existent PipelineRun I just deleted), linter errors addressed too

packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -83,11 +117,26 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
...
         const { reason, status: succeeded } = getStatus(taskRun);
+
         const { pipelineTaskName } = taskRunDetails[taskRunName] || {};
         const { params, resources: inputResources } = taskRun.spec.inputs;
         const { resources: outputResources } = taskRun.spec.outputs;
-        const steps = stepsStatus(task.spec.steps, taskRun.status.steps);
+
+        let steps = '';
+
+        if (!task.spec || !taskRun.status) {
```
AlanGreene: do you have a scenario that reliably gets us here? I'd like to see if there are other options as introducing fake data here can make this less flexible, especially when looking at how others might integrate the component

packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -190,6 +239,11 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
       );
     }
     const taskRuns = this.loadTaskRuns(pipelineRun, taskRunNames);
+
+    if (taskRuns === []) {
```
AlanGreene: `===` won't work here, `taskRuns.length` should be ok

packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -190,6 +239,11 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
...
+      return <div>No task runs found for this PipelineRun</div>;
```
AlanGreene: `TaskRuns`

also should be using react-intl here to externalise the string, either `<FormattedMessage ... />` or `intl.formatMessage(...)`

packages/components/src/components/TaskTree/TaskTree.js
```diff
@@ -23,9 +23,17 @@ class TaskTree extends Component {
 
   render() {
     const { selectedTaskId, taskRuns } = this.props;
+
+    if (!taskRuns) {
```
AlanGreene: ideally we'd catch this higher up the tree, but if we want to render nothing we can just return `null`

packages/utils/src/utils/index.js
```diff
@@ -32,11 +32,17 @@ export function getErrorMessage(error) {
...
-  return tasks.find(t => t.metadata.name === selectedTaskName);
+  if (tasks) {
+    return tasks.find(t => t.metadata.name === selectedTaskName);
+  }
+  return {};
 }
 
 export function selectedTaskRun(selectedTaskId, taskRuns = []) {
-  return taskRuns.find(run => run.id === selectedTaskId);
+  if (taskRuns) {
```
AlanGreene: taskRuns should always be defined (we default to `[]`) here unless we're somehow passing `null` instead of `undefined`

if that's the case it should be caught before we get here
similar for `selectedTask` above, ideally we wouldn't get that far without tasks

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -97,6 +97,21 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
 
   render() {
     const { error, intl, match, pipelineRun, tasks, taskRuns } = this.props;
+
+    if (!this.props.pipelineRun) {
```
AlanGreene: no need for `this.props`, we've already deconstructed the props to extract `pipelineRun` so you can reference it directly here and below

a-roberts:
> Looks like we have quite a few weak spots, some of these changes will certainly help but I think it might be worth looking at another approach (separate from this PR) in the longer term.
> 
> Added some comments / suggestions / questions. Feel free to reach out on slack.

Yeah definitely, at this point it was a little "trial and error" to try and figure out all of the variables that aren't end up set for some reason. I'd like it if instead we continued to be a loading state while the remaining PipelineRun data makes it into the store

Will address the comments now, thanks for taking a look

packages/components/src/components/TaskTree/TaskTree.js
a-roberts: Weird, with this and the below (index.js change) I can reliably get `Uncaught TypeError: steps.map is not a function`, keeping this unchanged for now I think 🤔

packages/components/src/components/PipelineRun/PipelineRun.js
a-roberts: Hmmmmmm, well I can reliably get taskRuns without a status coming in yeah, it's with five-ish rebuild runs in a row (you just keep rebuilding then clicking through to view that new one, then rebuild that, etc)

Same goes for the task steps, it's like the data just doesn't come in on time or we don't populate the store on time (and yet we do an async and wait on the response, so I'm guessing it's the websocket)...

a-roberts:
Ready for another look please @AlanGreene

src/containers/PipelineRun/PipelineRun.js
AlanGreene: missed this one? few of these:

```diff
-this.props.pipelineRun
+pipelineRun
```

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -97,6 +97,21 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
...
+      return <PipelineRun error="PipelineRun not found" loading={false} />;
+    }
+
+    if (!this.props.pipelineRun.status) {
+      pipelineRun.status = 'unknown';
```
AlanGreene: didn't spot this earlier but looks like something fishy going on here?

`status = 'unknown'`
then immediately below
`status = somethingElse`

was this supposed to be updating `reason` or perhaps a `status` field somewhere else?

packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -190,6 +228,16 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
       );
     }
     const taskRuns = this.loadTaskRuns(pipelineRun, taskRunNames);
+
+    if (taskRuns.length === 0) {
+      return (
+        <FormattedMessage
+          id="dashboard.taskRun.noTaskRuns"
+          defaultMessage="No TaskRuns found for this PipelineRun yet..."
```
AlanGreene: can we use an actual ellipsis here? `…`

packages/components/src/components/PipelineRun/PipelineRun.test.js
```diff
@@ -106,3 +106,54 @@ it('PipelineRunContainer handles error state', async () => {
   );
   await waitForElement(() => getByText('Error loading PipelineRun'));
 });
+
+// A scenario exists (typically with rebuild) where all PipelineRun data
```
AlanGreene: 👍

src/containers/PipelineRun/PipelineRun.js
a-roberts: Hmmmmmm, so what's there is intentional - if there's no status, set a status with no taskruns for that field

If there was actually a status anyway we enter the .status.taskRuns check and we provide an empty array of taskruns there

a-roberts:
^^ fixed a few comments and added a note for a query

src/containers/PipelineRun/PipelineRun.js
AlanGreene: within this single conditional block you assign a value to status, then immediately reassign it

so the `status= 'unknown'` is redundant

a-roberts: True, addressed and pushed, oops

a-roberts:
Ready for another look please 😅

jest.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
   coverageThreshold: {
     global: {
       statements: 90,
-      branches: 90,
+      branches: 85,
```
AlanGreene: @steveodonovan and I did some work earlier today to bring our coverage back up to a safer level. Can you revert this please? (I've verified your PR is fine locally with this at 90). It's too easy to let this stuff slip otherwise.

a-roberts: Great, yeah will do! Pained me to do this, thanks

a-roberts: Right I've rebased and noticed 90+ percent coverage across the board now on average

Issue #549, opened 2019-09-25
Namespace on the left in dashboard overwrites namespace in the create secret modal


CarolynMabbott:
When creating a secret the namespace on the left overwrites the namespace when you select namespace when creating a secret 
This means that it does not fetch secrets for the namespace selected in the create secret, only for the namespace on the left 

Below shows me attempting to create a secret in namespace tekton-pipelines when in the left is in namespace car 

![Screenshot 2019-09-25 at 13.55.25.png](https://images.zenhubusercontent.com/5cfe69e3171db04ec3637b76/22879ae5-6d86-4a1a-8c9e-7b03349624d4)

PR #612, opened 2019-10-02
Namespace on the left overwriting namespace in the create secret modal

carlos-logro:
issue: #549 

# Changes

the action `fetchServiceAccounts` was not being used correctly. Was fetching them when the component mounted without passing the selected namespace as a parameter. Changed it to make the fetch after namespace is selected.

### DIFF
--- a/src/components/SecretsModal/BasicAuthFields.js
+++ b/src/components/SecretsModal/BasicAuthFields.js
@@ -26,15 +26,13 @@ const BasicAuthFields = props => {
     serviceAccounts
   } = props;
 
-  const saItems = serviceAccounts
-    .filter(sa => sa.metadata.namespace === namespace)
-    .map(sa => (
-      <SelectItem
-        value={sa.metadata.name}
-        text={sa.metadata.name}
-        key={`${sa.metadata.namespace}:${sa.metadata.name}`}
-      />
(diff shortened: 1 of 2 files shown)


src/components/SecretsModal/BasicAuthFields.js
```diff
@@ -26,15 +26,13 @@ const BasicAuthFields = props => {
...
-  const saItems = serviceAccounts
-    .filter(sa => sa.metadata.namespace === namespace)
-    .map(sa => (
-      <SelectItem
-        value={sa.metadata.name}
-        text={sa.metadata.name}
-        key={`${sa.metadata.namespace}:${sa.metadata.name}`}
-      />
-    ));
+  const saItems = serviceAccounts.map(sa => (
```
AlanGreene: since the `namespace` prop is no longer being used it should be removed from line 22 and also from any tests / other code that are currently passing it.

carlos-logro: the namespace prop is used in `line 69` to see if the Service Account Dropdown should be disabled or not.

AlanGreene: ah so it is 👍

Issue #548, opened 2019-09-25
Secret creation: Cannot type in server URL field after the 10th server url


Megan-Wright:
Expected Behaviour:
We should be able to add as many server URLs for one secret as we would like.

Actual Behaviour: 
When creating a secret, the UI allows for seemingly unlimited server URLs to be added for one secret. However, after the 10th server url, it is not possible to type a URL into the field.
Up to "tekton.dev/git-9" we can edit both of the fields for each server URL, however after this point (from "tekton.dev/git-10") neither of the fields are able to be edited. This includes pasting a URL into the field. When you press "Submit", there is an error that the fields are empty. If you then try to type in the field again, the error disappears but no text is visible.

![Screenshot 2019-09-25 at 13.38.03.png](https://images.zenhubusercontent.com/5d25b016e0bbdc62005f7da6/d02ab8b2-18f4-4f94-9b6f-a32e5d2d714f)

<img width="487" alt="Screenshot 2019-09-25 at 13 41 22" src="https://user-images.githubusercontent.com/52741262/65601587-3a71b980-df9a-11e9-8b0f-bdf16aad4403.png">

PR #614, opened 2019-10-02
Bug Fix: Unable to type inside the 10th+ Server URL Field

carlos-logro:
issue: #548 

# Changes
Changed the approach when trying to obtain the index of the input field that is being changed. This index is then used to find the corresponding object in the array, change its value & set the state.

### DIFF
--- a/src/containers/SecretsModal/SecretsModal.js
+++ b/src/containers/SecretsModal/SecretsModal.js
@@ -238,11 +238,10 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
     });
   };
 
-  handleAnnotationChange = e => {
-    e.persist();
+  handleAnnotationChange = annotation => {
     this.setState(prevState => {
-      let { id } = e.target;
-      const { value } = e.target;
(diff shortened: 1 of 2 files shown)


src/containers/SecretsModal/SecretsModal.js
```diff
@@ -253,12 +253,14 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
         newInvalidFields.push(id);
       }
 
-      id = id
-        .split('-')
-        .splice(1, 2)
-        .join();
-      const index = id.substring(id.length - 1);
-      id = id.slice(0, -1);
+      let index;
```
AlanGreene: I don't follow what's happening here or how this addresses the issue. It could probably benefit from some comments, maybe including expected format of `id` etc.

We should avoid string manipulation to extract useful data as much as possible, preferring instead to pass objects containing the relevant fields.

carlos-logro: done!

PR #634, opened 2019-10-09
Kube API for frontend secrets

CarolynMabbott:
# Changes

Update create / delete secrets to use Kube API
Unpatches multiple service accounts when deleting secrets 

@Megan-Wright gets credit too 😸  

# Submitter Checklist

review them:

for more details._


src/api/comms.test.js
```diff
@@ -122,3 +148,62 @@ describe('post', () => {
...
+    fetchMock.mock(uri, data);
+    return patchAddSecret(uri, data).then(response => {
+      expect(response).toEqual(data);
+      fetchMock.restore();
+    });
+  });
+});
+
+describe('patchRemoveSecretBody', () => {
+  it('returns the correct index for ermoving a secret', () => {
```
a-roberts: removing*

src/api/index.js
```diff
@@ -264,30 +271,108 @@ export function rebuildPipelineRun(namespace, payload) {
...
+}) {
+  const uri = getKubeAPI('serviceaccounts', {
+    name: serviceAccountName,
+    namespace
+  });
+  const patch1 = await patchAddSecret(uri, secretName);
+  return patch1;
+}
+
+// Get list of service accounts where secrets patched to, returns service account in full
```
a-roberts: where secrets are* patched to

returns service accounts* as a list surely? not one service account in full

src/api/index.js
```diff
@@ -264,30 +271,108 @@ export function rebuildPipelineRun(namespace, payload) {
...
+}
+
+export async function unpatchServiceAccount(secretName, namespace) {
+  const saList = await getServiceAccounts({ namespace });
+  const secretSaList = await getSecretServiceAccountList(saList, secretName);
+
+  secretSaList.forEach(async element => {
+    await getIndexAndRemove(element, secretName, namespace);
+  });
+  return 'Completed unpatching';
```
a-roberts: Does this get used anywhere? Could check to see if we entered the above loop with a simple flag and say "There was nothing to do - no service account found using this secret" if so

src/api/index.test.js
```diff
@@ -12,72 +12,165 @@ limitations under the License.
...
+    resourceVersion: '333247',
+    selfLink: '/api/v1/namespaces/tekton-pipelines/serviceaccounts/default',
+    uid: '3233be5f-c9a3-11e9-a380-025000000001'
+  },
+  secrets: [
+    {
+      name: 'default-token-dcb4b'
+    }
+  ]
+};
```
a-roberts: Add a newline

src/api/index.test.js
```diff
@@ -215,17 +308,29 @@ it('getPipelineRuns With Query Params', () => {
...
+  return index.getPipelineRun({ name }).then(pipelineRun => {
+    expect(pipelineRun).toEqual(data);
+    fetchMock.restore();
+  });
+});
+
+it('deletePipelineRun', () => {
+  const name = 'foo';
+  const data = { fake: 'pipelineRun' };
+  fetchMock.delete(`end:${name}`, data);
```
a-roberts: Add a comment here so it's easier to understand/main please, what's end: do? Why's it now required?

src/api/index.test.js
```diff
@@ -235,7 +340,17 @@ it('deletePipelineRun', () => {
...
+  return index.deletePipelineRun({ name }).then(pipelineRun => {
+    expect(pipelineRun).toEqual(data);
+    fetchMock.restore();
+  });
+});
+
+it('deletePipelineRun', () => {
+  const name = 'foo';
+  const data = { fake: 'pipelineRun' };
+  fetchMock.delete(`end:${name}`, data);
```
a-roberts: Same as above wrt the end

src/api/index.test.js
```diff
@@ -12,72 +12,165 @@ limitations under the License.
...
     window.history.pushState({}, 'Title', '/path#hash');
-    expect(getAPIRoot()).toContain('/path');
+    expect(index.getAPIRoot()).toContain('/path');
   });
 });
 
 describe('getAPI', () => {
   it('returns a URI containing the given type', () => {
-    const uri = getAPI('pipelines');
+    const uri = index.getAPI('pipelines');
```
a-roberts: why is `index` prefixing now required?

src/api/index.test.js
```diff
@@ -581,34 +696,228 @@ it('getNamespaces', () => {
...
   };
   fetchMock.get(/namespaces/, data);
-  return getNamespaces().then(response => {
+  return index.getNamespaces().then(response => {
     expect(response).toEqual(data.items);
     fetchMock.restore();
   });
 });
 
+it('getServiceAccount', () => {
```
a-roberts: These test names aren't great...

src/api/index.test.js
```diff
@@ -581,34 +696,228 @@ it('getNamespaces', () => {
...
   const data = { items: 'serviceaccounts' };
   fetchMock.get(/serviceaccounts/, data);
-  return getServiceAccounts().then(response => {
+  return index.getServiceAccounts().then(response => {
     expect(response).toEqual(data.items);
     fetchMock.restore();
   });
 });
 
+it('getSecretServiceAccountList No Secrets Matched', () => {
```
a-roberts: I see the same test name later...

src/api/index.test.js
```diff
@@ -581,34 +696,228 @@ it('getNamespaces', () => {
...
+  const data = [serviceAccountPass];
+
+  return index
+    .getSecretServiceAccountList(data, 'secret-name')
+    .then(response => {
+      expect(response).toEqual(data);
+    });
+});
+
+it('getSecretServiceAccountList 2 Secret Matched', () => {
```
a-roberts: plurality, two* not 2

src/api/index.test.js
```diff
@@ -581,34 +696,228 @@ it('getNamespaces', () => {
...
+    },
+    {
+      name: 'secret-name'
+    }
+  ];
+  const response = index.getIndexOfSecret(data, 'secret-name');
+  expect(response).toEqual(1);
+});
+
+it('getIndexOfSecret secret does not exist', () => {
```
a-roberts: Test names should be of the format *verb* causes *action*, these are just collections of words atm

src/api/index.test.js
CarolynMabbott: The tests in jest require something to be in from the function to be mocked like 'index'
Which meant having to import * from index as index instead of individual functions so we had to put index. in front of them
It's how its done in the rest of the code and saw no way around doing it that way

src/api/index.js
CarolynMabbott: Returns a list of service accounts

src/api/index.js
CarolynMabbott: The return doesnt get used anywhere as its not being awaited
removed it now

src/api/index.test.js
CarolynMabbott: This is someones PR, i just added the `index.` before because had to add the index for jest to work

src/api/index.test.js
CarolynMabbott: This is someones PR, i just added the `index.` before because had to add the index for jest to work

src/api/index.js
a-roberts: My comment was about the comment 😸

src/api/index.test.js
CarolynMabbott: First one relates to if a secret isn't patched to a service account and checks it doesn't fall over and break if its not patched to a service account
Changed the name of the test now

src/api/comms.js
```diff
@@ -80,3 +88,45 @@ export function deleteRequest(uri) {
...
+    {
+      op: 'remove',
+      path: `serviceaccount/secrets/${indexOfSecret}`
+    }
+  ];
+  return patchRemoveBody;
+}
+
+export async function patchAddSecret(uri, secretName) {
+  const patchAddBody = await generateBodyForSecretPatching(secretName);
```
AlanGreene: `generateBodyForSecretPatching` is a synchronous function, no need for `await` here
same with `patchRemoveSecretBody` and `getPatchHeaders` below

once these are updated you'll also no longer need the `async` keyword on the function definitions

src/api/index.js
```diff
@@ -264,30 +271,108 @@ export function rebuildPipelineRun(namespace, payload) {
...
+  const uri = getKubeAPI('serviceaccounts', {
+    name: serviceAccountName,
+    namespace
+  });
+  const patch1 = await patchAddSecret(uri, secretName);
+  return patch1;
+}
+
+// Get list of service accounts where are secrets patched to, returns list of service accounts
+export async function getSecretServiceAccountList(saList, secretName) {
```
AlanGreene: no async operations in this function, `async` keyword should be removed

src/api/index.js
```diff
@@ -264,30 +271,108 @@ export function rebuildPipelineRun(namespace, payload) {
...
+}
+
+export async function unpatchServiceAccount(secretName, namespace) {
+  const saList = await getServiceAccounts({ namespace });
+  const secretSaList = await getSecretServiceAccountList(saList, secretName);
+
+  secretSaList.forEach(async element => {
+    await getIndexAndRemove(element, secretName, namespace);
+  });
+  return 'Completed unpatching';
```
AlanGreene: What is this return value used for? I only see it used in tests at the moment but it doesn't seem to be necessary. Instead the tests should assert that `getIndexAndRemove` (along with the other APIs) was called the expected number of times with the correct params.

package.json
```diff
@@ -82,7 +82,7 @@
     "glob": "^7.1.4",
     "html-webpack-plugin": "^3.2.0",
     "husky": "^1.3.1",
-    "jest": "^24.5.0",
+    "jest": "^24.9.0",
```
AlanGreene: was there a specific reason for the version update? not an issue, just curious

PR #659, opened 2019-10-22
Validation on import resources

CarolynMabbott:
# Changes

Adds validation for the import resources page
Checks thats its a more valid github address than before 
Added tests / checked it doesnt break anything  

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/ImportResources/ImportResources.js
+++ b/src/containers/ImportResources/ImportResources.js
@@ -38,6 +38,25 @@ async function getInstallNamespace() {
   return determineInstallNamespace();
 }
 
+function validateURL(url) {
+  if (!url.trim().startsWith('http://') && !url.trim().startsWith('https://')) {
+    return false;
+  }
+
+  if (url.trim() === '') {
+    return false;
+  }
+
+  if (url.includes('github') === false) {
+    return false;
+  }
+  if (url.includes('.') === false) {
+    return false;
+  }
+
+  return true;
+}
+
 export class ImportResources extends Component {
   constructor(props) {
     super(props);
@@ -100,9 +119,21 @@ export class ImportResources extends Component {
     const gitresourcename = 'git-source';
     const gitcommit = 'master';
 
-    if (repourl === '' || !namespace) {
+    // Without the if statement it will not display errors for both the namespace and the url at the same time
+    // The if / else statement inside is because the repourl needs different parameters set for invalidInput
+    // for the error to display when Submit is pressed. If set to the same then the errors dont appear on the page
+    const repourlValid = validateURL(repourl);
+    if (repourlValid === false || !namespace) {
+      if (repourlValid === false && repourl === '') {
+        this.setState({
+          invalidInput: repourl === ''
+        });
+      } else if (repourlValid === false) {
+        this.setState({
+          invalidInput: repourl
+        });
+      }
       this.setState({
-        invalidInput: repourl === '',
         invalidNamespace: !namespace
       });
       return;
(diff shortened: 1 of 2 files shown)


CarolynMabbott:
@googlebot I signed it!

src/containers/ImportResources/ImportResources.js
```diff
@@ -100,9 +119,18 @@ export class ImportResources extends Component {
     const gitresourcename = 'git-source';
     const gitcommit = 'master';
 
-    if (repourl === '' || !namespace) {
+    const valid = validateURL(repourl);
```
a-roberts: Comments would be great here, had to focus to understand this more than I'd have liked to

src/containers/ImportResources/ImportResources.js
```diff
@@ -100,9 +119,18 @@ export class ImportResources extends Component {
...
+    if (valid === false || !namespace) {
+      if (valid === false && repourl === '') {
+        this.setState({
+          invalidInput: repourl === ''
```
a-roberts: Wait what?

src/containers/ImportResources/ImportResources.js
```diff
@@ -100,9 +119,18 @@ export class ImportResources extends Component {
...
+        });
+      } else if (valid === false) {
+        this.setState({
+          invalidInput: repourl
```
a-roberts: How does this work?

src/containers/ImportResources/ImportResources.js
CarolynMabbott: Theres a wierd thing for the dashboard that if the repourl is empty it wont show up the error unless its specifically done setting the repourl to '' 
If i take it out then if the repourl is empty then when you press submit it wont let you submit but wont show up an error as the repourl hasnt got anything there 
The if statement is only needed for when the repourl is empty because the errors show up fine for everything else in the page but not when it is empty or contains only a space

src/containers/ImportResources/ImportResources.js
CarolynMabbott: Sorry, my bad, I'll add a comment in

src/containers/ImportResources/ImportResources.js
CarolynMabbott: It was a situation when it wouldnt return both errors about namespace and repourl without them being in the same loop
It checking that if the repourl has "valid" text in and if not and its not empty (as ifs its empty it doesnt work with setting invalidInput to repourl and if it has text in doesnt show the error on the dashboard with setting it to the state it sets it in the above if statement) then it sets it to invalidInput and errors on the page

a-roberts: Thanks, yeah I'd have no idea it does that from the code though, can you add comments to that effect please and see if you can refactor to make things any easier to maintain?

repourl presumably evaluates to true or false then? I don't mind if we add a few extra lines if it makes the code much clearer

src/containers/ImportResources/ImportResources.js
a-roberts: Alrighty, I'd use a more descriptive variable name than `valid` then and see my below comment about that repo url being empty bit (evaluating to true which I assume invalidInput wants) - again I'd prefer a few more lines to improve readability over conciseness, favour tidyness and readability where you can please

src/containers/ImportResources/ImportResources.js
CarolynMabbott: Yeah I'll add some comments in 
I'll see if i can refactor to make it better to look at / more maintainable 

repourl is the text inputted into the repository url field, thats what get used in validateURL 
valid is the result of the validateURL function which goes either yep valid or nope not valid

a-roberts: Great, thanks Carolyn that'd be useful 👍

Issue #506, opened 2019-09-06
Notification for deleting secrets


Megan-Wright:
### Expected Behaviour ###
A notification should appear on deleting a secret to confirm that the secret has been deleted.
### Actual Behaviour ### 
When a secret is deleted via the dashboard there is no notification.

vtereso:
This is also being tracked here: https://github.com/tektoncd/dashboard/issues/269.

a-roberts:
Merged so closing

PR #674, opened 2019-10-28
Add notifications for successful creation and deletion of secrets

Megan-Wright:
# Changes

For: https://github.com/tektoncd/dashboard/issues/506
Adds notifications that display on successful creation/deletion of secrets on the secrets panel. 
Updates relevant tests to check for success notifications

# Submitter Checklist

review them:

for more details._


src/containers/Secrets/Secrets.js
```diff
@@ -99,6 +111,31 @@ export /* istanbul ignore next */ class Secrets extends Component {
...
+            />
+          ))}
+        {deleteSuccess && (
+          <InlineNotification
+            kind="success"
+            title="Success:"
+            subtitle="Secret(s) deleted successfully"
+            iconDescription="Clear Notification"
+            className="notificationComponent"
+            data-testid="errorNotificationComponent"
```
carlos-logro: I did notice that each `Notification` component has the same `data-testeid`, also, asa DRY approach i'd suggest doing something like:

```
{(createSuccess || deleteSuccess) && (
             <InlineNotification
               kind="success"
               title="Success:"
               subtitle={createSuccess ? "Secret created successfully" : "Secret(s) deleted successfully"}
               iconDescription="Clear Notification"
               className="notificationComponent"
               data-testid="successNotificationComponent"
               onCloseButtonClick={this.props.clearNotification}
               lowContrast
             />
           ))}
```

src/actions/secrets.js
```diff
@@ -69,10 +69,12 @@ export function deleteSecret(secrets) {
       const deleteWithinTimePromise = Promise.race([response, timeout]);
       return deleteWithinTimePromise;
     });
-
     Promise.all(deletePromises)
       .then(() => {
-        dispatch({ type: 'SECRET_DELETE_SUCCESS' });
+        dispatch({
+          type: 'SECRET_DELETE_SUCCESS',
+          success: 'Secret(s) deleted successfully'
```
carlos-logro: Since the `subtitle` prop in `InlineNotification` is hardcoded, you really don't need to pass a string. With the type and reducer declared takes care of it.

src/reducers/secrets.js
```diff
@@ -80,9 +80,40 @@ function isFetching(state = false, action) {
...
+    default:
+      return state;
+  }
+}
+
+function createSuccessMessage(state = null, action) {
+  switch (action.type) {
+    case 'SECRET_DELETE_SUCCESS':
+    case 'SECRET_CREATE_SUCCESS':
+      return action.success;
```
carlos-logro: Once the switch statement reaches the case you can just return true here and probably false instead of null below.

src/containers/Secrets/Secrets.js
Megan-Wright: The ```data-testid``` is only used in one test, to see if an error occurs, but as you say the create and delete success I don't think need to have this property and so I have removed it from them. As for the format of the notifications, I have attempted to code it . the way you have suggested however I then get a create success notification for both creating and deleting secrets.

src/actions/secrets.js
Megan-Wright: Ok, I've removed the ```success``` property for both create and delete

src/reducers/secrets.js
Megan-Wright: Done 😸

src/containers/Secrets/Secrets.js
```diff
@@ -99,6 +111,29 @@ export /* istanbul ignore next */ class Secrets extends Component {
             lowContrast
           />
         )}
+        {createSuccess &&
+          (!deleteSuccess && (
+            <InlineNotification
+              kind="success"
+              title="Success:"
```
AlanGreene: we need to ensure all strings are externalised so they can be translated, use either `intl.formatMessage` or the `FormattedMessage` component as needed.

Once that's done `npm run i18n:extract` will ensure the message bundles are updated correctly with any new strings. These changes should also be included in the PR.

src/containers/Secrets/Secrets.test.js
```diff
@@ -127,9 +134,9 @@ it('click add new secret & modal appears', () => {
 
   jest.spyOn(API, 'getServiceAccounts').mockImplementation(() => []);
 
-  const { getByTestId } = render(
+  const { getByTestId } = renderWithRouter(
```
AlanGreene: we have a `renderWithIntl` helper so you don't need to manage this manually

src/reducers/index.js
```diff
@@ -289,6 +289,14 @@ export function getSecretsErrorMessage(state) {
   return secretSelectors.getSecretsErrorMessage(state.secrets);
 }
 
+export function createSecretsSuccessMessage(state) {
```
AlanGreene: naming... selectors should only be retrieving data from the store, so should be `get...` or similar

src/reducers/secrets.js
```diff
@@ -80,9 +80,10 @@ function isFetching(state = false, action) {
 function errorMessage(state = null, action) {
   switch (action.type) {
     case 'SECRETS_FETCH_FAILURE':
-    case 'SECRET_DELETE_FAILURE':
     case 'SECRET_CREATE_FAILURE':
       return action.error;
+    case 'SECRET_DELETE_FAILURE':
```
AlanGreene: same as the cases above, change can be reverted

src/reducers/index.js
Megan-Wright: Done 😸

src/reducers/secrets.js
Megan-Wright: I'm not sure I understand, this case is separate to the ```SECRET_CREATE_FAILURE``` case?

src/containers/Secrets/Secrets.test.js
Megan-Wright: Ok, I have changed it to use ```renderWithIntl```

src/containers/Secrets/Secrets.js
Megan-Wright: Ok, I have changed all of the user-visible strings to use ```intl.formatMessage```

src/reducers/secrets.js
AlanGreene: you're doing the same thing here (`return action.error;`) as in the cases above so there's no functional difference from what was there before, it's just duplicating the return unnecessarily

Megan-Wright: ok, reverted

a-roberts:
/lgtm
/approve
/meow space

@AlanGreene's feeling unwell too today and I'd really like to be in for the release, looks like the comments have been addressed

PR #691, opened 2019-11-04
Replace resource url with tekton-releases bucket url

devholic:
# Changes

Update release resource file links in README

## Reference
[Bucket Browser](https://console.cloud.google.com/storage/browser/tekton-releases/dashboard/latest)

# Submitter Checklist

review them:

- [N/A] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [N/A] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Tekton Dashboard is a general purpose, web-based UI for Tekton Pipelines. It all
    and its dependencies:
     
    ```bash
-   kubectl apply --filename https://github.com/tektoncd/dashboard/releases/download/v0.2.0/release.yaml
+   kubectl apply --filename https://storage.googleapis.com/tekton-releases/dashboard/latest/release.yaml
    ```
 
    _(Previous versions will be available at `previous/$VERSION_NUMBER`, e.g.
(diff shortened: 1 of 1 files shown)


a-roberts:
Hey @devholic, thanks for the PR and for getting involved!

Can you please mention the benefits for pointing to the Tekton release bucket and not a GitHub release?

With a GitHub release it's very easy to make changes if we've missed something out in hte yaml (and we have inadvertently removed the entire tekton-release bucket before, thus invalidating some download links)

devholic:
@a-roberts 

When new dashboard version is released, old urls in documents (e.g. README) required to be changed. But if we use bucket link, there is no need to making changed in documents since bucket link refers latest release.

Also, [tekton pipeline also uses bucket link](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#installing-tekton-pipelines-1), so I think it would be nice if we keep consistency 🙂

ilkka:
> With a GitHub release it's very easy to make changes if we've missed something out in hte yaml (and we have inadvertently removed the entire tekton-release bucket before, thus invalidating some download links)

Btw right now the `kubectl apply` in the README doesn't work because the yaml filename in the 0.2.1 GitHub release is not `release.yaml`.

a-roberts:
> > With a GitHub release it's very easy to make changes if we've missed something out in the yaml (and we have inadvertently removed the entire tekton-release bucket before, thus invalidating some download links)
> 
> Btw right now the `kubectl apply` in the README doesn't work because the yaml filename in the 0.2.1 GitHub release is not `release.yaml`.

Thanks, https://github.com/tektoncd/dashboard/pull/700 to fix for the time being

The `latest` folder was actually out of date in the Tekton bucket and we've been publishing using test folders, so I've just updated the bucket to contain the newest release files.

@mnuttall shall we go ahead and make this change? I hadn't realised we had direct edit access to the bucket in the browser (legit thought it was going to be updated only on running the release pipelines which can take a while 😅 ).

mnuttall:
Hi @devholic thank you for this and I'm sorry for the long delay in getting back to you. Rather than fix this once we'd like a more permanent solution, so I'm closing this PR in favour of https://github.com/tektoncd/dashboard/issues/921.

Issue #693, opened 2019-11-04
Dashboard breaks when trying to access pipeline/task runs


alan-ghelardi:
# Expected Behavior
The dashboard shows the pipeline/task run in question when I click over it. 
# Actual Behavior
When I run a pipeline via Triggers or creating it directly via the Dashboard and click over it or over one of its task runs, an empty page is displayed. By looking at the Javascript console, I saw the following error:
```shell
Error: "Minified React error #185; visit https://reactjs.org/docs/error-decoder.html?invariant=185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings. "
        Ts http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:78414
        enqueueSetState http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:46460
        setState http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:69:1429
        _ http://192.168.99.100/main.fb83ad3156121feb3461.js:1:371339
        value http://192.168.99.100/main.fb83ad3156121feb3461.js:1:371723
        value http://192.168.99.100/main.fb83ad3156121feb3461.js:1:371556
        Us http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:94965
        unstable_runWithPriority http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:9:3878
        lo http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:41497
        Us http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:90945
        As http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:79617
        ho http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:41768
        unstable_runWithPriority http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:9:3878
        lo http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:41497
        ho http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:41714
        po http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:41649
        Ts http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:78608
        enqueueSetState http://192.168.99.100/vendors.aca89a18cebd6eaf84b9.js:91:46460
[...]

PR #696, opened 2019-11-05
Only selectDefaultStep in componentDidMount

a-roberts:
Fix for https://github.com/tektoncd/dashboard/issues/693

Prevents infinite loop occurring, initially raised in Slack https://tektoncd.slack.com/archives/CHTHGRQBC/p1572780310027400

Tested with the pipeline provided (renders fine, we get no logs because the PipelineRun never really kicks off owing to the intentionally unsatisfied volume claim)

I also ran pipeline0 with the same code and observed the pipeline0-task kubectl apply was selected anyway, so I'm expecting this code here wasn't really necessary but definitely problematic for this scenario

### DIFF
--- a/packages/components/src/components/Task/Task.js
+++ b/packages/components/src/components/Task/Task.js
@@ -28,10 +28,6 @@ class Task extends Component {
     this.selectDefaultStep();
   }
 
-  componentDidUpdate() {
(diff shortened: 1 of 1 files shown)


a-roberts:
@AlanGreene could I get a review for this one please?

Issue #611, opened 2019-10-02
Secret creation errors should show in the modal


a-roberts:
# Expected Behavior
Display the notification on the modal so users can easily change values and try again

# Actual Behavior
The modal closes - it's often inconvenient to reinput data though, so why not display it on the form and suggest what went wrong (e.g. a failure creating it, so I might go on to choose a different namespace) - what's the standard practice?

# Steps to Reproduce the Problem

1. Create a secret  that's going to error
2. Observe where the error is displayed and that the modal's now vanished, prompting reinput of data

kimholmes:
@a-roberts There is a pattern for this in Carbon. Here is a screenshot. I can send you the link in Slack.
![form_notification](https://user-images.githubusercontent.com/21316329/67132890-57379080-f1d8-11e9-8d18-fa8d70bdc258.png)

kimholmes:
FYI - @Megan-Wright @AlanGreene The design team can provide a mock-up of this if helpful.

kimholmes:
Please see #727 for the updated design for Secrets. Note that we removed the usage of modals. The screen should appear as a new page:
![Secrets_Create_New1](https://user-images.githubusercontent.com/21316329/68722377-43870c00-0583-11ea-8c24-f90d9242c334.png)

@AlanGreene I did see this when I tried to recreate the errors (I think your build):
![Secrets_Create_Modal](https://user-images.githubusercontent.com/21316329/68722488-a082c200-0583-11ea-8c6d-2597e3301871.png)

Is this a good example of the errors that could appear? If so, we can create an example in the new design that will address this.

a-roberts:
@kimholmes Looks good, thanks for sharing - the error that appears as a notification appears when the request itself fails (for example when there's a secret that already exists with the same name in the name space).

We plan to get *this* in with tests before the above larger change, also if modals are being removed for secrets (going a bit meta) I assume the same principle applies for creating PipelineRuns manually and the access token create page for the extension.

kimholmes:
@AlanGreene Ok. that makes sense now. And yes, it would be the same for the other scenarios as well. Thanks for the heads up.

kimholmes:
@AlanGreene @Megan-Wright 
Here is the guideline (per Carbon).
1. The user has accidentally selected a namespace that already has a Secret by the same name. The user clicks "Create."
[...]

PR #719, opened 2019-11-12
Secret error notification displays in modal

Megan-Wright:
# Changes

For: https://github.com/tektoncd/dashboard/issues/611

The error notification for creating a secret shows within the secrets modal
The modal stays open, allowing users to edit the secret they want to create
The modal closes on successfully creation and displays a success message

# Submitter Checklist

review them:

for more details._


AlanGreene:
@Megan-Wright @a-roberts what's the issue that requires the introduction of enzyme? We should be able to set props on the component without enzyme, we already do this in many of our tests.

a-roberts:
> @Megan-Wright @a-roberts what's the issue that requires the introduction of enzyme? We should be able to set props on the component without enzyme, we already do this in many of our tests.

That'd be excellent, well we've got a service accounts dropdown and we want to set the text - we had a look at https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/Dropdown/Dropdown-test.js#L147 - so perhaps we don't need to use Enzyme after all...

AlanGreene:
@Megan-Wright I've opened a PR against your fork with a fix for the failing test and description of what I changed. Let me know if you have any questions.

https://github.com/Megan-Wright/dashboard/pull/1

src/reducers/secrets.js
```diff
@@ -77,11 +77,11 @@ function isFetching(state = false, action) {
   }
 }
 
-function errorMessage(state = null, action) {
+function errorMessage(state = false, action) {
   switch (action.type) {
     case 'SECRETS_FETCH_FAILURE':
     case 'SECRET_CREATE_FAILURE':
-      return action.error;
+      return 'The server reported a conflict';
```
AlanGreene: This doesn't seem right, in the case of `FETCH_FAILURE` why would this be caused by a conflict? Could the create also fail for other reasons?

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -370,68 +371,83 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
...
+              invalidFields={invalidFields}
+              onChange={this.handleAnnotationChange}
+              onRemove={this.handleRemove}
+              onAdd={this.handleAdd}
+            />
+            {errorMessage &&
+              (open && (
+                <InlineNotification
+                  kind="error"
+                  title="Error:"
```
AlanGreene: need to use `react-intl` for user-facing strings
that includes any headings, labels, tooltips, placeholders, etc.

some more on the Modal above

src/containers/Secrets/Secrets.js
```diff
@@ -86,32 +95,32 @@ export /* istanbul ignore next */ class Secrets extends Component {
     this.handleDeleteSecretToggle();
   };
 
+  handleCreateSecretClick(result) {
+    if (result) {
```
AlanGreene: This can be simplified:

```js
handleCreateSecretClick(openNewSecret) {
  this.setState({
    openNewSecret
  });
}
```

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -186,10 +187,10 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
       annotationsObject[key] = value;
     }
     postData.metadata.annotations = annotationsObject;
-
     if (Object.keys(invalidFields).length === 0) {
-      this.props.createSecret(postData, namespace);
-      this.props.handleNew();
+      await this.props.createSecret(postData, namespace).then(result => {
```
AlanGreene: Should we also be waiting for the `handleCreateSecret` here? Combining async/await and regular promise usage in the same statement can be confusing. Sticking to a single style makes it clearer what the intention is, and can help to avoid subtle errors.

package-lock.json
```diff
@@ -5054,7 +5054,8 @@
       "dependencies": {
         "@formatjs/intl-relativetimeformat": {
           "version": "4.1.1",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/@formatjs/intl-relativetimeformat/-/intl-relativetimeformat-4.1.1.tgz",
```
carlos-logro: this shouldn't be included.

src/actions/secrets.js
```diff
@@ -166,13 +166,18 @@ export function createSecret(postData, namespace) {
         namespace: serviceAccount.metadata.namespace,
         secretName: postData.metadata.name
       });
+      dispatch({
+        type: 'CLEAR_SECRET_ERROR_NOTIFICATION'
+      });
       dispatch({
         type: 'SECRET_CREATE_SUCCESS'
       });
+      return false;
```
carlos-logro: wouldn't it be better to use `SECRET_CREATE_SUCCESS` and `SECRET_CREATE_FAILURE` to initialize the result variable and get the value from the reducer?

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -370,68 +371,83 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
...
-            onAdd={this.handleAdd}
-          />
-          <div
-            ref={el => {
-              this.annotationsEnd = el;
-            }}
-          />
-        </form>
-      </Modal>
+      <>
```
carlos-logro: are the `<> </>` needed?

src/reducers/secrets.js
Megan-Wright: Yep.... think I changed it during testing just to make sure the right error was being got from the reducer - changed it back now

src/containers/SecretsModal/SecretsModal.js
Megan-Wright: ok, updated this now :)

src/containers/Secrets/Secrets.js
Megan-Wright: Done 😸

src/containers/SecretsModal/SecretsModal.js
Megan-Wright: Yep good point - updated to await `handleCreateSecret` as well

package-lock.json
Megan-Wright: ok, removed :)

src/containers/SecretsModal/SecretsModal.js
Megan-Wright: Nope, removed :)

src/actions/secrets.js
Megan-Wright: Do you mean that in the reducer `errorMessage` should return false instead of the error? i'm a little confused as to how this would work as then the reducer wouldn't be returning the error?

src/containers/SecretsModal/SecretsModal.js
AlanGreene: Not quite what I meant, you still have async/await and promise.then mixed here. Rewriting to use just async/await you'd have:

```js
const result = await this.props.createSecret(postData, namespace);
await this.props.handleCreateSecret(result);
```

package-lock.json
AlanGreene: This is caused by devs using different versions of npm, either way is fine for future reference. That said, we should probably all update to an agreed minimum version to avoid this unnecessary noise in future changes.

src/containers/SecretsModal/SecretsModal.js
Megan-Wright: ah ok, got it :)

Megan-Wright:
@AlanGreene Carlos has said he's happy with this now - does everything look good to you? :)

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -425,6 +434,28 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
...
+            (open && (
+              <InlineNotification
+                kind="error"
+                title={intl.formatMessage({
+                  id: 'dashboard.secretsModal.error.title',
+                  defaultMessage: 'Error:'
+                })}
+                subtitle={intl.formatMessage({
+                  id: 'dashboard.secretsModal.annotations.error.subtitle',
+                  defaultMessage: getErrorMessage(errorMessage)
```
AlanGreene: If this is an error coming back from the API it shouldn't be processed using react-intl as we have no way of knowing the value in advance.

`subtitle={getErrorMessage(errorMessage)}` is fine here

Megan-Wright:
@AlanGreene thanks :) done both of those changes now

Issue #713, opened 2019-11-08
Create secret > Screen will not scroll when adding more than two server URLs. Seems like nothing happens


rnapoles-tktn:
Problem:
When clicking the "plus" icon to add a third (or more) Server URL the UI will not scroll to display the added pair, looks like nothing happens, you can click many times and the added rows are not displayed. If the user tries to save the secret the UI displays "Server URL required" in red, however the UI does not display the empty field or does not provide a hint that the user has to manually scroll to display the URL fileds with missing values.

Steps to reproduce:
Create a secret
Add one URL
Click the + icon to add a second URL, the screen scrolls to display the second row
Click the + icon to add a third URL, the screen does NOT scroll to display the third row
Clcik the + icon a few more times, seems like nothing happens
Clcik the Submit button.
"Server URL required" is displayed in red, but it is not obvious what is missing.

PR #728, opened 2019-11-13
Secrets modal scrolls to end when server urls are added

Megan-Wright:
# Changes

For: https://github.com/tektoncd/dashboard/issues/713

Secrets modal now scrolls to the end when a server URL is added so that it is clear this has happened successfully without the user needing to scroll.
# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/SecretsModal/SecretsModal.js
+++ b/src/containers/SecretsModal/SecretsModal.js
@@ -47,6 +47,29 @@ function validateInputs(value, id) {
   return true;
 }
 
+// Used to determine if an annotation has been added or removed and therefore if the modal should scroll down
+function handleScroll(state, prevState) {
+  let shouldScroll = true;
+  if (state.name !== prevState.name) {
+    shouldScroll = false;
+  } else if (state.namespace !== prevState.namespace) {
+    shouldScroll = false;
+  } else if (state.username !== prevState.username) {
+    shouldScroll = false;
+  } else if (state.password !== prevState.password) {
+    shouldScroll = false;
+  } else if (state.serviceAccount !== prevState.serviceAccount) {
+    shouldScroll = false;
+  } else if (state.serviceAccounts !== prevState.serviceAccounts) {
+    shouldScroll = false;
+  } else if (state.accessTo !== prevState.accessTo) {
+    shouldScroll = false;
+  } else if (state.annotations.label !== prevState.annotations.label) {
+    shouldScroll = false;
+  }
+  return shouldScroll;
+}
+
 export /* istanbul ignore next */ class SecretsModal extends Component {
   constructor(props) {
     super(props);
@@ -72,13 +95,20 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
     };
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps, prevState) {
     const { webSocketConnected } = this.props;
     const { webSocketConnected: prevWebSocketConnected } = prevProps;
     if (webSocketConnected && prevWebSocketConnected === false) {
       this.props.fetchServiceAccounts();
     }
+    if (handleScroll(this.state, prevState)) {
+      this.scrollToBottom();
+    }
   }
+
+  scrollToBottom = () => {
+    this.annotationsEnd.scrollIntoView({ behavior: 'auto' });
+  };
 
   handleSubmit = () => {
     const invalidFields = [];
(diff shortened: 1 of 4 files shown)


src/containers/SecretsModal/SecretsModal.js
```diff
@@ -78,8 +78,13 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
     if (webSocketConnected && prevWebSocketConnected === false) {
       this.props.fetchServiceAccounts();
     }
+    this.scrollToBottom();
```
AlanGreene: Wouldn't we only want to do this if a new annotation was added? It could be quite disruptive if it randomly scrolled down on some unrelated update.

Megan-Wright: yeah I realised this after I put in the PR - working on it only doing it when an annotation is added now.

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -47,6 +47,8 @@ function validateInputs(value, id) {
   return true;
 }
 
+window.HTMLElement.prototype.scrollIntoView = function() {};
```
a-roberts: What does this do? A comment above would be useful

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -72,14 +74,43 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
...
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps, prevState) {
     const { webSocketConnected } = this.props;
     const { webSocketConnected: prevWebSocketConnected } = prevProps;
     if (webSocketConnected && prevWebSocketConnected === false) {
       this.props.fetchServiceAccounts();
     }
+    if (this.state.name !== prevState.name) {
```
a-roberts: It'd be useful to separate into the one function (e.g. shouldScroll that checks all of these things, passing in prevState and state), so it's tidier

Add a comment for said method please to explain why we need to do it

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -383,6 +414,11 @@ export /* istanbul ignore next */ class SecretsModal extends Component {
             handleAdd={this.handleAdd}
             invalidFields={invalidFields}
           />
+          <div
+            ref={el => {
```
a-roberts: What's el?

Megan-Wright: As far as I understand this line creates a reference to an element, so el would be the element created at the bottom of the modal

src/containers/SecretsModal/SecretsModal.js
Megan-Wright: done 😸

src/containers/SecretsModal/SecretsModal.js
Megan-Wright: done 😸

src/containers/SecretsModal/SecretsModal.js
AlanGreene: We shouldn't be reaching out to native DOM functionality like this, especially not overriding prototypes from the application code.

If we absolutely need this (I'm guessing jsdom doesn't implement scrollIntoView) we should do it from the test code instead.

src/containers/SecretsModal/SecretsModal.js
```diff
@@ -47,6 +47,32 @@ function validateInputs(value, id) {
   return true;
 }
 
+// Used to determine if an annotation has been added or removed and therefore if the modal should scroll down
+function handleScroll(state, prevState) {
+  let shouldScroll = true;
```
AlanGreene: do we need all of these checks or can we simplify and only check the annotations?

Megan-Wright: Unfortunately this is the only way Adam and I found of checking it.. I realise it's inefficient but there is no way of checking the annotations against the previous state as they get updated much before this point is reached.

src/containers/SecretsModal/SecretsModal.js
Megan-Wright: ok, I've moved this to the test code

AlanGreene: still needs to be removed from here

a-roberts:
@AlanGreene I reckon this is good to go in, as Megan mentions for checking new annotations we tried checking the current annotation state to the previous while in componentDidUpdate, but as we had already updated at this point our scrollTo only took us 90% of the page (there's no didUpdate and finished hook) and hence this approach

@Megan-Wright lemme know if this is accurate

Issue #675, opened 2019-10-28
UI Design Epic and PipelineRun Interactive Diagram Implementation


kimholmes:
# Expected Behavior
- A user of the Tekton Dashboard needs to find what's failed in his PipelineRun at a glance. 
- A user of the Tekton Dashboard needs to drill into a failed step and quickly access the log file that corresponds to the failed attempted.
![PLRdiagram1sm](https://user-images.githubusercontent.com/21316329/67706212-c0649400-f98e-11e9-88cc-e304fb4caa5c.png)

# Actual Behavior
![Tek_Diagram_existing](https://user-images.githubusercontent.com/21316329/67693629-622cb680-f978-11e9-9e0d-a28beeeaf847.png)

# Additional Info
See the comments for high fidelity designs for the Tekton dashboard PipelienRun Diagram, based on the Carbon design system. It includes a map of the intended behaviors and HTML/CSS specs.

We can use the map of the behaviors to iterate on as FED works through the code.

kimholmes:
View the high fidelity mockups at the following git links. Click on object to view associated specs:
### **PipelineRun Diagram View "Closed"**
https://kimholmes.github.io/Exported-Specs/PipelineTopology_Closed/
![PLRdiagram1](https://user-images.githubusercontent.com/21316329/67696702-b5553800-f97d-11e9-866b-97655e5f2ba9.png)

### **PipelineRun Diagram View "Open" - Node Selected**
https://kimholmes.github.io/Exported-Specs/PipelineTopology_DetailsLogs/
![PLRdiagram2](https://user-images.githubusercontent.com/21316329/67697429-e5e9a180-f97e-11e9-9052-8592f5ed8653.png)

### **PipelineRun Diagram View Logs "View more"**
https://kimholmes.github.io/Exported-Specs/PipelineTopology_LogsModal/
![PLRdiagram3](https://user-images.githubusercontent.com/21316329/67697735-66100700-f97f-11e9-9b5f-eee00de3942e.png)

kimholmes:
### **Hover states:**
![PLR_DiagramHover](https://user-images.githubusercontent.com/21316329/67704251-218a6880-f98b-11e9-8fb2-5872dcf2f9f6.png)

### **Node selection:**
![PLR_DiagramNodeSelection](https://user-images.githubusercontent.com/21316329/67704505-9e1d4700-f98b-11e9-91fe-09dc95282773.png)

### **Accordion selection:**
[...]

PR #732, opened 2019-11-13
First pass PipelineGraph UI component

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/675

Add PipelineGraph UI component using elkjs and vx.
It does not currently support user interaction (zoom, pan, click, etc.)
but this will come in future PRs.

It also does not currently render steps when the tasks are expanded (see screenshot).

Add storybook stories to cover the two main components:
- Node: represents a node in the graph (Task or Step)
- PipelineGraph: collection of Nodes + Links

image

# Submitter Checklist

review them:

for more details._


packages/components/src/components/Graph/Node.js
```diff
@@ -0,0 +1,94 @@
...
+import './Graph.scss';
+
+export default class Node extends PureComponent {
+  render() {
+    const { children, height, status, type, width, label = type } = this.props;
+
+    let StatusIcon;
+    if (type === 'Task' || status === 'error') {
+      switch (status) {
+        case 'error':
```
steveodonovan: This reads a little strange, make sense to move state === error to its own conditional above ?

AlanGreene: I had that originally but didn't like the duplication and potential for inconsistency in the icons used, we can have a Task with status 'error' or a Step with status 'error'. Tasks always show an icon, Steps only show icon on error.

Open to any suggestions to clean this up.

Issue #736, opened 2019-11-13
Tekton dashboard is refreshing about every minute especially on VPN


mtamboli:
# Expected Behavior
Dashboard should not get refreshed. Talked to @a-roberts on this. 

# Actual Behavior
Dashboard gets refreshed and you loose the work in progress. For example, if you are creating a webhook, you will have to restart unless you finish the task before the refresh.

# Steps to Reproduce the Problem

1. Keep the Tekton dashboard open on Chrome or safari
2. You will see refresh happening after a minute or so
3. This happened consistently when connected on VPN

# Additional Info

![image](https://user-images.githubusercontent.com/29490950/68796300-9e227580-0620-11ea-87b7-6ccdd80256ac.png)

AlanGreene:
From the screenshot it looks like your websocket connection is getting disconnected (request to ws://<host>/v1/websockets/resources). When this happens and the connection is re-established, the dashboard will make requests in the background to fetch the latest state of resources that may have changed during the disconnect. This may cause some parts of the UI to update to show the new information but should not cause a full page refresh.

@mtamboli @a-roberts can you provide more details on the behaviour experienced? Is this only affecting the webhooks extension pages or is this behaviour also present on core dashboard pages?

mtamboli:
I noticed this when creating webhook but pretty sure it is happening on other pages also like looking at PipelineRuns.

AlanGreene:
I managed to reproduce some issues that sound similar to what you've described on the PipelineRuns page against a local environment by disabling and re-enabling the port-forward to the dashboard service.

PR open to address those: https://github.com/tektoncd/dashboard/pull/737
Specifically this should prevent the Create PipelineRun dialog from being reset / closed on reconnect, and prevent the PipelineRuns list from momentarily reverting to a blank / loading state on reconnect.

@a-roberts I'm not seeing anything obvious in the webhook extensions code similar to the reset in PipelineRuns.

a-roberts:
@AlanGreene many thanks for the dashboard fix which is now merged, I'm not seeing anything obvious either on the extension side. On the webhook page we do a call to fetch webhooks on componentDidMount. I'm going to open a followup issue to investigate things on the webhooks extension side specifically and close this.

PR #737, opened 2019-11-13
Prevent PipelineRuns page from resetting on websocket reconnect

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/736

In `componentDidUpdate` we were calling `reset` on websocket
reconnect. This resulted in the component's state being reset,
closing the 'Create PipelineRun' dialog if open, among other effects.

Update the logic so we no longer call reset in this case, and
also prevent displaying the loading state unless there are no
existing PipelineRuns displayed. Also update the TaskRuns
loading state to match this.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/PipelineRuns/PipelineRuns.js
+++ b/src/containers/PipelineRuns/PipelineRuns.js
@@ -75,12 +75,10 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
       webSocketConnected: prevWebSocketConnected
     } = prevProps;
 
-    if (
-      namespace !== prevNamespace ||
-      !isEqual(filters, prevFilters) ||
-      (webSocketConnected && prevWebSocketConnected === false)
-    ) {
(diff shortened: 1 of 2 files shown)


a-roberts:
@AlanGreene looks good, should we do the same for TaskRuns as well here?

I see the same code at https://github.com/tektoncd/dashboard/blob/master/src/containers/TaskRuns/TaskRuns.js#L58 that would need hardening

As mentioned in your comment on the issue, I'm not seeing any reliance on data from the websocket for our webhooks extension code so I think your change + the same change for TaskRuns would be welcome

AlanGreene:
@a-roberts thanks for the review. That particular instance is safe as it's just doing a fetch, no reset. I will however update the TaskRuns loading state with the same logic I added to PipelineRuns to check for existing content first.

AlanGreene:
@a-roberts ready for another look

PR #749, opened 2019-11-15
Update Utils Test Coverage

carlos-logro:
# Changes
Added some missing test cases in `dashboard/src/utils/index` for coverage purposes. Hopefully this update will bump the global coverage above 90%.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/utils/index.test.js
+++ b/src/utils/index.test.js
@@ -11,7 +11,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { isStale, sortRunsByStartTime, typeToPlural } from '.';
+import * as API from '../api';
+
+import {
+  fetchLogs,
+  getGitValues,
+  isStale,
+  sortRunsByStartTime,
+  typeToPlural
+} from '.';
 
 it('sortRunsByStartTime', () => {
   const a = { name: 'a', status: { startTime: '0' } };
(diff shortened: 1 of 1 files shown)


src/utils/index.test.js
```diff
@@ -57,3 +65,36 @@ it('isStale', () => {
   expect(isStale(incomingResource, state)).toBe(true);
   expect(isStale(existingResource, state)).toBe(false);
 });
+
+it('fetchLogs', () => {
```
AlanGreene: This test is essentially testing itself, it essentially goes like this:
1. define mock function
1. spyOn API
1. call mock function
1. expect mock function to have been called

Instead we really want to test the actual `fetchLogs` function by calling it (not a mocked version of it). Something like this:

```js
it('fetchLogs', () => {
  const stepName = 'kubectl-apply';
  const stepStatus = { container: 'step-kubectl-apply' };
  const taskRun = { pod: 'pipeline-run-123456', namespace: 'default' };

  const logs = 'fake logs';
  jest.spyOn(API, 'getPodLog').mockImplementation(() => logs);

  const returnedLogs = fetchLogs(stepName, stepStatus, taskRun);
  expect(API.getPodLog).toHaveBeenCalledWith(expect.objectContaining({
    container: stepStatus.container,
    name: taskRun.pod,
    namespace: taskRun.namespace
  }));
});
```

src/utils/index.test.js
```diff
@@ -57,3 +65,36 @@ it('isStale', () => {
...
+    })
+  );
+  expect(mock).toHaveBeenCalledTimes(1);
+  expect(mock).toHaveBeenCalledWith(stepName, stepStatus, taskRun);
+});
+
+it('getGitValues', () => {
+  const url = 'https://github.com/user/repo';
+
+  const mock = jest.fn(getGitValues);
```
AlanGreene: similar issue here, we don't want to mock `getGitValues`

Issue #688, opened 2019-11-01
Update Tekton Dashboard to work with 0.8 Tekton Pipeline


a-roberts:
We've blown up before (see: fatally erroring when loading the PipelineRun page) when `conditionals` were introduced.

@dibbles is working on Triggers for our 0.3.0 release and we'd like to make that our 0.8.0 Pipelines release too. 

Check we don't break with any new Pipeline features 😄 

See https://github.com/tektoncd/dashboard/pull/649 for background.

See https://github.com/tektoncd/pipeline/releases/tag/v0.8.0 for changes.

Here's what I noticed and must be checked

- [ ] resourceSpec (having this will enable the other Dashboard PR for import resources and PipelineRun to go in)
- [ ] pipelineSpec
- [ ] Pipeline metrics (check we don't blowup, todo feature to display them eventually)

a-roberts:
I'll take a look at this one, first step will be to create an advanced enough Pipeline/Task to try and display the output of.

a-roberts:
It does blow up

![image](https://user-images.githubusercontent.com/12814972/68140969-99750780-ff24-11e9-92af-fb60995f573c.png)

My Tekton resources are https://github.com/a-roberts/pipeline-08-example.git

So whoever picks this up, import resources from my repository using the Tekton Dashboard, isolate what fails and look for `pipelineName` references

For convenience I did a simple `npm run start` to get full debug output instead of just using the proxy and the code in question (there may be other problems too) is at

```
index.js?bd11:179 Uncaught TypeError: Expected "pipelineName" to be defined
    at eval (index.js?bd11:179)
    at generatePath (react-router.js?4f2e:261)
    at eval (router.js?7dfc:167)
    at eval (PipelineRuns.js?bc95:110)
    at Array.map (<anonymous>)
    at PipelineRuns (PipelineRuns.js?bc95:86)
    at PipelineRuns (react-hot-loader.development.js?c2cb:821)
    at renderWithHooks (react-dom.development.js?61bb:15108)
    at mountIndeterminateComponent (react-dom.development.js?61bb:17342)
    at beginWork$1 (react-dom.development.js?61bb:18486)
```



@AlanGreene FYI, I'm going to work on other issues as we wouldn't want this fixed anyway before we can get Duane's code in for Triggers which keeps us at 0.7 until we change that.

mnuttall:
[...]

PR #764, opened 2019-11-20
Bypass createPipelineRunsByPipelineURL method when Pipeline is Embedded

carlos-logro:
as per issue: #688 

# Changes

With the new feature to embed pipeline inside the pipelineRuns yaml, the `PipelineRun` component was erroring out because it couldn't reference the "pipeline" used when trying to create the url that when clicked, applies the `tekton.dev/pipeline:{PIPELINE_NAME}` filter. 
Screen Shot 2019-11-20 at 3 08 52 PM

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRuns/PipelineRuns.js
+++ b/packages/components/src/components/PipelineRuns/PipelineRuns.js
@@ -110,14 +110,18 @@ const PipelineRuns = ({
               {url ? <Link to={url}>{pipelineRunName}</Link> : pipelineRunName}
             </StructuredListCell>
             <StructuredListCell>
-              <Link
-                to={createPipelineRunsByPipelineURL({
-                  namespace,
-                  pipelineName: pipelineRefName
-                })}
-              >
-                {pipelineRefName}
-              </Link>
+              {pipelineRefName ? (
+                <Link
+                  to={createPipelineRunsByPipelineURL({
+                    namespace,
+                    pipelineName: pipelineRefName
+                  })}
+                >
+                  {pipelineRefName}
+                </Link>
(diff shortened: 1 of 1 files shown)


packages/components/src/components/PipelineRuns/PipelineRuns.js
```diff
@@ -127,10 +127,13 @@ const PipelineRuns = ({
             {!pipelineName && (
               <StructuredListCell>
                 <Link
-                  to={createPipelineRunsByPipelineURL({
```
AlanGreene: See the cell above for how we handle this for the pipelineRun link: https://github.com/tektoncd/dashboard/pull/764/files#diff-2989be414084b3e10b098bac6149761fR125

We should do similar here, output an empty string instead of the `Link` component when `pipelineRefName` isn't available. This avoids rendering unnecessary DOM elements, and also avoid an accessibility issue where we'd have an invalid link with no URL and no text.

AlanGreene: See the cell above for how we handle this for the pipelineRun link: https://github.com/tektoncd/dashboard/pull/764/files#diff-2989be414084b3e10b098bac6149761fR125

We should do similar here, output an empty string instead of the `Link` component when `pipelineRefName` isn't available. This avoids rendering unnecessary DOM elements, and also avoid an accessibility issue where we'd have an invalid link with no URL and no text.

carlos-logro: PR updated! Thanks for the feedback @AlanGreene

Issue #759, opened 2019-11-20
TaskRun page changes to PipelineRun page when clicking all namespaces


CarolynMabbott:
# Expected Behavior
I change namespace on the left and it stays on the page TaskRun when i do the steps below

# Actual Behavior
When i change the namespace on the left to All Namespaces it goes to PipelineRun page

# Steps to Reproduce the Problem

1. On TaskRuns click on the task (eg, pipeline0-task)
2. Change namespace on the left to All Namespaces
3.

# Additional Info
Happens when you do specific steps above

AlanGreene:
fyi: https://github.com/tektoncd/dashboard/issues/569#issuecomment-551144646
In the updated designs the namespace dropdown will not be displayed on the detail pages

CarolynMabbott:
It's not on a detail page, when you click on the task on the TaskRun page it sets the filter on the page to be the task you clicked on and changes the namespace on the left to that specific namespace. When you change the namespace on the left back to All Namespaces is when the change to PipelineRuns page happens

AlanGreene:
by 'detail page' I mean any page displaying information on a single resource (not a list)

AlanGreene:
Fixed by https://github.com/tektoncd/dashboard/pull/768

Issue #758, opened 2019-11-20
PipelineResources page changes to PipelineRun page when clicking all namespaces 


CarolynMabbott:
# Expected Behavior
I change namespace on the left and it stays on the page PipelineResources when i do the steps below

# Actual Behavior
When i change the namespace on the left to All Namespaces I expect it to stay on the pipeline resources page 

# Steps to Reproduce the Problem

1. Click on a pipelinerun 
2. Go to PipelineResources
3.  Change namespace on left to All Namespaces

# Additional Info
Seems to be only happening when do it in that specific order

AlanGreene:
fyi: https://github.com/tektoncd/dashboard/issues/569#issuecomment-551144646
In the updated designs the namespace dropdown will not be displayed on the detail pages

AlanGreene:
Fixed by https://github.com/tektoncd/dashboard/pull/768

PR #768, opened 2019-11-21
Namespaces changed to all Namespaces doesn't change page

CarolynMabbott:
References issue #758 
References issue #759 

# Changes

Fixes bug where pipelineresource,  taskrun and pipelines pages get changed to pipelineruns page on change to all namespaces
Added tests to make sure it does as expected on TaskRun, PipelineResources and pipelines pages 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/SideNav/SideNav.js
+++ b/src/containers/SideNav/SideNav.js
@@ -73,6 +73,19 @@ export class SideNav extends Component {
 
     if (namespace === ALL_NAMESPACES) {
       this.props.selectNamespace(namespace);
+      const currentURL = this.props.match.url;
+      if (currentURL.includes(urls.taskRuns.all())) {
+        history.push(urls.taskRuns.all());
+        return;
+      }
+      if (currentURL.includes(urls.pipelineResources.all())) {
+        history.push(urls.pipelineResources.all());
+        return;
+      }
+      if (currentURL.includes(urls.pipelines.all())) {
+        history.push(urls.pipelines.all());
+        return;
+      }
       history.push('/');
       return;
     }
--- a/src/containers/SideNav/SideNav.test.js
+++ b/src/containers/SideNav/SideNav.test.js
@@ -16,7 +16,7 @@ import { Provider } from 'react-redux';
 import configureStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 import { fireEvent } from 'react-testing-library';
-import { ALL_NAMESPACES, paths } from '@tektoncd/dashboard-utils';
+import { ALL_NAMESPACES, paths, urls } from '@tektoncd/dashboard-utils';
 
 import { renderWithRouter } from '../../utils/test';
 import SideNavContainer, { SideNav } from './SideNav';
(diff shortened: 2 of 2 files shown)


src/containers/SideNav/SideNav.js
```diff
@@ -73,8 +73,20 @@ export class SideNav extends Component {
 
     if (namespace === ALL_NAMESPACES) {
       this.props.selectNamespace(namespace);
-      history.push('/');
-      return;
+      const currentURL = this.props.match.url;
```
AlanGreene: Ideally this code wouldn't be aware of the URL structure, instead it should use the [router utils](https://github.com/tektoncd/dashboard/blob/master/packages/utils/src/utils/router.js)

- instead of checking if the URL includes '/namespaces/' we should check for `match.params.namespace`
- instead of checking if the URL includes '/taskruns' we should check for `urls.taskruns.all()`
- etc.

This allows us to change the URL structure in a central location without things breaking (apart from the odd test that still needs to be updated to use the utils).

We should also ensure any other affected pages are updated to ensure consistency across the dashboard experience. Is there any way we can do this without needing to explicitly add handling for each affected page?

CarolynMabbott: The `namespaces` and `taskruns` was jto ensure it was on the taskruns page as the url will always have those components in to avoid it being confused with another page 

The only pages I noticed the specific issue on was where the url changed to include the `namespace/<namespace>/`. This does appear to happen for PipelineResources and TaskRuns, not sure how we could do this for every page without having to explicity handle each page 

The filters get removed when you change namespace in the dashboard currently. Would be good to keep the filters on when you switch namespace if we assume that a user may have the same task but in multiple namespaces

a-roberts:
Any update on this one? Sounds like a bug we want fixing for 0.3.0

src/containers/SideNav/SideNav.test.js
```diff
@@ -154,6 +154,108 @@ it('SideNav redirects to root when all namespaces selected on namespaced URL', a
...
+      />
+    </Provider>
+  );
+  fireEvent.click(getByText(namespace));
+  fireEvent.click(getByText(/all namespaces/i));
+  expect(selectNamespace).toHaveBeenCalledWith(ALL_NAMESPACES);
+  expect(push).toHaveBeenCalledWith('/taskruns');
+});
+
+it('SideNav redirects to TaskRuns page when all namespaces selected on namespaced URL on taskruns', async () => {
```
AlanGreene:
```suggestion
it('SideNav redirects to TaskRuns page when all namespaces selected on namespaced URL on PipelineResources', async () => {
```

src/containers/SideNav/SideNav.test.js
```diff
@@ -142,7 +142,7 @@ it('SideNav redirects to root when all namespaces selected on namespaced URL', a
       <SideNav
         extensions={[]}
         history={{ push }}
-        match={{ params: { namespace } }}
+        match={{ params: { namespace }, url: '/secrets' }}
```
AlanGreene: use `urls` util instead of hardcoding URLs, including in tests
this makes it far easier to maintain and avoids unnecessary code change if we modify the URL structures

there are a few more below that need updating too

PR #770, opened 2019-11-25
Support yaml format in req/resp

MartinForReal:
Signed-off-by: Fan Shang Xiang <fanshangxiang@gmail.com>

# Changes

support yaml format in req/resp

# Submitter Checklist

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -74,8 +74,8 @@ func Register(resource endpoints.Resource) *restful.Container {
 
 func registerKubeAPIProxy(r endpoints.Resource, container *restful.Container) {
 	proxy := new(restful.WebService)
-	proxy.Consumes(restful.MIME_JSON, "text/plain", "application/json-patch+json").
-		Produces(restful.MIME_JSON, "text/plain", "application/json-patch+json").
+	proxy.Consumes(restful.MIME_JSON, "text/plain", "application/json-patch+json", "application/yaml").
+		Produces(restful.MIME_JSON, "text/plain", "application/json-patch+json", "application/yaml").
 		Path("/proxy")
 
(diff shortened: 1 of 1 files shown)


AlanGreene:
@MartinForReal thanks for the PR. Is there an issue describing this change somewhere? Can you provide some detail / background here to help us understand the need for this change?

MartinForReal:
@googlebot I signed it!

MartinForReal:
@AlanGreene I found that the Kubernetes API server has support for YAML format. And we need to query manifest file and convert it to YAML.  So why don't we add support for producing YAML format instead?

AlanGreene:
The dashboard processes all data in JSON format (or JSON patch for some operations on secrets/service accounts) , can you provide some examples of where you think it would benefit from requesting data in YAML format instead?

> we need to query manifest file and convert it to YAML

Can you elaborate on this? Are there specific resources and uses you have in mind?

MartinForReal:
such as /namespaces/{namespace}/pipelines/{pipelinename}

AlanGreene:
Are you referring to the pages in the dashboard that display the Pipeline and Task definitions?
For example:
image

If so then I don't think that requesting this data in YAML format is necessary. Converting from JSON to YAML is a simple and inexpensive operation, and it ensures that we only have a single source of truth for our data in the application.

We use data from these definitions in a number of places in the dashboard, so having a single source of truth (the JSON data we retrieve and store) helps to ensure all elements of the UI related to that resource are consistent. We're also beginning to integrate this type of YAML display in other contexts so it will be even more important that the content is consistent with other data displayed on the page for that same resource.

AlanGreene:
Closing this now and we can revisit if we find a use case for requesting YAML directly.

Issue #772, opened 2019-11-25
Triggers resource tables have different style than other Dashboard tables


ncskier:
# Expected Behavior
I expect everything in the Dashboard to have a consistent style. In particular, I expect the tables used in the Triggers views to match the visual style of the tables in the other Dashboard views.

# Actual Behavior
The Triggers tables have a style inconsistent with the rest of the Dashboard.

Here is a comparison showing the inconsistencies between the TriggerBindings table and the PipelineRuns table:
<img width="1274" alt="Screen Shot 2019-11-25 at 9 54 38 AM" src="https://user-images.githubusercontent.com/16961496/69553410-2c2a2480-0f6e-11ea-8e5e-6138aa527cfa.png">
<img width="1274" alt="Screen Shot 2019-11-25 at 9 54 50 AM" src="https://user-images.githubusercontent.com/16961496/69553420-30564200-0f6e-11ea-920e-e086ae512519.png">

AlanGreene:
All list views should be updated to use Carbon DataTable for consistency and to make use of the action toolbar and other DataTable features. This will eliminate most of the style inconsistencies between the list views as we'll no longer be using the StructuredList styles.

We have a common [Table component](https://github.com/tektoncd/dashboard/blob/master/packages/components/src/components/Table/Table.js) now (thanks @carlos-logro) that should make this easier.

AlanGreene:
Also see https://github.com/tektoncd/dashboard/issues/775 for updated designs of the Pipeline(Run)s and Task(Run)s pages.

carlos-logro:
@Megan-Wright during the surfacing of the TriggerBindings details page I stumbled upon some nits that should be taken care of in this issue:

1) Table title has too many padding
<img width="712" alt="Screen Shot 2019-12-02 at 1 45 48 PM" src="https://user-images.githubusercontent.com/49996607/69987077-b04f4f80-150c-11ea-8f7a-71eefb83e125.png">

2) even when no batchActions or toolbar buttons are passed as props, the space for the element still shows up. 
<img width="736" alt="Screen Shot 2019-12-02 at 1 45 35 PM" src="https://user-images.githubusercontent.com/49996607/69987100-b8a78a80-150c-11ea-915c-927d99771962.png">

# How it should look:
[...]

Issue #670, opened 2019-10-25
TriggerBinding details paneling first pass


ncskier:
This work is under epic https://github.com/tektoncd/dashboard/issues/509

# Expected Behavior
I expect to be able to view the details of a [TriggerBinding](https://github.com/tektoncd/triggers/blob/master/docs/triggerbindings.md) resource in the Dashboard.

This panel should have two [tabs](https://www.carbondesignsystem.com/components/tabs/code/), `Details` and `YAML`.
### YAML view
The `YAML` view should display the YAML definition of the TriggerBinding resource (`kubectl get triggerbinding TB_NAME -oyaml`). The YAML should be formatted like a code snippet.

Here is the latest wireframe from #509 for the `YAML` view.
![tb-detail-yaml](https://user-images.githubusercontent.com/16961496/67604199-f6b0d200-f748-11e9-8159-f55e67650772.png)

### Details view
The `Details` view should be a nicely formatted representation of the important information in the TriggerBinding's YAML. The `Details` view must include the following information:
- Labels
- list of params each with the following:
  - name
  - value

Here is the latest wireframe from #509 for the `Details` view.
![tb-detail](https://user-images.githubusercontent.com/16961496/67604219-04feee00-f749-11e9-9a3c-38015991c662.png)

# Actual Behavior
We have no paneling for viewing TriggerBinding resource details.

# Additional Info
- TriggerBindings README: [triggerbindings.md](https://github.com/tektoncd/triggers/blob/master/docs/triggerbindings.md)
- TriggerBindings type definition: [trigger_binding_types.yaml](https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/trigger_binding_types.go)
- TriggerBinding example: [triggerbinding.yaml](https://github.com/tektoncd/triggers/blob/master/examples/triggerbindings/triggerbinding.yaml)

mii-w:
![Screen Shot 2019-11-05 at 10 02 09 AM](https://user-images.githubusercontent.com/51759906/68222479-10f47680-ffb9-11e9-8f71-0bdd869a4384.png)
![Screen Shot 2019-11-05 at 10 44 08 AM](https://user-images.githubusercontent.com/51759906/68222613-4305d880-ffb9-11e9-8e30-0eb8b2b8cad1.png)

mii-w:
[...]

PR #797, opened 2019-12-02
TriggerBindings Details Page

carlos-logro:
issue: #670 

## Details with Labels 
Screen Shot 2019-12-03 at 1 38 09 PM

## Details with no Labels
Screen Shot 2019-12-02 at 3 43 45 PM

## Yaml Tab
Screen Shot 2019-12-02 at 3 43 12 PM

**Added some changes to Table Component that should be included in the PR for issue #772. These necessary changes have been commented in the issue.**

# Submitter Checklist

review them:

for more details._


packages/utils/src/utils/router.js
```diff
@@ -125,6 +125,12 @@ export const paths = {
   triggerBindings: {
     all() {
       return '/triggerbindings';
+    },
+    byNamespace() {
+      return byNamespace({ path: '/triggerbindings' });
+    },
+    byName() {
+      return byNamespace({ path: '/triggerbindings/:triggerbindingName' });
```
AlanGreene:
```suggestion
      return byNamespace({ path: '/triggerbindings/:triggerBindingName' });
```

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,321 @@
...
+import { fetchTriggerBinding } from '../../actions/triggerBindings';
+
+const initialState = {
+  showYAML: false,
+  showDetails: true
+};
+
+export /* istanbul ignore next */ class TriggerBindingContainer extends Component {
+  /* eslint-disable react/sort-comp */
+  constructor(props) {
```
AlanGreene:
```diff
-constructor(props) {
-  super(props);
-  this.state = initialState;
-}
+state = initialState;
```

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,321 @@
...
+        name,
+        value
+      })
+    );
+
+    return (
+      <div id="triggerBinding">
+        <h1>{triggerbindingName}</h1>
+        <Tabs
+          ariaLabel="listbox"
```
AlanGreene: Carbon already adds the relevant ARIA attributes and roles to the resulting elements, no need for `ariaLabel`, `role`, etc. here.

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,321 @@
...
+          iconDescription="show menu options"
+          role="navigation"
+          tabContentClassName="tab-content"
+        >
+          <Tab
+            label={intl.formatMessage({
+              id: 'dashboard.triggerBinding.details',
+              defaultMessage: 'Details'
+            })}
+            onClick={() => {
```
AlanGreene: Carbon already handles keyboard accessibility and switching the tab content, so we should be able to remove the `onClick`, `selected`, and `tabIndex` here.

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,321 @@
...
+                  defaultMessage: 'Labels:'
+                })}
+              </span>
+              {formattedLabelsToRender.length === 0
+                ? intl.formatMessage({
+                    id: 'dashboard.triggerBinding.noLabels',
+                    defaultMessage: 'None'
+                  })
+                : formattedLabelsToRender.map(label => (
+                    <Tag type="blue">{label}</Tag>
```
AlanGreene: missing `key` prop

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,321 @@
...
+              title={intl.formatMessage({
+                id: 'dashboard.triggerBinding.title',
+                defaultMessage: 'Parameters'
+              })}
+              headers={headersForParameters}
+              rows={rowsForParameters}
+              selectedNamespace={selectedNamespace}
+              emptyTextAllNamespaces={intl.formatMessage({
+                id: 'dashboard.triggerBinding.noParams',
+                defaultMessage: 'No parameters found for this binding.'
```
AlanGreene: TriggerBinding

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,321 @@
...
+            </p>
+            <Table
+              title={intl.formatMessage({
+                id: 'dashboard.triggerBinding.title',
+                defaultMessage: 'Parameters'
+              })}
+              headers={headersForParameters}
+              rows={rowsForParameters}
+              selectedNamespace={selectedNamespace}
+              emptyTextAllNamespaces={intl.formatMessage({
```
AlanGreene: To ensure consistency between the 2 messages here we should store the result of `intl.formatMessage` in a variable and use it for both props.

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,321 @@
...
+    const rowsForParameters = triggerBinding.spec.params.map(
+      ({ name, value }) => ({
+        id: name,
+        name,
+        value
+      })
+    );
+
+    return (
+      <div id="triggerBinding">
```
AlanGreene: avoid using ids unless necessary, a class would be fine here and lends itself better to reuse

src/containers/TriggerBinding/TriggerBinding.scss
```diff
@@ -0,0 +1,43 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+@import '~carbon-components/scss/globals/scss/vars';
+
+#triggerBinding {
```
AlanGreene: avoid as much page-specific styling as possible, this should be consistent with the other trigger details pages

src/nls/messages_en.json
```diff
@@ -66,6 +66,17 @@
...
+    "dashboard.triggerBinding.details": "Details",
+    "dashboard.triggerBinding.errorloading": "Error loading TriggerBinding",
+    "dashboard.triggerBinding.labels": "Labels:",
+    "dashboard.triggerBinding.namespace": "Namespace:",
+    "dashboard.triggerBinding.noLabels": "None",
+    "dashboard.triggerBinding.noParams": "No parameters found for this binding.",
+    "dashboard.triggerBinding.title": "Parameters",
+    "dashboard.triggerBinding.unavailable": "TriggerBinding not available",
+    "dashboard.triggerBinding.value": "Value",
+    "dashboard.triggerBinding.yaml": "Yaml",
```
AlanGreene: no need to translate `YAML`, and it should be capitalised in use as it's an acronym

src/nls/messages_en.json
```diff
@@ -66,6 +66,17 @@
     "dashboard.taskRun.status.running": "Running",
     "dashboard.taskRun.status.succeeded": "Completed",
     "dashboard.taskRun.status.waiting": "Waiting",
+    "dashboard.triggerBinding.dateCreated": "Date Created:",
```
AlanGreene: We've recently removed colons from labels in other parts of the UI, should we do the same here? Might be OK since this is presented differently

src/nls/messages_en.json
```diff
@@ -66,6 +66,17 @@
...
+    "dashboard.triggerBinding.details": "Details",
+    "dashboard.triggerBinding.errorloading": "Error loading TriggerBinding",
+    "dashboard.triggerBinding.labels": "Labels:",
+    "dashboard.triggerBinding.namespace": "Namespace:",
+    "dashboard.triggerBinding.noLabels": "None",
+    "dashboard.triggerBinding.noParams": "No parameters found for this binding.",
+    "dashboard.triggerBinding.title": "Parameters",
```
AlanGreene: id is a bit ambiguous / potentially misleading, maybe something more specific would help?

```suggestion
    "dashboard.triggerBinding.parametersTitle": "Parameters",
```

src/nls/messages_en.json
carlos-logro: Only reason I add the colons in this one is because values are appearing to the right of the label.

carlos-logro:
@AlanGreene thank you for the feedback! Every comment has been addressed and PR updated.

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,308 @@
...
+    }
+  }
+
+  fetchData() {
+    const { match } = this.props;
+    const { namespace, triggerBindingName } = match.params;
+    this.props.fetchTriggerBinding({ name: triggerBindingName, namespace });
+  }
+
+  handleTabSelect = index => {
```
AlanGreene: Carbon handles this for us. The content to be displayed should be a child of the corresponding Tab component. This is important for accessibility, ensuring the correct labels and navigation are used.

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,308 @@
...
+    if (error) {
+      return TriggerBindingContainer.notification({
+        kind: 'error',
+        intl
+      });
+    }
+    if (!triggerBinding) {
+      return TriggerBindingContainer.notification({
+        kind: 'error',
+        message: 'TriggerBinding not available',
```
AlanGreene: need to use intl for this message
it's also the same text as the info title, is this intentional?

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,308 @@
...
+        intl
+      });
+    }
+
+    let labels = '';
+    const formattedLabelsToRender = [];
+    if (triggerBinding.metadata.labels) {
```
AlanGreene: Is this likely to be reused for other components? Might be a good idea to extract it to a utility function.

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,308 @@
...
+      labels = JSON.stringify(triggerBinding.metadata.labels)
+        .replace('{', '')
+        .replace('}', '')
+        .replace(/['"]+/g, '')
+        .replace('$', '');
+
+      const labelsSplitOnComma = labels.split(',');
+      labelsSplitOnComma.forEach(foo => {
+        const splitOnColon = foo.split(':');
+        formattedLabelsToRender.push(`  ${splitOnColon[0]}: ${splitOnColon[1]}
```
AlanGreene: why do we add leading whitespace?
can we rename `foo` to something more meaningful?

also, for readability you can do:

```js
const [key, value] = label.split(':');
...push(`${key}: ${value}`);
```

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,308 @@
...
+            })}
+          />
+          <Tab label="YAML" />
+        </Tabs>
+        {showDetails && (
+          <div className="details">
+            <p>
+              <span>
+                {intl.formatMessage({
+                  id: 'dashboard.triggerBinding.dateCreated',
```
AlanGreene: a lot of these strings are generic, not specific to TriggerBindings, e.g. 'Date Created', 'Details', 'Labels:'
we should reuse where possible instead of duplicating, this will keep our string count down when it comes to translation and help with consistency

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,308 @@
...
+              isSortable
+            />
+          </div>
+        )}
+        {showYAML && (
+          <div className="bx--snippet--multi">
+            <code>
+              <pre>{jsYaml.dump(triggerBinding)}</pre>
+            </code>
+          </div>
```
AlanGreene: This is going to be duplicated on most resources, it would be good to have this as a simple component that just accepts the resource as a prop

src/containers/TriggerBinding/TriggerBinding.scss
```diff
@@ -0,0 +1,38 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+@import '~carbon-components/scss/globals/scss/vars';
+
+.triggerBinding {
+
+  h1 {
```
AlanGreene: will any of this be specific to TriggerBindings or should it be common across all pages?

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,308 @@
...
+              </span>
+              {formattedLabelsToRender.length === 0
+                ? intl.formatMessage({
+                    id: 'dashboard.triggerBinding.noLabels',
+                    defaultMessage: 'None'
+                  })
+                : formattedLabelsToRender.map(label => (
+                    <Tag
+                      type="blue"
+                      key={`${triggerBindingName}:label:${label}`}
```
AlanGreene: `label` would be fine as long as it's unique
the `key` only needs to be unique for this specific use of `map`, not globally

src/containers/TriggerBinding/TriggerBinding.js
carlos-logro: ok so ill make the content look like Adam has in his Trigger details page PR

src/containers/TriggerBinding/TriggerBinding.js
carlos-logro: Yeah looking at this now its kinda funky.

src/containers/TriggerBinding/TriggerBinding.js
carlos-logro: if you are referring to the extraction of the labels then yes. This is also being used in other Trigger Details.

src/containers/TriggerBinding/TriggerBinding.js
carlos-logro: what keyword should I use instead of triggerBinding? metadata?

src/containers/TriggerBinding/TriggerBinding.js
carlos-logro: agree. Include that in this PR or separate issue?

src/containers/TriggerBinding/TriggerBinding.scss
carlos-logro: Not sure, but I believe so.

src/containers/TriggerBinding/TriggerBinding.js
AlanGreene: for fields that are actually from the resource metadata that works 👍 

for other generic things like 'Details' maybe we go with something like 'dashboard.resource.detailsTab'?

src/containers/TriggerBinding/TriggerBinding.js
AlanGreene: separate is fine, should go in before we end up duplicating this too much though

carlos-logro:
@AlanGreene changes have been made. Before this is merged PRs #809 & #810 should go in first.

carlos-logro:
@AlanGreene PR has been rebased to use `ViewYAML` component & `formatLabels` util.

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,251 @@
...
+        value
+      })
+    );
+
+    const emptyTextMessage = intl.formatMessage({
+      id: 'dashboard.triggerBinding.noParams',
+      defaultMessage: 'No parameters found for this TriggerBinding.'
+    });
+
+    console.log(triggerBinding.metadata.labels);
```
a-roberts: 👀

src/containers/TriggerBinding/TriggerBinding.scss
AlanGreene: if this is common we should move it somewhere else for reuse, and remove/rename references to `triggerBinding`

src/containers/TriggerBinding/TriggerBinding.js
carlos-logro: seriouscatcover

carlos-logro: updated*

src/containers/TriggerBinding/TriggerBinding.js
```diff
@@ -0,0 +1,249 @@
...
+                title={intl.formatMessage({
+                  id: 'dashboard.triggerBinding.parametersTableTitle',
+                  defaultMessage: 'Parameters'
+                })}
+                headers={headersForParameters}
+                rows={rowsForParameters}
+                selectedNamespace={selectedNamespace}
+                emptyTextAllNamespaces={emptyTextMessage}
+                emptyTextSelectedNamespace={emptyTextMessage}
+                isSortable
```
AlanGreene: let's leave out the sorting for consistency with other pages until we can address https://github.com/tektoncd/dashboard/issues/842

src/containers/TriggerBinding/TriggerBinding.scss
AlanGreene: Created an issue to follow up on this: https://github.com/tektoncd/dashboard/issues/844

carlos-logro:
@AlanGreene update done! Spoke with @a-roberts to see if he was also using those styles and he said yes so I moved the styles to a newly created scss folder given that I didn't know where else to add it.

Issue #796, opened 2019-12-02
Assess compatibility with Tekton 0.9 (Dashboard and Webhooks Extension)


a-roberts:
With 0.9.0 imminent we should check what works and what doesn't

Update: it's out, see https://github.com/tektoncd/pipeline/releases/tag/v0.9.0

a-roberts:
I'm going to have a look at this while my two big PRs (EventListener and TriggerTemplate details) are on hold

a-roberts:
https://storage.cloud.google.com/tekton-release-staging/pipeline/previous/v0.9.0-rc1/release.yaml @afrittoli has kindly staged this yaml, just installed it and all pods are running, so will start the testing

a-roberts:
Ok, import resources definitely doesn't work which we know about, https://github.com/tektoncd/dashboard/issues/774 to fix

Everything else looks fine (rendering simple pipelines, logs, details, status), PipelineResources, PipelineRuns, secrets. Todo have a look at the release notes once Andrea's ready and assess (we shouldn't be unable to render any pages now as we have @AlanGreene's ErrorBoundary changes in).

PR #804, opened 2019-12-04
Change 'serviceAccount' to 'serviceAccountName'

### LINKED ISSUE #774: Use serviceAccountName not serviceAccount for Tekton 0.9
We use it here for example https://github.com/tektoncd/dashboard/blob/58e63b990a48db808fa925ddf968416c524e6509/src/api/index.js#L200 - should be serviceAccountName

Under this item check for any other references relying on the field being called `serviceAccount` for TaskRuns and PipelineRuns too.

If you find a way to have this work with both 0.8 and 0.9 that'd be great 😄

ncskier:
# Changes
Fixes #774
Pipelines v0.9.0 works with only 'serviceAccountName'
Pipelines v0.8.0 works with both 'serviceAccount' and 'serviceAccountName'

/cc @AlanGreene 
/cc @a-roberts 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,13 @@
 
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/kubernetes/dashboard/blob/master/LICENSE)
 
-Tekton Dashboard is a general purpose, web-based UI for Tekton Pipelines. It allows users to manage and view Tekton PipelineRuns and TaskRuns and the resources involved in their creation, execution, and completion. It also allows filtering of PipelineRuns and TaskRuns by label. 
+Tekton Dashboard is a general purpose, web-based UI for Tekton Pipelines. It allows users to manage and view Tekton PipelineRuns and TaskRuns and the resources involved in their creation, execution, and completion. It also allows filtering of PipelineRuns and TaskRuns by label.
 
 ![Dashboard UI workloads page](docs/dashboard-ui.png)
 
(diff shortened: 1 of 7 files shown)


ncskier:
Since this is the Dashboard README and the monitor task is a separate installation from the dashboard project, do you think it's alright that I include v0.9.0 as well?

@a-roberts I've updated the README, so let me know what you think 🙂

a-roberts:
> Since this is the Dashboard README and the monitor task is a separate installation from the dashboard project, do you think it's alright that I include v0.9.0 as well?
> 
> @a-roberts I've updated the README, so let me know what you think 🙂

Yep, although my testing for 0.9 with Triggers lead to the same serviceAccount vs serviceAccountName problem, see https://github.com/tektoncd/dashboard/issues/796, and I'm sure the monitor task would fail for other reasons

While yeah we could have the dashboard only work with 0.9 OK, having the extension lag behind isn't great.

@mnuttall what do you think?

Also because this would be a breaking change for 0.7 users, I made a 0.8 branch here (technically it's 0.8 and 0.7 - so we at least know when things are good before moving up with this change and any others).

mnuttall:
Readme changes look good. If @dibbles wants to pull this into the 0.3 release I'm ok with that. 

> having the extension lag behind isn't great.

It's not great but it's not enough reason on its own to hold this change out.

Issue #817, opened 2019-12-10
DataTables isSortable function doesn't work for 'computed' values 


Megan-Wright:
# Expected Behaviour 
We should be able to have tables that can be sorted by various values. For example: PipelineRuns should be able to be sorted by name, namespace, pipeline, last transition time, etc.

# Actual Behaviour
When sorting is enabled for DataTables (which it isn't currently due to this problem), trying to sort by any column that has a 'computed' value - e.g. a link or a date/time doesn't work.

steveodonovan:
> When sorting is enabled for DataTables (which it isn't currently due to this problem)

FYI, paging is also an important part of why this wasn't enabled. The apiserver does not support sorting natively. I suggest having this piece worked before enabling sorting

PR #833, opened 2019-12-11
DataTable's Sort Function for Computed Values

carlos-logro:
issue: #817 

# Changes
By passing a custom sort function to the DataTable, the component is now able to sort strings, Links & Formatted Timestamps. 

# Submitter Checklist

review them:

for more details._

/cc @AlanGreene @a-roberts

### DIFF
--- a/packages/components/src/components/Table/Table.js
+++ b/packages/components/src/components/Table/Table.js
@@ -25,7 +25,7 @@ import {
   TableToolbarContent
 } from 'carbon-components-react';
 
-import { ALL_NAMESPACES } from '@tektoncd/dashboard-utils';
+import { ALL_NAMESPACES, rowSorter } from '@tektoncd/dashboard-utils';
 
 import './Table.scss';
 
@@ -60,6 +60,7 @@ const Table = props => {
         rows={dataRows}
         headers={dataHeaders}
         isSortable={isSortable}
+        sortRow={rowSorter}
         render={({
           rows,
           headers,
--- a/packages/utils/src/utils/index.js
+++ b/packages/utils/src/utils/index.js
@@ -158,3 +158,31 @@ export function formatLabels(labelsRaw) {
 
   return formattedLabelsToRender;
 }
+
+function comparer(a, b) {
+  if (a < b) {
+    return -1;
+  }
+  if (a > b) {
+    return 1;
+  }
+  return 0;
+}
+
+export function rowSorter(cellA, cellB, { sortDirection, sortStates }) {
+  let valueA = cellA;
+  let valueB = cellB;
+
+  if (typeof cellA === 'object') {
+    valueA = cellA.props.children || new Date(cellA.props.date);
+  }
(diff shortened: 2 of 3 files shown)


carlos-logro:
Unit test failing because global test coverage is below threshold.

steveodonovan:
@carlos-logro FYI https://github.com/tektoncd/dashboard/issues/817#issuecomment-565459177 See my comment here. Paging will be an important piece here.I looked into it way back, it was far from trivial.

carlos-logro:
> @carlos-logro FYI [#817 (comment)](https://github.com/tektoncd/dashboard/issues/817#issuecomment-565459177) See my comment here. Paging will be an important piece here.I looked into it way back, it was far from trivial.

what do you mean by paging?

steveodonovan:
> what do you mean by paging?

`Pagination` If you're not familiar its limiting the result set to whats called a page with arrows to navigate through the `pages`. It's a very common pattern and we're going to need it. Otherwise too many resources for any given page will kill the UI. Sorting is something which needs to happen at the back end otherwise you pages are never going to be sorted.

@AlanGreene @a-roberts One of you might be more familiar with plans in this area but backend sorting and proper pagination is whats needed here. We could have 1000's of resources being loaded in this UI easily. The data has to be managed properly.

a-roberts:
> > what do you mean by paging?
> 
> `Pagination` If you're not familiar its limiting the result set to whats called a page with arrows to navigate through the `pages`. It's a very common pattern and we're going to need it. Otherwise too many resources for any given page will kill the UI. Sorting is something which needs to happen at the back end otherwise you pages are never going to be sorted.
> 
> @AlanGreene @a-roberts One of you might be more familiar with plans in this area but backend sorting and proper pagination is whats needed here. We could have 1000's of resources being loaded in this UI easily. The data has to be managed properly.

So from our perspective in Hursley our priority is really around supporting our Triggers work with Kabanero and we haven't mentioned pagination in many months, I think the next juicy piece of UI work we're interested in is the showing of service accounts for secrets but there's a lot of other priorities too outside of UI work

Would it be reasonable to do this as a separate PR or do you think if one had, say, 1000 PipelineRuns in a namespace we'd simply be crippled on sort?

carlos-logro:
but wouldn't the pagination part be done in the reducer or at mapStateToProps? The sorting that I added in this PR just overrides the one the DataTable component had as default. And the table only sorts items that were passed as rows so I'm assuming all the pagination logic is done way before passing those row items.

steveodonovan:
lets discuss at the planning meeting later today

/hold

a-roberts:
We've decided it's best to have a proper look into pagination and sorting before enabling this, so let's go ahead and make a new issue and then we can get this PR in once we're happy (otherwise users may experience dreadful performance we think and especially to do with the CSS animations)

AlanGreene:
Closing this PR for now since there are other changes required before this can go in as discussed above.

Issue #829, opened 2019-12-11
Ensure dashboard correctly handles inline Pipeline specs


AlanGreene:
# Expected Behavior
The dashboard should be capable of displaying PipelineRuns containing inline Pipeline specs.
Any code / UI relating to PipelineRuns should not make assumptions about having a standalone Pipeline definition (i.e. `pipelineRef`).

# Actual Behavior
The PipelineRuns component makes assumptions about the presence of `pipelineRun.spec.pipelineRef.name` which breaks the UI in the case of an inline Pipeline spec.

There may be other instances...

# Additional Info
We had a similar issue with TaskRuns (including a recent regression https://github.com/tektoncd/dashboard/issues/821), so we should make sure this is addressed consistently across the dashboard application.

a-roberts:
Carlos' PR is merged for this so closing

PR #835, opened 2019-12-11
Handle inline Pipeline spec

carlos-logro:
issue: #829 
at the same time deals with #840

# Changes
#764 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRuns/PipelineRuns.js
+++ b/packages/components/src/components/PipelineRuns/PipelineRuns.js
@@ -97,51 +97,59 @@ const PipelineRuns = ({
     }
   });
 
-  const pipelineRunsFormatted = pipelineRuns.map(pipelineRun => ({
-    id: `${pipelineRun.metadata.namespace}:${pipelineRun.metadata.name}`,
-    name: (
-      <Link
-        to={createPipelineRunURL({
-          namespace: pipelineRun.metadata.namespace,
-          pipelineRunName: createPipelineRunDisplayName({
-            pipelineRunMetadata: pipelineRun.metadata
-          })
-        })}
-      >
-        {pipelineRun.metadata.name}
-      </Link>
-    ),
-    pipeline: !hidePipeline && (
-      <Link
-        to={createPipelineRunsByPipelineURL({
-          namespace: pipelineRun.metadata.namespace,
-          pipelineName: pipelineRun.spec.pipelineRef.name
-        })}
-      >
-        {pipelineRun.spec.pipelineRef.name}
-      </Link>
-    ),
-    namespace: !hideNamespace && pipelineRun.metadata.namespace,
-    status: (
-      <div className="definition">
-        <div
-          className="status"
-          data-status={getStatus(pipelineRun).status}
-          data-reason={getStatus(pipelineRun).reason}
-        >
-          <div className="status-icon">
-            {getPipelineRunStatusIcon(pipelineRun)}
+  const pipelineRunsFormatted = pipelineRuns.map(pipelineRun => {
+    const { namespace, annotations } = pipelineRun.metadata;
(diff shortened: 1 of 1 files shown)


packages/components/src/components/PipelineRuns/PipelineRuns.js
```diff
@@ -97,51 +97,59 @@ const PipelineRuns = ({
...
-          data-reason={getStatus(pipelineRun).reason}
-        >
-          <div className="status-icon">
-            {getPipelineRunStatusIcon(pipelineRun)}
+  const pipelineRunsFormatted = pipelineRuns.map(pipelineRun => {
+    const { namespace, annotations } = pipelineRun.metadata;
+    const pipelineRunName = createPipelineRunDisplayName({
+      pipelineRunMetadata: pipelineRun.metadata
+    });
+    const pipelineRefName = pipelineRun.spec.pipelineRef.name;
```
AlanGreene: do we still have an issue here?

carlos-logro: noup this line takes care of undefined values: https://github.com/tektoncd/dashboard/blob/0c73c4086cbf9f54ede001c4fa122007a0caac1e/packages/components/src/components/PipelineRuns/PipelineRuns.js#L120

AlanGreene: If `pipelineRun.spec.pipelineRef` is undefined wouldn't `pipelineRun.spec.pipelineRef.name` throw an error?

carlos-logro: double checked this and its just initializing `pipelineRefName` as undefined. no errors.

packages/components/src/components/PipelineRuns/PipelineRuns.js
```diff
@@ -97,51 +97,59 @@ const PipelineRuns = ({
...
+    const statusIcon = getPipelineRunStatusIcon(pipelineRun);
+    const pipelineRunStatus = getPipelineRunStatus(pipelineRun, intl);
+    const url = createPipelineRunURL({
+      namespace,
+      pipelineRunName,
+      annotations
+    });
+    return {
+      id: `${namespace}:${pipelineRunName}`,
+      name: <Link to={url}>{pipelineRunName}</Link>,
```
AlanGreene: Noticed another small issue here, we need to make sure we handle `createPipelineRunURL` not returning a value as we do for TaskRuns. Consumers may not have a location to link to in all cases so may want to display just the name without making it a link.

```suggestion
      name: url ? <Link to={url}>{pipelineRunName}</Link> : pipelineRunName,
```

packages/components/src/components/PipelineRuns/PipelineRuns.js
```diff
@@ -97,51 +97,59 @@ const PipelineRuns = ({
...
+            })}
+          >
+            {pipelineRefName}
+          </Link>
+        ) : (
+          ''
+        )),
+      namespace: !hideNamespace && namespace,
+      status: (
+        <div className="definition">
```
AlanGreene: Can you rebase and make sure this still works with long status messages and the change from https://github.com/tektoncd/dashboard/pull/841? The icon and text should appear on the same line, and long text should not wrap below the icon, e.g.
image
and
image

packages/components/src/components/PipelineRuns/PipelineRuns.js
AlanGreene: strange that `pipelineRef` is present but as long as it's not causing any issues 🤷‍♂ let's keep an eye on it just in case

carlos-logro:
@AlanGreene rebased and tested 👍 everything is outputting how its supposed to.

Issue #832, opened 2019-12-11
PipelineRuns and TaskRuns tables should include status information to help identify issues


AlanGreene:
# Expected Behavior
The PipelineRuns and TaskRuns lists should include information to help the user identify "interesting" runs, e.g. timed out / failed, at a glance and provide links to more relevant information for further debugging.

# Actual Behavior
We currently display the status message in the table. In case of failures, this can be quite a large chunk of text. Here's an example of the TaskRuns list:
![image](https://user-images.githubusercontent.com/2829095/70632005-fa7bb380-1c25-11ea-872e-0c7cea1a9497.png)

The large status content stretches the column of the table, causing other fields to wrap, reducing overall readability of the table. If we truncate the content and/or hide it in a tooltip users lose the ability to copy the content which can be quite useful.

# Additional Info
The latest designs for these views (https://github.com/tektoncd/dashboard/issues/775 and https://github.com/tektoncd/dashboard/issues/776) omit the status message and have only the icon in the table. The message is displayed in a tooltip.

![image](https://user-images.githubusercontent.com/2829095/70632885-880bd300-1c27-11ea-8149-d1d078bec0b4.png)

An alternative I've proposed is to include the status reason in the table as this would differentiate between different categories of failure, e.g. timeout, cancelled, missing resources, etc.. We could (should?) include the status message in the details page and if desired also as a tooltip on the table as in the design. Having the status message on the details page would be very useful as it can often be used in conjunction with the logs to more quickly hone in on the problem.

Example of the same TaskRuns as above, displaying the reason instead of the message:
![image](https://user-images.githubusercontent.com/2829095/70633111-e933a680-1c27-11ea-8129-ee03d539c730.png)

With this I can see more clearly that one of the "failures" was due to the TaskRun being manually cancelled, whereas the other is a genuine failure, and we make more real estate available for other columns which now no longer need to wrap.

Values that could be displayed for status:
- `Running`
- `Succeeded`
- `Failed` (unknown cause, or a step failed)
- other failure status values:
  - `ConditionCheckFailed`
  - `CouldntGetPipeline`
  - `CouldntGetTask`
  - `ExceededResourceQuota`
  - `InvalidPipelineResourceBindings`
  - `PipelineRunCancelled`
  - `PipelineRunTimeout`
  - `PipelineValidationFailed`
  - `TaskRunCancelled`
  - `TaskRunTimeout`
  - etc.

[...]

PR #836, opened 2019-12-12
Update PipelineRuns and TaskRuns status column

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/832

Only display status icon in table, include reason and longer
message in tooltip.

Also display the status message on the details page with
option to copy to clipboard.

image

# Submitter Checklist

review them:

for more details._


AlanGreene:
For now I've given 'cancelled' PipelineRuns and TaskRuns a different icon colour to the other 'failed' status icons. While this is technically OK from an a11y point of view (we provide differentiating info in the tooltip), it would be better to select another icon for these so we're not relying on colour to differentiate between them when scanning visually.

Issue #840, opened 2019-12-13
PipelineRuns dashboard view has some styling issues


steveodonovan:
# Description

Seeing the icon for a pipeline run status appear on top of the status. Also the header above the overflow menu looks off. @Megan-Wright @AlanGreene Possibly related to the move to data tables? either of you aware ? Also seeing in the storybook

![Screenshot 2019-12-13 at 12 27 54](https://user-images.githubusercontent.com/18128367/70800783-06da4a80-1da5-11ea-9694-60c83877cb4a.png)

Megan-Wright:
@steveodonovan Yep we were trying to get the move to data tables in ASAP due to it preventing other work from being done - so the styling is by no means perfect.. I'll have a go at fixing up the issues you've raised now 😸

carlos-logro:
although its not the issue being tackled, I noticed this issue and included the simple solution in the PR #833

Issue #837, opened 2019-12-12
Incomplete Table Division when column contains buttons


carlos-logro:
# Expected Behavior
Table division should go across the whole table.
<img width="1394" alt="Screen Shot 2019-12-12 at 2 43 02 PM" src="https://user-images.githubusercontent.com/49996607/70743431-b9e76d00-1ced-11ea-88f0-d29509f2c662.png">

# Actual Behavior
Division is partially across.
<img width="1386" alt="Screen Shot 2019-12-12 at 2 41 21 PM" src="https://user-images.githubusercontent.com/49996607/70743519-e9967500-1ced-11ea-9cf5-9f199d92905b.png">
<img width="1408" alt="Screen Shot 2019-12-12 at 2 41 28 PM" src="https://user-images.githubusercontent.com/49996607/70743521-e9967500-1ced-11ea-8d3e-bc64e473437c.png">

# Steps to Reproduce the Problem
Visit one of these pages:
- Pipelines
- PipelineRuns
- PipelineResources
- Tasks
- TaskRuns

# Additional Info

PR #841, opened 2019-12-13
Update CSS for data tables

Megan-Wright:
# Changes

For: https://github.com/tektoncd/dashboard/issues/837 and https://github.com/tektoncd/dashboard/issues/840

Fixes status icon being positioned above its status in PipelineRun and TaskRun tables, and extends the table header CSS to include the overflow column.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRuns/PipelineRuns.scss
+++ b/packages/components/src/components/PipelineRuns/PipelineRuns.scss
@@ -14,9 +14,12 @@ limitations under the License.
 @import '../../scss/vars';
 
 .definition .status {
+  display:flex;
+
   .status-icon {
     margin-right: 0.5em;
     vertical-align: sub;
+    flex-shrink: 0;
   }
 
   &:not([data-status]) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -5054,7 +5054,8 @@
       "dependencies": {
         "@formatjs/intl-relativetimeformat": {
           "version": "4.1.1",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/@formatjs/intl-relativetimeformat/-/intl-relativetimeformat-4.1.1.tgz",
+          "integrity": "sha512-k+WADErd1ORTcJ/4LyrVzJcXkcXL3l9wKCsMT2d7Kq4dbap2OKv2GLBGdKR3H7SBKCooyDRV5W79et9egGzsyQ==",
           "requires": {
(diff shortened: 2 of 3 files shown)


src/components/Definitions/Definitions.scss
```diff
@@ -40,6 +40,7 @@ main {
   .status-icon {
     margin-right: 0.5em;
     vertical-align: sub;
+    float: left;
```
steveodonovan: I think this change should be added to pipelineRuns.scss under the packages dir, this causes the story book and third party consumes to still have the issue 

storybook screen grab with the changeset

Screenshot 2019-12-13 at 14 58 26

Megan-Wright: ah ok, I'll update that now

packages/components/src/components/PipelineRuns/PipelineRuns.scss
```diff
@@ -17,6 +17,7 @@ limitations under the License.
   .status-icon {
     margin-right: 0.5em;
     vertical-align: sub;
+    float: left;
```
carlos-logro: I mentioned in issue 840 that this css is not needed if the the div wrapper is removed from the icon. This is covered in PR #835 https://github.com/tektoncd/dashboard/blob/0c73c4086cbf9f54ede001c4fa122007a0caac1e/packages/components/src/components/PipelineRuns/PipelineRuns.js#L136

AlanGreene: @carlos-logro see my review comment, using flexbox on the container gives us a cleaner layout. If we could combine the 2 approaches (removing the container, and applying flexbox on a higher level container) that would be great.

Megan-Wright: @AlanGreene I made the change you suggested - this fixes everything nicely for the PipelineRun view but messes up the TaskRun view to look like this:
Screenshot 2019-12-16 at 14 03 53
Not sure why this is happening.... suggestions?

AlanGreene: You'll need to stop the icon from shrinking when other content in the flex container tries to steal space:

```
.definition .status {
  display: flex;

  .status-icon {
    flex-shrink: 0;
  }
}
```

Megan-Wright: thanks, made that change now :)

Issue #787, opened 2019-11-28
Add ability to rebuild a PipelineRun from the PipelineRuns table


CarolynMabbott:
I want to be able to rerun a pipelinerun from the pipelineruns page in the 3 dot buttons
I dont want to have to click on the actual pipelinerun to press the rerun button

PR #858, opened 2020-01-02
Ability to rerun pipelineruns from PipelineRuns page

CarolynMabbott:
# Changes

References issue #787 

Can rerun any pipelineruns from the PipelineRuns page 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/PipelineRuns/PipelineRuns.js
+++ b/src/containers/PipelineRuns/PipelineRuns.js
@@ -41,7 +41,11 @@ import {
   isFetchingPipelineRuns,
   isWebSocketConnected
 } from '../../reducers';
-import { cancelPipelineRun, deletePipelineRun } from '../../api';
+import {
+  cancelPipelineRun,
+  deletePipelineRun,
+  rerunPipelineRun
+} from '../../api';
 
 const initialState = {
   showCreatePipelineRunModal: false,
@@ -87,6 +91,11 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
   deleteRun = pipelineRun => {
     const { name, namespace } = pipelineRun.metadata;
     deletePipelineRun({ name, namespace });
+  };
+
+  rerun = pipelineRun => {
+    const { name, namespace } = pipelineRun.metadata;
+    rerunPipelineRun(namespace, { pipelinerunname: name });
   };
 
   toggleModal = showCreatePipelineRunModal => {
@@ -192,6 +201,13 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
               { name: resource.metadata.name }
             )
         }
+      },
+      {
+        actionText: intl.formatMessage({
+          id: 'dashboard.rerunPipelineRun.actionText',
+          defaultMessage: 'Rerun'
+        }),
+        action: this.rerun
       }
     ];
   };
--- a/src/containers/PipelineRuns/PipelineRuns.test.js
+++ b/src/containers/PipelineRuns/PipelineRuns.test.js
@@ -17,6 +17,7 @@ import { Provider } from 'react-redux';
 import thunk from 'redux-thunk';
 import configureStore from 'redux-mock-store';
 import { Route } from 'react-router-dom';
+import { urls } from '@tektoncd/dashboard-utils';
 import { renderWithRouter } from '../../utils/test';
 import * as API from '../../api';
 import PipelineRunsContainer from './PipelineRuns';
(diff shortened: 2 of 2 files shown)


carlos-logro:
@CarolynMabbott i see you updated the actionButtons in the test case but there isn't actually any testing that checks if the right buttons with respective actions is actually being executed. Here is an [example](https://github.com/tektoncd/dashboard/blob/9e8e862b151eafacf92053bb621244b36eaee75b/packages/components/src/components/TaskTree/TaskTree.test.js#L37), basically you want to test that when the button is clicked, rebuilding method is called.

packages/components/src/components/PipelineRuns/PipelineRuns.test.js
```diff
@@ -72,7 +72,19 @@ it('PipelineRuns renders data', () => {
         {
           actionText: intl.formatMessage({
             id: 'test.actionText',
-            defaultMessage: 'TestAction'
+            defaultMessage: 'Stop'
```
AlanGreene: similar to @carlos-logro's comment
there's no need for the test to exactly mirror the actions provided in the UI
the important things would be to test:
- support for rendering multiple actions
- that when an action is clicked, the correct click handler is called

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -89,6 +93,11 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
     deletePipelineRun({ name, namespace });
   };
 
+  rerunRun = pipelineRun => {
```
AlanGreene: how about just 'rerun'?

packages/components/src/components/PipelineRuns/PipelineRuns.test.js
AlanGreene: let's revert this change, it's not adding anything to this test and the functionality is tested in the container below

src/containers/PipelineRuns/PipelineRuns.test.js
```diff
@@ -161,6 +161,7 @@ beforeEach(() => {
   jest.spyOn(API, 'getServiceAccounts').mockImplementation(() => []);
   jest.spyOn(API, 'getPipelines').mockImplementation(() => []);
   jest.spyOn(API, 'getPipelineResources').mockImplementation(() => []);
+  jest.spyOn(API, 'rerunPipelineRun').mockImplementation(() => []);
```
AlanGreene: can we spy on this in the test where it's actually used or is it needed across multiple tests?

src/containers/PipelineRuns/PipelineRuns.test.js
```diff
@@ -362,3 +363,44 @@ it('An invalid filter value is disallowed and reported', async () => {
...
+  const match = {
+    params: {},
+    url: '/pipelineruns'
+  };
+
+  const onSelect = jest.fn();
+  const { queryByText, getByTestId, getByText } = renderWithRouter(
+    <Provider store={mockTestStore}>
+      <Route
+        path="/pipelineruns"
```
AlanGreene: use router utils (e.g. `paths.pipelineRuns.all()` and `urls.pipelineRuns.all()`) instead of hardcoding, makes it easier to keep track of the values used and to update them in future

src/containers/PipelineRuns/PipelineRuns.test.js
```diff
@@ -362,3 +363,44 @@ it('An invalid filter value is disallowed and reported', async () => {
...
+            fetchPipelineRuns={() => Promise.resolve()}
+            pipelineRuns={pipelineRunsTestStore.pipelineRuns.byId}
+          />
+        )}
+      />
+    </Provider>,
+    { route: '/pipelineruns' }
+  );
+
+  const filterValue = 'baz:bam';
```
AlanGreene: no need to filter if all we're testing is the rerun function

src/containers/PipelineRuns/PipelineRuns.test.js
```diff
@@ -362,3 +363,44 @@ it('An invalid filter value is disallowed and reported', async () => {
...
 });
+
+it('TaskTree handles rerun event in PipelineRuns page', async () => {
+  const mockTestStore = mockStore(testStore);
+  const match = {
+    params: {},
+    url: '/pipelineruns'
+  };
+
+  const onSelect = jest.fn();
```
AlanGreene: this doesn't appear to be used, the only reference is `onSelect.mockClear()` at the end of the test

src/containers/PipelineRuns/PipelineRuns.test.js
```diff
@@ -362,3 +363,44 @@ it('An invalid filter value is disallowed and reported', async () => {
...
+    <Provider store={mockTestStore}>
+      <Route
+        path="/pipelineruns"
+        render={props => (
+          <PipelineRunsContainer
+            {...props}
+            match={match}
+            error={null}
+            loading={false}
+            namespace="namespace-1"
```
AlanGreene: the container should be able to get many of these props from the store so shouldn't need to pass them explicitly in the test

Issue #857, opened 2020-01-02
Full width of secrets table body doesn't render when there are no secrets created


Megan-Wright:
Expected Behaviour: The secrets table should be consistent with the rest of the tables when there are no secrets present.

Actual Behaviour:
When a user has no secrets created, the table doesn't render the annotations column within the table body.
![Screenshot 2020-01-02 at 14.21.45.png](https://images.zenhubusercontent.com/5d25b016e0bbdc62005f7da6/868ce897-b881-4c89-ac7b-e93d6a115081)

Steps to recreate: 
Navigate to the secrets panel when there are no secrets created.

PR #859, opened 2020-01-03
Fix rendering issue with empty secrets table

Megan-Wright:
# Changes

For: https://github.com/tektoncd/dashboard/issues/857

Increases number of columns in a table body by one to account for the 'select box' column used within the secrets table. This fixes the issue where the last column would not render in the last column of an empty secrets table. 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/Table/Table.js
+++ b/packages/components/src/components/Table/Table.js
@@ -135,7 +135,11 @@ const Table = props => {
                 <TableBody>
                   {!loading && dataRows.length === 0 && (
                     <TableRow>
-                      <TableCell colSpan={headers.length}>
+                      <TableCell
+                        colSpan={
+                          headers.length + (batchActionButtons.length ? 1 : 0)
+                        }
+                      >
                         <div className="noRows">
                           {selectedNamespace === ALL_NAMESPACES
                             ? emptyTextAllNamespaces


CarolynMabbott:
When you switch to a namespace like `docker` with no secrets it reverts back to the broken state, please fix this :)

Megan-Wright:
@CarolynMabbott oops, fixed now 😸

CarolynMabbott:
Checked it again and it seems fine. Doesnt break when changing namespaces or deleting secrets 

/lgtm 
/meow space

packages/components/src/components/Table/Table.js
```diff
@@ -135,7 +135,7 @@ const Table = props => {
                 <TableBody>
                   {!loading && dataRows.length === 0 && (
                     <TableRow>
-                      <TableCell colSpan={headers.length}>
+                      <TableCell colSpan={headers.length + 1}>
```
AlanGreene: should this be conditional on the presence of batch actions? i.e. column span would match headers length for a normal table, and add the +1 when batch actions are specified (which triggers the checkbox column)

Megan-Wright: Good point - updated that now :)

packages/components/src/components/Table/Table.js
```diff
@@ -135,13 +135,24 @@ const Table = props => {
...
+                      {batchActionButtons.length > 0 && (
+                        <TableCell colSpan={headers.length + 1}>
+                          <div className="noRows">
+                            {selectedNamespace === ALL_NAMESPACES
+                              ? emptyTextAllNamespaces
+                              : emptyTextSelectedNamespace}
+                          </div>
+                        </TableCell>
+                      )}
+                      {batchActionButtons.length === 0 && (
```
AlanGreene: no need to duplicate this, the only difference is the length so let's just modify that:

```js
<TableCell colSpan={headers.length + (batchActionButtons.length ? 1 : 0)}>
  <div className="noRows">
    {selectedNamespace === ALL_NAMESPACES
      ? emptyTextAllNamespaces
      : emptyTextSelectedNamespace}
  </div>
</TableCell>
```

Megan-Wright: ok, done that now :)

Issue #682, opened 2019-10-30
TestExtensionRegistration still (still) flaky (again (again (again)))


a-roberts:
See https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_dashboard/678/pull-tekton-dashboard-unit-tests/1189597155866710018/

There are lots of related issues and @carlos-logro added debug info so have a look and see if we can fix it last 😄 

```
I1030 17:41:33.698] {"level":"debug","msg":"Registering path: /v1/extensions/extension1/robots"}
I1030 17:41:33.698] {"level":"debug","msg":"Registering path: /v1/extensions/extension1/secrets"}
I1030 17:41:33.698] {"level":"debug","msg":"Extension Controller detected extension 'extension2' created"}
I1030 17:41:33.698] {"level":"info","msg":"Adding Extension extension2"}
I1030 17:41:33.698] {"level":"debug","msg":"Registering path: /v1/extensions/extension2/robots"}
I1030 17:41:33.699] {"level":"debug","msg":"Registering path: /v1/extensions/extension2/secrets"}
I1030 17:41:33.699] --- FAIL: TestExtensionRegistration (10.04s)
I1030 17:41:33.699]     routes_test.go:318: In TestExtensionRegistration
I1030 17:41:33.699]     routes_test.go:344: robots.secrets
I1030 17:41:33.699]     routes_test.go:461: waiting for registration
I1030 17:41:33.699]     routes_test.go:462: services:
I1030 17:41:33.699]     routes_test.go:463: start time: 2019-10-30 17:41:23.642775569 +0000 UTC m=+0.180869565
I1030 17:41:33.700]     routes_test.go:464: --- ---
I1030 17:41:33.700]     routes_test.go:478: counter: 1
I1030 17:41:33.700]     routes_test.go:479: time: 2019-10-30 17:41:23.64775723 +0000 UTC m=+0.185851255
I1030 17:41:33.700]     routes_test.go:483: extensionCreates:
I1030 17:41:33.700]     routes_test.go:484: 1
I1030 17:41:33.700]     routes_test.go:490: 
I1030 17:41:33.701]     routes_test.go:478: counter: 2
I1030 17:41:33.701]     routes_test.go:479: time: 2019-10-30 17:41:23.651183986 +0000 UTC m=+0.189278037
I1030 17:41:33.701]     routes_test.go:483: extensionCreates:
I1030 17:41:33.701]     routes_test.go:484: 2
[...]

PR #877, opened 2020-01-08
Fix extension test

vtereso:
This PR fixes https://github.com/tektoncd/dashboard/issues/682

# Changes

Fixed test:
  - Added docstring to test
  - Simplified test and better assertion using HTTP mock
  - Add router.Handler to avoid global state pollution within tests
  - Simplify extension proxy handling

Doc string cleanups:
  - Use gofmt styling for docstrings
  - Clarify some of the docstrings

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/controllers/kubernetes/extension.go
+++ b/pkg/controllers/kubernetes/extension.go
@@ -10,29 +10,37 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
-// package scoped variables to be utilized within EventHandler funcs
-var installNamespace string
+// extensionHandler is a wrapper around the router Handler so that it can be
+// used within informer handler functions
+type extensionHandler struct {
+	*router.Handler
+	installNamespace string
+}
 
-// Registers the K8s shared informer that reacts to extension service updates
-func NewExtensionController(sharedK8sInformerFactory k8sinformer.SharedInformerFactory, dashboardNamespace string) {
+// NewExtensionController registers the K8s shared informer that reacts to
+// extension service updates
+func NewExtensionController(sharedK8sInformerFactory k8sinformer.SharedInformerFactory, dashboardNamespace string, handler *router.Handler) {
 	logging.Log.Debug("In NewExtensionController")
-	installNamespace = dashboardNamespace
+	h := extensionHandler{
+		Handler:          handler,
+		installNamespace: dashboardNamespace,
+	}
 	extensionServiceInformer := sharedK8sInformerFactory.Core().V1().Services().Informer()
 	// ResourceEventHandler interface functions only pass object interfaces
 	// Inlined functions to keep mux in scope rather than reconciler
 	extensionServiceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    serviceCreated,
-		UpdateFunc: serviceUpdated,
-		DeleteFunc: serviceDeleted,
+		AddFunc:    h.serviceCreated,
+		UpdateFunc: h.serviceUpdated,
+		DeleteFunc: h.serviceDeleted,
 	})
 }
 
-func serviceCreated(obj interface{}) {
+func (e extensionHandler) serviceCreated(obj interface{}) {
 	service := obj.(*v1.Service)
-	if service.Namespace == installNamespace {
-		if value := service.Labels[router.ExtensionLabelKey]; value == router.ExtensionLabelValue {
+	if service.Namespace == e.installNamespace {
+		if value := service.Labels[router.ExtensionLabelKey]; value == router.ExtensionLabelValue && service.Spec.ClusterIP != "" {
 			logging.Log.Debugf("Extension Controller detected extension '%s' created", service.Name)
-			router.RegisterExtension(service)
+			e.RegisterExtension(service)
 			data := broadcaster.SocketData{
 				MessageType: broadcaster.ExtensionCreated,
 				Payload:     obj,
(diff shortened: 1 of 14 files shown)


vtereso:
I installed the latest webhooks extension release and the proxy functionality is still working as intended.

pkg/controllers/kubernetes/extension.go
```diff
@@ -42,40 +50,37 @@ func serviceCreated(obj interface{}) {
...
-			router.UnregisterExtension(oldService)
+			e.UnregisterExtension(oldService)
+			updates++
 		}
-		if value := newService.Labels[router.ExtensionLabelKey]; versionUpdated && value == router.ExtensionLabelValue {
+		if value := newService.Labels[router.ExtensionLabelKey]; versionUpdated && value == router.ExtensionLabelValue && newService.Spec.ClusterIP != "" {
 			logging.Log.Debugf("Extension Controller Update: Add new extension '%s'", newService.Name)
-			router.RegisterExtension(newService)
+			e.RegisterExtension(newService)
+			updates += 2
```
a-roberts: Why is this two? Based on the switch later are we basically using this number as a status code of sorts (if so I'd rather we didn't, use an enum like object so it's more readable)

pkg/router/router.go
```diff
@@ -26,50 +27,150 @@ import (
...
-	registerPropertiesEndpoint(resource, wsContainer)
-	registerWebsocket(resource, wsContainer)
-	registerHealthProbe(resource, wsContainer)
-	registerReadinessProbe(resource, wsContainer)
-	registerExtensions(wsContainer)
-	registerKubeAPIProxy(resource, wsContainer)
-	return wsContainer
+	h := &Handler{
+		Container:       restful.NewContainer(),
+		uidExtensionMap: make(map[string]*Extension),
```
a-roberts: We definitely need the *?

pkg/router/router.go
```diff
@@ -26,50 +27,150 @@ import (
...
+	registerHealthProbe(resource, h.Container)
+	registerReadinessProbe(resource, h.Container)
+	registerKubeAPIProxy(resource, h.Container)
+	h.registerExtensions()
+	return h
+}
+
+// Handler is an HTTP handler with internal configuration to avoid global state
+type Handler struct {
+	*restful.Container
```
a-roberts: Same question as above - do we definitely need all the explicit pointer refs?

pkg/router/router.go
```diff
@@ -164,152 +267,80 @@ func registerPropertiesEndpoint(r endpoints.Resource, container *restful.Contain
...
-	extensions := []Extension{}
-	extensionMutex.RLock()
-	for _, e := range uidExtensionMap {
-		extensions = append(extensions, e)
-	}
-	extensionMutex.RUnlock()
-	return extensions
+// MarshalJSON marshals the Extension into JSON.
+func (e Extension) MarshalJSON() ([]byte, error) {
+	type Alias Extension
```
a-roberts: What does this do and how come it's now needed?

pkg/router/router.go
```diff
@@ -164,152 +267,80 @@ func registerPropertiesEndpoint(r endpoints.Resource, container *restful.Contain
...
+	return json.Marshal(&struct {
+		URL string `json:"url"`
+		*Alias
+	}{
+		URL:   e.URL.String(),
+		Alias: (*Alias)(&e),
```
a-roberts: Same q as above (comments explaining this would be nice for maintenance)

pkg/router/router.go
```diff
@@ -164,152 +267,80 @@ func registerPropertiesEndpoint(r endpoints.Resource, container *restful.Contain
...
+// extensionPath constructs the extension path (excluding the root) used by
+// restful.Route
+func extensionPath(extName, path string) string {
+	return strings.TrimSuffix(fmt.Sprintf("%s/%s", extName, path), "/")
 }
 
-// Returns target port if exists, else source == target
-func getServicePort(svc corev1.Service) string {
+// getServicePort returns the target port if exists or the source port otherwise
+func getServicePort(svc *corev1.Service) string {
```
a-roberts: Why's this now a *?

pkg/router/routes_test.go
```diff
@@ -310,247 +317,165 @@ func makeFake(t *testing.T, r *endpoints.Resource, resourceType, namespace, reso
...
-						Port: extensionPorts[1],
-					},
-				},
+		},
+		{
+			name: "Three endpoint",
+			extensionEndpoints: []string{
+				"robots",
+				"cowboys",
+				"aliens",
```
a-roberts: 😄

pkg/router/routes_test.go
```diff
@@ -310,247 +317,165 @@ func makeFake(t *testing.T, r *endpoints.Resource, resourceType, namespace, reso
...
 
-	// Labels not supported on fake client, manual filter
-	serviceList, err := r.K8sClient.CoreV1().Services(installNamespace).List(metav1.ListOptions{})
-	if err != nil {
-		t.Fatalf("Error obtaining services from K8sClient")
-	}
+			// Assert routes
+			replyCode := http.StatusNoContent
+			extensionProxyRoot := fmt.Sprintf("%s%s/%s", server.URL, ExtensionRoot, extensionService.Name)
+			defer gock.Off()
```
a-roberts: What does this stuff do? `Assert routes` is pretty vague, we've got it now as an import but I shouldn't have to go look it up to understand this block and why it's now necessary

pkg/router/routes_test.go
```diff
@@ -310,247 +317,165 @@ func makeFake(t *testing.T, r *endpoints.Resource, resourceType, namespace, reso
...
+			select {
+			case event := <-subscriber.SubChan():
+				switch event.MessageType {
+				case broadcaster.ExtensionDeleted:
+					break
+				default:
+					t.Fatalf("Unexpected event message type: %s", event.MessageType)
+				}
+			case <-time.After(time.Second):
+				t.Fatalf("Timed out waiting for extension deletion")
```
a-roberts: Can drop the f

pkg/controllers/kubernetes/extension.go
vtereso: I am using those numbers like their bitwise representations so 1 = 01, 2 = 10. If it was to increment one it would be ambiguous which case was triggered in the situation where only a service is either now an extension or no longer an extension (the extension label is removed).

pkg/router/router.go
vtereso: We used to construct the url each time we handled proxy requests. Instead, now when extension registration happens, I construct the URL. However, url.URL is not marshalled as expected (see: https://play.golang.org/p/6ffOLym8hz2). This overrides the default MarshalJSON method so that url.URL marshals as expected with the old payload.

pkg/router/routes_test.go
vtereso: Before there was a dummy extension that the test would explicitly make requests to, which "mocked" an actual extension. Using gock, the requests actually go through the dashboard test server and the proxy requests are intercepted (the transport level is intercepted on clients, where both the dashboard and gock are using the default client), which is a more appropriate mock/behavior test.

vtereso: I'll update the comment

pkg/router/router.go
vtereso: All services are passed by the informer where they are guaranteed as being non-nil once passed to the router package. Since the structs are relatively big
```
type Service struct {
	metav1.TypeMeta `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
	Spec ServiceSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
	Status ServiceStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
}
```

I changed it to a pointer so that the data wouldn't be copied.

pkg/router/router.go
vtereso: The old function definition was
```go
func Register(resource endpoints.Resource) *restful.Container
```
so I continued to use that pattern.

pkg/router/router.go
vtereso: I don't think we need to, but since the return type of `restful.NewContainer()` is `*restful.Container` and the register functions expect this type I embedded the pointer.

pkg/router/router.go
a-roberts: 👍

pkg/router/router.go
a-roberts: 👍

pkg/router/routes_test.go
a-roberts: Sounds much better - excellent, thanks

pkg/controllers/kubernetes/extension.go
vtereso: Updated to use strings

PR #878, opened 2020-01-08
Show logs when no step name is provided

### LINKED ISSUE #709: Dashboard doesn't show details/logs of runs if no step names are provided
# Expected Behavior
I can view logs of task runs associated to pipeline runs.

# Actual Behavior
Dashboard renders an anchor "Not run" for each completed task run related to a pipeline run. 
# Steps to Reproduce the Problem

1. Install the files attached.
2. Create a pipeline run.
3. Click on the pipeline run created above.

# Additional Info
I can describe the resources associated to the pipeline run in question and view their logs through kubectl and tkn, but the dashboard always renders an anchor "Not run" for each task run and nothing else. There are no errors in the Javascript console and all requests have been succeeded, except this one:
```shell
Error
/proxy/apis/route.openshift.io/v1/:1 Failed to load resource: the server responded with a status of 404 (Not Found)
```
[files.tar.gz](https://github.com/tektoncd/dashboard/files/3820895/files.tar.gz)

ncskier:
# Changes
Fixes #709
A TaskRun has all of the information about the steps in its status
(including the default step names when no name is provided). So, I
changed the TaskRun container and PipelineRun component to only use the
TaskRun's status for retreiving information about the TaskRun
(previously they were using the Task definition as well).

Here's an example of the default step names when no name is provided:
Screen Shot 2020-01-08 at 2 18 45 PM

I also updated the TaskTree's Storybook entry to be more descriptive.

/cc @AlanGreene 
/cc @a-roberts 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRun/PipelineRun.js
+++ b/packages/components/src/components/PipelineRun/PipelineRun.js
@@ -20,6 +20,7 @@ import { FormattedMessage, injectIntl } from 'react-intl';
 import {
   getErrorMessage,
   getStatus,
+  reorderSteps,
   selectedTask,
   selectedTaskRun,
   stepsStatus,
@@ -118,7 +119,11 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
           };
           taskRun = theRun; // eslint-disable-line no-param-reassign
         } else {
-          steps = stepsStatus(task.spec.steps, taskRun.status.steps);
+          const reorderedSteps = reorderSteps(
+            taskRun.status.steps,
+            task.spec.steps
+          );
+          steps = stepsStatus(reorderedSteps, reorderedSteps);
         }
 
         return {
--- a/packages/utils/src/utils/index.js
+++ b/packages/utils/src/utils/index.js
@@ -112,6 +112,27 @@ export function stepsStatus(taskSteps, taskRunStepsStatus = []) {
   return steps;
 }
 
+/*
+  When steps are retreived from a TaskRun's status, their order is not guaranteed.
+  This function reorders the given unorderedSteps to match the order specified by the given orderedSteps.
+  Each step in unorderedSteps must have the same name as a step in the orderedSteps.
+  Unnamed steps are automatically given the name 'unnamed-NUM' where NUM is the step number starting from 1.
+    ex: 'unnamed-3' is the 3rd step
+  None of the unorderedSteps will have an empty name, but some of the orderedSteps may have empty names.
+*/
+export function reorderSteps(unorderedSteps, orderedSteps) {
+  if (!unorderedSteps || !orderedSteps) {
+    return [];
+  }
+  return orderedSteps.map(({ name }, idx) => {
+    let findName = name;
+    if (name === '') {
+      findName = `unnamed-${idx + 1}`;
+    }
+    return unorderedSteps.find(step => step.name === findName);
+  });
+}
+
 export function isRunning(reason, status) {
   return status === 'Unknown' && reason === 'Running';
 }
--- a/src/containers/TaskRun/TaskRun.js
+++ b/src/containers/TaskRun/TaskRun.js
@@ -24,7 +24,12 @@ import {
   StepDetails,
   TaskTree
 } from '@tektoncd/dashboard-components';
-import { getStatus, stepsStatus, taskRunStep } from '@tektoncd/dashboard-utils';
+import {
+  getStatus,
+  reorderSteps,
+  stepsStatus,
+  taskRunStep
+} from '@tektoncd/dashboard-utils';
 
 import { fetchLogs } from '../../utils';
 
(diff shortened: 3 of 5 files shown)


packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -118,7 +118,7 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
           };
           taskRun = theRun; // eslint-disable-line no-param-reassign
         } else {
-          steps = stepsStatus(task.spec.steps, taskRun.status.steps);
+          steps = stepsStatus(taskRun.status.steps, taskRun.status.steps);
```
AlanGreene: I remember now why we used the task spec

we may not have status for all steps, so would end up leaving some steps out of the list until they begin executing
using the task spec we get the full list of steps even if we don't have status for them yet

ncskier: Hm, in what case were you seeing that happen?

When I run a Pipeline, all of the steps seem to appear even before they begin executing:
Screen Shot 2020-01-08 at 4 05 22 PM

AlanGreene: yeah I'm not seeing it anymore either... ok let's go for it 👍

ncskier: Ok, thanks @AlanGreene! The PR's ready for review now 🙂

ncskier:
When working on this issue, I wanted to clean up some of the PipelineRun component and TaskRun container code; however, it is not within the scope of this issue. So, I've created issue https://github.com/tektoncd/dashboard/issues/885 to track the code clean up piece of work.

packages/utils/src/utils/index.js
```diff
@@ -112,6 +112,32 @@ export function stepsStatus(taskSteps, taskRunStepsStatus = []) {
...
+  When steps are retreived from a TaskRun's status, their order is not guaranteed.
+  This function reorders the given unorderedSteps to match the order specified by the given orderedSteps.
+  Each step in unorderedSteps must have the same name as a step in the orderedSteps.
+  Unnamed steps are automatically given the name 'unnamed-NUM' where NUM is the step number starting from 1.
+    ex: 'unnamed-3' is the 3rd step
+  None of the unorderedSteps will have an empty name, but some of the orderedSteps may have empty names.
+*/
+export function reorderSteps(unorderedSteps, orderedSteps) {
+  if (
+    typeof unorderedSteps === 'undefined' ||
```
AlanGreene: is `!unorderedSteps || !orderedSteps` sufficient here?

packages/utils/src/utils/index.js
```diff
@@ -112,6 +112,32 @@ export function stepsStatus(taskSteps, taskRunStepsStatus = []) {
...
+    typeof orderedSteps === 'undefined'
+  ) {
+    return [];
+  }
+  const steps = [];
+  orderedSteps.forEach(({ name }, idx) => {
```
AlanGreene: can use `map` instead of `forEach` + `push`

src/containers/TaskRun/TaskRun.js
```diff
@@ -101,14 +106,17 @@ export /* istanbul ignore next */ class TaskRunContainer extends Component {
...
       return null;
     }
-    let { steps } = taskRun.status;
-    if (task) {
-      steps = task.spec.steps; // eslint-disable-line
-    }
+
+    const { steps } = taskRun.status;
+    const stepDefinitions =
+      typeof taskRun.spec.taskSpec !== 'undefined'
```
AlanGreene: can this be just `taskRun.spec.taskSpec`?

packages/utils/src/utils/index.js
ncskier: Thanks Alan, I always forget that I can use that syntax with JavaScript 👍

packages/utils/src/utils/index.js
ncskier: Ooo, good point, nice!

ncskier:
It says the unit tests have been running for an hour 😓 

ncskier:
Hmmm, looks like the tests worked now?

@AlanGreene thanks for the review! I've addressed your suggestions, so please take another look when you get the chance 🙂

packages/components/src/components/PipelineRun/PipelineRun.js
AlanGreene: @ncskier looks like there was another reason for `task.spec.steps` after all, it was used to populated the step definition data displayed on the details tab

with this change we're now showing essentially the same status information (only difference being the step id on the details tab) on both the status and details tabs

I think reverting this line here and in TaskRun should resolve this issue without affecting the display of logs when no step name is given. Testing it out now... 🤞

Issue #868, opened 2020-01-07
Left nav not showing selected resource type when viewing a single trigger resource


AlanGreene:
# Expected Behavior
The left nav should always show the selected resource type, as it does for Pipelines, runs, etc.
![image](https://user-images.githubusercontent.com/2829095/71893450-9fb17b00-3143-11ea-8da8-0f234f25dff9.png)

# Actual Behavior
Follow up from https://github.com/tektoncd/dashboard/pull/784#pullrequestreview-339181730

When viewing a list of trigger resources (i.e. EventListeners, TriggerBindings, TriggerTemplates) the nav item is selected as expected.
![image](https://user-images.githubusercontent.com/2829095/71893472-b22bb480-3143-11ea-96aa-80acdce68921.png)

When viewing a single trigger resource there is no nav item selected.
![image](https://user-images.githubusercontent.com/2829095/71893503-ca033880-3143-11ea-9029-8c9231626443.png)

# Additional Info
In SideNav.js, the `to` prop of the SideNavMenuItems should be wrapped in a call to `this.getPath` to ensure it accounts for the namespace in the URL.

Once this is done the nav items should be selected as expected, but there may be other changes required to ensure the namespace isn't doubled in the constructed URLs / API calls for example.

carlos-logro:
merged so closing.
/close

PR #887, opened 2020-01-11
Show Selected Resource Type when Viewing a Single Trigger

carlos-logro:
issue: #868 

# Changes
Add `this.getPath` to each trigger resource in the SideNav `to` prop.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -200,13 +200,16 @@ export /* istanbul ignore next */ class App extends Component {
                     exact
                     component={EventListeners}
                   />
-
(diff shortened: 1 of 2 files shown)


AlanGreene:
@carlos-logro looks like we're missing a `<Route>` in App.js for these resource types that'll handle the `byNamespace` URLs. See Pipelines for example.

The containers for these resources would also need to be updated to read the namespace from the URL (`props.match.params`), and fallback to the selected namespace if none found. Again see src/containers/Pipelines for example.

PR #896, opened 2020-01-14
Allow consumers of the components to disable the byPipeline link on PipelineRuns page

AlanGreene:
# Changes

Allow consumers of the PipelineRuns component to disable the
link to the filtered list of PipelineRuns by pipelineName.

Previously we supported hiding the Pipeline column entirely but
this is not desirable. Instead we want to allow consumers to
render the Pipeline name as plain text instead of a clickable
link.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRuns/PipelineRuns.js
+++ b/packages/components/src/components/PipelineRuns/PipelineRuns.js
@@ -50,7 +50,6 @@ const PipelineRuns = ({
     return `${reason}: ${message}`;
   },
   hideNamespace,
-  hidePipeline,
   intl,
   loading,
   selectedNamespace,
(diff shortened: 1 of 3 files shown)


packages/components/src/components/PipelineRuns/PipelineRuns.js
```diff
@@ -141,27 +147,21 @@ const PipelineRuns = ({
...
-            title={pipelineRefName}
-          >
+        pipelineRefName &&
+        (pipelineRunsByPipelineURL ? (
+          <Link to={pipelineRunsByPipelineURL} title={pipelineRefName}>
             {pipelineRefName}
           </Link>
         ) : (
-          ''
+          pipelineRefName
```
steveodonovan: Will this cause bad rendering for inline pipeline runs ?

AlanGreene: `pipelineRefName` will be falsey in that case so nothing gets rendered as expected
example included in the storybook with no pipelineRef

PR #901, opened 2020-01-15
Add Cache-Control header to static resources to help caching

nzmsv:
# Changes

Prior to this change putting the dashboard behind ALB broke static resource caching.  ALB seems to remove the Date header from the response it proxies, however it will obey explicit Cache-Control headers.

Static resources coming from webpack already have a content hash as part of their name so they are already immutable.

Note: the first draft of this PR used the Etag header instead.  See the discussion below.

# Submitter Checklist

review them:

- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -20,6 +20,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -59,6 +60,8 @@ const ExtensionDisplayNameKey = "tekton-dashboard-display-name"
 const ExtensionRoot = "/v1/extensions"
 
 var webResourcesDir = os.Getenv("WEB_RESOURCES_DIR")
+var webResourcesStaticPattern = regexp.MustCompile("^/([[:alnum:]]+\\.)?[[:alnum:]]+\\.(js)|(css)|(png)$")
+var webResourcesStaticExcludePattern = regexp.MustCompile("^/favicon.png$")
 
 // Register returns an HTTP handler that has the Dashboard REST API registered
 func Register(resource endpoints.Resource) *Handler {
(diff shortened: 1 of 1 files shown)


a-roberts:
/ok-to-test
@AlanGreene FYI

nzmsv:
> @nzmsv thanks for the PR, adding caching support would be very useful for a lot of use cases.
> 
> Since the URLs for the static resources already contain a content-based hash we know that content can be cached long-term without requiring periodic revalidation as in the case of using entity tags or similar approaches. Instead of sending the Etag header in the response, we could set a long max-age on the Cache-Control header and eliminate the conditional requests from the browser.
> 
> Something like `Cache-Control: public, max-age=31536000, immutable`
> 
> Would this work for your use case?

Thanks, this is a great idea!  I tested the Cache-Control header behind ALB and it works correctly.  I updated the PR accordingly.

pkg/router/router.go
```diff
@@ -59,6 +60,7 @@ const ExtensionDisplayNameKey = "tekton-dashboard-display-name"
 const ExtensionRoot = "/v1/extensions"
 
 var webResourcesDir = os.Getenv("WEB_RESOURCES_DIR")
+var webResourcesPattern = regexp.MustCompile("^/([[:alnum:]]+\\.)?[[:alnum:]]+\\.(js)|(css)|(png)$")
```
AlanGreene: The favicon (the only png I see loaded) does not include a hash, so if we can exclude it for now this PR should be good to go. We'll need to make some changes on the frontend for the favicon so I'll create an issue to track that separately.

```suggestion
var webResourcesPattern = regexp.MustCompile("^/([[:alnum:]]+\\.)?[[:alnum:]]+\\.(js)|(css)$")
```

nzmsv: We also load the Tekton logo png (in the upper right corner).  I'll add a specific exclusion for the favicon.

Issue #782, opened 2019-11-27
Steps displayed after a failed one are marked as "not-run"


afrittoli:
# Expected Behavior
Steps are marked as pass/failed/not-run matching the actual status, regardless of their order.

# Actual Behavior
If one step fails, steps displayed after that are marked as not-run even if they were executed, and logs are displayed correctly.

This is probably a side effect to the fact that steps are not displayed in order of execution to begin with, nonetheless the UI should not rely on that to assume that a step was not run.

In my case, the step with a failure was the last one to be executed. See attached screenshot.
![image (2)](https://user-images.githubusercontent.com/2205608/69734271-9980b580-1126-11ea-8e90-88b6a42fbb09.png)

# Steps to Reproduce the Problem

1.
2.
3.

# Additional Info

AlanGreene:
Backing out https://github.com/tektoncd/dashboard/pull/653 should resolve this now that tektoncd/pipeline correctly sets the status of these steps. We just need to make sure we're ok in terms of versions for our next planned kabanero supporting release.

a-roberts:
> Backing out #653 should resolve this now that tektoncd/pipeline correctly sets the status of these steps. We just need to make sure we're ok in terms of versions for our next planned kabanero supporting release.

Happy for master of dashboard to (to our best ability) work with the latest version of Tekton Pipelines, so yeah anybody who picks this up should do the backout of the commit Al refers to

Worst case scenario for Kabanero is that we cherry pick any juicy commits and remember this one incase something does come up

PR #653, opened 2019-10-18
Fix error showing TaskRun steps that have not run

ncskier:
# Changes

Fixes #644
This fix shows a neutral "Not run" status for steps that follow a step with an
error. Previously, these steps showed a misleading green "Completed" status.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/dashboard/blob/master/CONTRIBUTING.md)
for more details._

PR #906, opened 2020-01-16
Fix error showing TaskRun steps that have not run (again)

ncskier:
# Changes

This fixes an error seen when using the Tekton Dashboard with Tekton
Pipelines v0.8.0. In order to support users on Tekton Pipelines v0.8.0
and earlier, the fix from https://github.com/tektoncd/dashboard/pull/653 is going back into the Dashboard codebase.
The fix was previously reverted when Tekton Pipelines v0.9.0 fixed the
root cause of the issue (https://github.com/tektoncd/dashboard/pull/795/files).

Screen Shot 2020-01-16 at 3 15 54 PM

This issue was brought to my attention by a user running the Dashboard v0.3.0 with Tekton Pipelines v0.8.0: https://github.com/kabanero-io/kabanero-pipelines/issues/86#issuecomment-574892024

Note, we don't want to reintroduce issue https://github.com/tektoncd/dashboard/issues/782 with this change.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRun/PipelineRun.js
+++ b/packages/components/src/components/PipelineRun/PipelineRun.js
@@ -24,7 +24,8 @@ import {
   selectedTask,
   selectedTaskRun,
   stepsStatus,
-  taskRunStep
+  taskRunStep,
+  updateUnexecutedSteps
 } from '@tektoncd/dashboard-utils';
 
 import { Log, RunHeader, StepDetails, TaskTree } from '..';
@@ -261,7 +262,10 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
 
     const { definition, reason, status, stepName, stepStatus } = taskRunStep(
       selectedStepId,
-      taskRun
+      {
+        ...taskRun,
+        steps: updateUnexecutedSteps(taskRun.steps)
+      }
     );
 
     const logContainer = (
--- a/packages/components/src/components/Task/Task.js
+++ b/packages/components/src/components/Task/Task.js
@@ -17,6 +17,8 @@ import ChevronRight from '@carbon/icons-react/lib/chevron--right/16';
 import CloseFilled from '@carbon/icons-react/lib/close--filled/16';
 import { Spinner, Step } from '@tektoncd/dashboard-components';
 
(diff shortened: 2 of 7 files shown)


AlanGreene:
Do we still have the issue described in https://github.com/tektoncd/dashboard/issues/782 with this workaround?

ncskier:
@AlanGreene if I understand correctly, https://github.com/tektoncd/dashboard/issues/782 was only an issue because the steps were being displayed out of order (cc @afrittoli is this correct?).

A side-effect of PR https://github.com/tektoncd/dashboard/pull/878 was that it introduced the `reorderSteps()` function which orders steps in the UI according to the Task definition (code [here](https://github.com/tektoncd/dashboard/pull/878/files#diff-44384c8c75c7b5aab1477fe8f0736133R123)).

So, I don't think that https://github.com/tektoncd/dashboard/issues/782 is a concern any longer. However, I'm having trouble reproducing a PipelineRun with an out-of-order status. @AlanGreene where would I put a unit test for this? I don't see any `TaskRun.test.js` files.

a-roberts:
@ncskier @AlanGreene good catch, I propose we do this PR into v0.4.0 and then recreate the 0.4.0 release using it, this would ensure our current master branch is still somewhat sync with Tekton Pipelines *latest we can work with*, which is going to be 0.9.2 atm

Otherwise we're going to have this code enter master, we'll do the release pipeline and then we'd have to revert the change immediately after.

Thoughts?

AlanGreene:
I can't reproduce the ordering issue either, even without #878, tried in a clean environment with pipelines 0.8. I'm fine with adding this back to master and leaving it, showing the steps as 'not run' instead of 'failed' seems a better reflection of what actually happened.

/lgtm

a-roberts:
> I can't reproduce the ordering issue either, even without #878, tried in a clean environment with pipelines 0.8. I'm fine with adding this back to master and leaving it, showing the steps as 'not run' instead of 'failed' seems a better reflection of what actually happened.
> 
> /lgtm

Ah, hadn't appreciated that, yeah let's get it in then

/approve

Issue #919, opened 2020-01-20
Trigger resources not displayed on OpenShift


a-roberts:
Surely introduced by https://github.com/tektoncd/dashboard/issues/815

On our 0.3 release, no check is made to see if the EventListener CRD is known
It is with the 0.4 release

Request details

```
Request URL: https://tekton-dashboard-tekton-pipelines.apps.myopenshiftsystem.ibm.com/proxy/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions/eventlisteners.tekton.dev
Request Method: GET
Status Code: 404 Not Found
Remote Address: 9.30.254.155:443
Referrer Policy: no-referrer-when-downgrade
```

This could be related but the k8s version is very different (1.13+gke), so assume it's different: https://github.com/tektoncd/dashboard/issues/904

There's nothing for https://tekton-dashboard-tekton-pipelines.apps.myopenshiftsystem.ibm.com/proxy/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions

Pod error is an `unknown` so I'm gonna guess it's RBAC

a-roberts:
Fixed with

```
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  verbs:
  - get
  - list
```
 in the cluster role, PR coming up

a-roberts:
PR https://github.com/tektoncd/dashboard/pull/920

a-roberts:
Merged and in the 0.4.1 release I've just done, closing

PR #920, opened 2020-01-20
Add RBAC for getting CRD info, used to detect Triggers installation

a-roberts:
For https://github.com/tektoncd/dashboard/issues/919

Without it we get a cryptic unknown error from the proxy and Trigger resources aren't displayed on master (including the 0.4.0 release)

Fixed about 30 linter errors too

### DIFF
--- a/config/dev-tekton-dashboard.yaml
+++ b/config/dev-tekton-dashboard.yaml
@@ -1,5 +1,5 @@
 # ------------------- Dashboard Service Account ------------------- #
-
+---
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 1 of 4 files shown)


a-roberts:
@mnuttall FYI - I've resolved the yaml linter pain in this commit, be aware you'll have to pick these lovely changes up for your Kustomize work

yamllint was actually added to plumbing ages ago, it's just that we've not touched the yaml until now 😄

AlanGreene:
So the interesting change here is adding the following to the ClusterRole:
```yaml
- apiGroups: ["apiextensions.k8s.io"]
    resources: ["customresourcedefinitions"]
    verbs: ["get", "list"]
```

My environment mustn't have been as clean as I thought when I tested this, thanks for the fix 😸 

Looks good
/lgtm

a-roberts:
@AlanGreene exactly, thanks for taking a look - it looks like we've still got a flaky e2e test going on (@CarolynMabbott FYI, I think you made a change recently to look for the dashboard being made up but perhaps it still needs a little work?)

a-roberts:
```
I0120 15:46:48.839] error: error validating "STDIN": error validating data: ValidationError(Deployment.spec.template.spec.containers[0].resources): unknown field "env" in io.k8s.api.core.v1.ResourceRequirements; if you choose to ignore these errors, turn validation off with --validate=false
```

oh, looking

a-roberts:
Hit the other flaky now, what a time to be alive (TestExtension this time)

Issue #961, opened 2020-01-27
Step details tab does not show the step definition


AlanGreene:
# Expected behavior
The step details tab on both the PipelineRun and TaskRun pages should show the definition of the selected step (i.e. command, args, etc.)

# Actual behavior
The step details tab displays the same information as the status tab with the addition of the step id.

# Additional Info
It looks like this was introduced by #878
https://github.com/tektoncd/dashboard/pull/878#discussion_r371201832

PR #908, opened 2020-01-17
Fix unnamed steps bug

ncskier:
# Changes

Seeing "Cannot read property 'name' of undefined" error for Tasks with
unnamed steps. This was supposed to be fixed in PR #878
https://github.com/tektoncd/dashboard/pull/878)

The bug appears to be with the step names starting from 0 instead of 1.

### Breaking Change
This fixes a bug related to the Tekton Pipelines version. Tekton Pipelines v0.8.0 names unnamed steps starting from `1`. On the other hand, Tekton Pipelines v0.9.0 and up name unnamed steps starting from `0`. So, this PR will _**"break" the Dashboard for Tekton Pipelines v0.8.0**_. However, this PR will _**"fix" the Dashboard for Tekton Pipelines v0.9.0 and up**_.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/dashboard/blob/master/CONTRIBUTING.md)
for more details._

# Release Notes
```
When using Tekton Pipelines v0.8.0 or lower, TaskRuns & PipelineRuns that run
unnamed steps will no longer be displayed properly on the Dashboard.
```

PR #962, opened 2020-01-27
Fix display of step details

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/961

Update logic for display of step status and step details tabs
so the correct information (command, args, script, etc.) are
available to display in the step details tab.

This is a temporary measure to fix a recent regression, and
this code will undergo significant refactoring as part of the
updated designs which remove the status tab entirely.

Before:
image

After:
image

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/StepDefinition/StepDefinition.js
+++ b/packages/components/src/components/StepDefinition/StepDefinition.js
@@ -106,6 +106,15 @@ class StepDefinition extends Component {
 
   render() {
     const { definition, intl } = this.props;
+    const { container, imageID, terminated, ...stepDefinition } =
+      definition || {};
+
+    const resource = definition
+      ? stepDefinition
+      : intl.formatMessage({
+          id: 'dashboard.step.definitionNotAvailable',
+          defaultMessage: 'description: step definition not available'
+        });
 
     const paramsResources = this.getIOTables();
     return (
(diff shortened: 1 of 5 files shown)


a-roberts:
@AlanGreene sounds super important 👀 , so is this something that works with Tekton 0.10 but not older versions? And does this fix:

image

AlanGreene:
This particular change would be fine with older Tekton versions since it's just modifying how we pass existing data around internally in the client, but a previous fix for a different issue in the same area (https://github.com/tektoncd/dashboard/pull/908) requires 0.9+

I'm not seeing that error on the PipelineRun page, can you share repro details?

a-roberts:
> This particular change would be fine with older Tekton versions since it's just modifying how we pass existing data around internally in the client, but a previous fix for a different issue in the same area (#908) requires 0.9+
> 
> I'm not seeing that error on the PipelineRun page, can you share repro details?

Ok, sounds good to me then

So in my case I've got this and I've only just noticed today, I don't mean to hijack this issue although this could be related (see attachment)

(p.s it's .yaml but .txt, but GitHub file extensions for attachments...)
[pipelinerunnameerror.txt](https://github.com/tektoncd/dashboard/files/4117477/pipelinerunnameerror.txt)

AlanGreene:
I'm not seeing anything obvious in the data that would cause it, most likely a timing issue with loading the task definitions or something along those lines. It's possible that this change might resolve it since we're less likely to be missing a step / name now in earlier processing passes, but I suspect it won't have much impact.

a-roberts:
> I'm not seeing anything obvious in the data that would cause it, most likely a timing issue with loading the task definitions or something along those lines. It's possible that this change might resolve it since we're less likely to be missing a step / name now in earlier processing passes, but I suspect it won't have much impact.

Yep, fair enough. I'm happy to get this in 😄 

image

My PipelineRun now renders and the details look good now

/lgtm
/approve

PR #954, opened 2020-01-24
Comment Extension Test

vtereso:
Fixes: https://github.com/tektoncd/dashboard/issues/902
# Changes
Comment out the extension test due to unreliable informer

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

Issue #902, opened 2020-01-16
Flaky test extension problems (timeout deleting extension)


a-roberts:
# Expected Behavior
Our unit tests should reliably pass

# Actual Behavior

```
I0116 11:06:46.485] --- FAIL: TestExtension (1.32s)
I0116 11:06:46.485]     --- FAIL: TestExtension/Empty_endpoints (1.08s)
I0116 11:06:46.486]         routes_test.go:444: Timed out waiting for extension deletion
I0116 11:06:46.486]     --- PASS: TestExtension/One_endpoint (0.08s)
I0116 11:06:46.486]     --- PASS: TestExtension/Two_endpoints (0.07s)
I0116 11:06:46.486]     --- PASS: TestExtension/Three_endpoint (0.10s)
I0116 11:06:46.487] === RUN   TestMarshalJSON_Extension
I0116 11:06:46.487] === RUN   TestMarshalJSON_Extension/Extension_1
I0116 11:06:46.487] --- PASS: TestMarshalJSON_Extension (0.00s)
I0116 11:06:46.487]     --- PASS: TestMarshalJSON_Extension/Extension_1 (0.00s)
I0116 11:06:46.487] FAIL
I0116 11:06:46.487] FAIL	github.com/tektoncd/dashboard/pkg/router	1.798s
I0116 11:06:46.488] ?   	github.com/tektoncd/dashboard/pkg/testutils	[no test files]
I0116 11:06:46.488] ?   	github.com/tektoncd/dashboard/pkg/utils	[no test files]
I0116 11:06:46.488] ?   	
```

seen with https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_dashboard/901/pull-tekton-dashboard-unit-tests/1217759636258557952/

# Steps to Reproduce the Problem

1. Run our unit tests for a PR on Prow (should be able to do this locally too)

# Additional Info
We have a long history with our extension testing 😿

vtereso:
I have tracked it down to the informer event handler not being called (~4/100 times locally).

vtereso:
I raised an issue regarding the informers: https://github.com/kubernetes/kubernetes/issues/87472#issuecomment-577881040. It was decided to comment out the test again.

mnuttall:
Test disabled in another PR.

PR #973, opened 2020-01-29
Remove flaky extension test

a-roberts:
@vtereso mentioned being away for many weeks recently so https://github.com/tektoncd/dashboard/pull/954 isn't likely to be addressed soon, and I really want our tests to stop being flaky - so remove the culprit

This is for https://github.com/tektoncd/dashboard/issues/902

### DIFF
--- a/pkg/router/routes_test.go
+++ b/pkg/router/routes_test.go
@@ -11,19 +11,13 @@ import (
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


mnuttall:
/lgtm
/approve

We'll look to reinstate this if we start seeing more dashboard extensions.

PR #935, opened 2020-01-22
Get versions for dashboard and pipelines

CarolynMabbott:
# Changes

First pass of getting version of dashboard and pipelines 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/dashboard/blob/master/CONTRIBUTING.md)
for more details._

PR #983, opened 2020-01-30
Add about page to display versions

CarolynMabbott:
# Changes
Get versions for dashboard and pipelines 
Display in an about page 

# Submitter Checklist

review them:

for more details._


pkg/endpoints/cluster.go
```diff
@@ -162,3 +164,70 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
 	properties := Properties{InstallNamespace: os.Getenv("INSTALLED_NAMESPACE")}
 	response.WriteEntity(properties)
 }
+
+// Get dashboard version
+func (r Resource) GetDashboardVersion(request *restful.Request, response *restful.Response) {
+	config, err := rest.InClusterConfig()
```
a-roberts: Surely we've got one of these we should be reusing instead of creating new ones in both methods

pkg/endpoints/cluster.go
```diff
@@ -162,3 +164,70 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
...
+	deployments, err := api1.Deployments(properties.InstallNamespace).List(listOptions)
+
+	for _, deployment := range deployments.Items {
+		deploymentLabels := deployment.GetLabels()
+		version = deploymentLabels["version"]
+	}
+
+	if version == "" {
+		response.WriteEntity("Unknown")
+		return
```
a-roberts: We could return an error here and catch on the frontend, displaying some meaningful warning (instead of a normal return which is a http code 200)

pkg/endpoints/cluster.go
```diff
@@ -162,3 +164,70 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
 	properties := Properties{InstallNamespace: os.Getenv("INSTALLED_NAMESPACE")}
 	response.WriteEntity(properties)
 }
+
+// Get dashboard version
+func (r Resource) GetDashboardVersion(request *restful.Request, response *restful.Response) {
```
a-roberts: No return type here - either return a nice response or the actual version as a string, and/or an error

pkg/endpoints/cluster.go
```diff
@@ -162,3 +164,70 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
...
+	if version == "" {
+		response.WriteEntity("Unknown")
+		return
+	}
+	response.WriteEntity(version)
+	return
+}
+
+// Get pipelines version
+func (r Resource) GetPipelineVersion(request *restful.Request, response *restful.Response) {
```
a-roberts: Same comments for the previous one, we could potentially combine this with the previous (and change the selector?) - we'd reduce plenty code dupe

pkg/endpoints/cluster.go
```diff
@@ -162,3 +164,70 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
...
+		LabelSelector: "app.kubernetes.io/component=controller,app.kubernetes.io/name=tekton-pipelines",
+	}
+	deployments, err := api1.Deployments("tekton-pipelines").List(listOptions)
+
+	for _, deployment := range deployments.Items {
+		deploymentAnnotations := deployment.Spec.Template.GetAnnotations()
+		version = deploymentAnnotations["tekton.dev/release"]
+
+		if version == "" {
+			deploymentImage := deployment.Spec.Template.Spec.Containers[0].Image
```
a-roberts: Is it always going to be this container?

pkg/endpoints/cluster.go
```diff
@@ -162,3 +164,70 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
...
+	response.WriteEntity(version)
+	return
+}
+
+// Get pipelines version
+func (r Resource) GetPipelineVersion(request *restful.Request, response *restful.Response) {
+	config, err := rest.InClusterConfig()
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
```
a-roberts: Really 👀

pkg/endpoints/cluster.go
```diff
@@ -162,3 +164,70 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
...
+	deployments, err := api1.Deployments("tekton-pipelines").List(listOptions)
+
+	for _, deployment := range deployments.Items {
+		deploymentAnnotations := deployment.Spec.Template.GetAnnotations()
+		version = deploymentAnnotations["tekton.dev/release"]
+
+		if version == "" {
+			deploymentImage := deployment.Spec.Template.Spec.Containers[0].Image
+			if strings.Contains(deploymentImage, "pipeline/cmd/controller") {
+				s := strings.SplitAfter(deploymentImage, ":")
```
a-roberts: Make sure these are safe/error checked for $(go panic reduction chance--)

pkg/router/routes_test.go
```diff
@@ -35,14 +35,16 @@ import (
...
+	"/v1/websockets",   // No response code
+	"rerun",            // Returns 201
+	ExtensionRoot,      // Response codes dictated by extension logic
+	"health",           // Returns 204
+	"readiness",        // Returns 204
+	"proxy",            // Kube API server has its own standard
+	"ingress",          // Ingress will not exist
+	"endpoints",        // Route or Ingress will not exist
+	"dashboardversion", // Returns version of dashboard running
+	"pipelineversion",  // Returns version of the pipelines running
```
a-roberts: `of Tekton Pipelines`

src/actions/about.js
```diff
@@ -0,0 +1,25 @@
+/*
+Copyright 2019 The Tekton Authors
```
a-roberts: 2020

src/actions/about.js
```diff
@@ -0,0 +1,25 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { getDashboardVersion, getPipelineVersion } from '../api';
+
+export async function fetchDashboardInfo(namespace) {
```
a-roberts: I'm pretty sure this isn't actually an action, so for example notice async dispatch usage at https://github.com/tektoncd/dashboard/blob/master/src/actions/secrets.js, this is more like an API function for comms or index

src/actions/about.test.js
```diff
@@ -0,0 +1,26 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import * as API from '../api';
+import { fetchDashboardInfo } from './about';
+
+it('fetchServiceAccounts', async () => {
```
a-roberts: Test name

src/containers/About/About.js
```diff
@@ -0,0 +1,122 @@
+/*
+Copyright 2019 The Tekton Authors
```
a-roberts: 2020

pkg/router/router.go
```diff
@@ -218,6 +218,8 @@ func registerEndpoints(r endpoints.Resource, container *restful.Container) {
 	wsv1.Route(wsv1.POST("/{namespace}/rerun").To(r.RerunPipelineRun))
 	wsv1.Route(wsv1.GET("/{namespace}/ingress").To(r.GetIngress))
 	wsv1.Route(wsv1.GET("/{namespace}/endpoints").To(r.GetEndpoints))
+	wsv1.Route(wsv1.GET("/{namespace}/dashboardversion").To(r.GetDashboardVersion))
```
AlanGreene: as mentioned on the [previous PR](https://github.com/tektoncd/dashboard/pull/935#discussion_r369502053) would it make sense to include this in the properties endpoint so we can get them all together instead of requiring separate calls for each value?

src/containers/About/About.js
```diff
@@ -0,0 +1,122 @@
...
+          value: pipelineVersion
+        };
+      }
+
+      initialRows.push(data);
+    });
+
+    return (
+      <>
+        <h1>About</h1>
```
a-roberts: We should have i18n for this string

packages/components/src/components/Breadcrumbs/Breadcrumbs.js
```diff
@@ -46,6 +46,7 @@ export default function Breadcrumbs({ labels, match }) {
 }
 
 const breadcrumbLabels = {
+  '/about': 'About',
```
AlanGreene: the Breadcrumbs component is not currently used and its purpose / behaviour will be changing so no need to include changes here for now

src/containers/About/About.test.js
```diff
@@ -0,0 +1,66 @@
+/*
+Copyright 2019 The Tekton Authors
```
a-roberts: 2020

pkg/endpoints/cluster.go
```diff
@@ -162,3 +164,70 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
...
+		}
+	}
+
+	if version == "" {
+		response.WriteEntity("Unknown")
+		return
+	}
+	response.WriteEntity(version)
+	return
+}
```
AlanGreene: should we also provide the triggers version?

pkg/router/router.go
```diff
@@ -218,6 +218,8 @@ func registerEndpoints(r endpoints.Resource, container *restful.Container) {
...
+	wsv1.Route(wsv1.GET("/{namespace}/pipelineversion").To(r.GetPipelineVersion))
```
AlanGreene: Do we need to include the namespace in the URL?

pkg/endpoints/cluster.go
CarolynMabbott: Triggers will be included in the future

pkg/endpoints/cluster.go
CarolynMabbott: It's been consistent throughout 0.10.0 / 0.10.1 releases 
Before that is the code to get below (0.9.0)

src/actions/about.js
CarolynMabbott: Have moved this function to index

packages/components/src/components/Breadcrumbs/Breadcrumbs.js
CarolynMabbott: Removed from breadcrumbs

pkg/endpoints/cluster.go
CarolynMabbott: I have removed them in both functions and now am using the K8sClient one instead and combined it with getting the deployments

pkg/endpoints/cluster.go
CarolynMabbott: I have put errors in if cant get the deployment 
I have put errors in if it is unknown and it returns a 500 and the front end just doesn't display that field for now

pkg/endpoints/cluster.go
CarolynMabbott: Have returned an error if there is an error in getting the deployment
Have changed the errors, same as above

It's not just the selectors which are different, it's the namespace which it searches as well as when the PR for openshift pipelines comes in it'll also be checking for that install as well

pkg/endpoints/cluster.go
CarolynMabbott: removed this

pkg/endpoints/cluster.go
CarolynMabbott: I have added some safe / error checking to make sure that it doesnt break or return the wrong version back

pkg/endpoints/cluster.go
CarolynMabbott: I have put errors in if cant get the deployment 
I have put errors in if it is unknown and it returns a 500 
It however doesnt display any error on the dashboard yet that says there was an issue, the logs will say there was an issue

pkg/router/router.go
AlanGreene: pasting slack message here for reference:

if you add the versions into the existing `/v1/properties` endpoint you won’t need to register new URLs or worry about namespace
see:
- https://github.com/tektoncd/dashboard/blob/a0866ee7b14bbc81bd97a5ccea0bce4c8bd6bcd9/pkg/router/router.go#L268
- https://github.com/tektoncd/dashboard/blob/a0866ee7b14bbc81bd97a5ccea0bce4c8bd6bcd9/pkg/endpoints/cluster.go#L161

we could have a response something like:

```
{
  installNamespace: 'tekton-pipelines',
  versions: {
    dashboard: '0.5.0',
    pipeline: '0.10.1',
    triggers: '0.2.1'
  }
}
```

or have each version at the same level as the InstallNamespace

Having them all in one API simplifies things from the front end, as we only need to make 1 request on page load to get the info, and it’s unlikely to change.

CarolynMabbott: Have added these to the properties endpoint

src/containers/About/About.js
a-roberts: 👀

src/nls/messages_en.json
```diff
@@ -126,6 +126,8 @@
     "dashboard.step.definitionNotAvailable": "description: step definition not available",
     "dashboard.step.statusNotAvailable": "No status available",
     "dashboard.step.stepDefinition": "Step definition",
+    "dashboard.tableHeader.Propety": "Property",
```
a-roberts: Property*

CarolynMabbott: Fixed these

pkg/router/router.go
CarolynMabbott: removed these now in favour of using the properties

pkg/endpoints/dashboard.go
```diff
@@ -0,0 +1,100 @@
...
+				if strings.Contains(s[1], "@") {
+					t := strings.Split(s[1], "@")
+					version = t[0]
+				}
+			}
+		}
+	}
+
+	if version == "" {
+		logging.Log.Errorf("Error getting the tekton pipelines deployment version. Version is unknown")
```
a-roberts: Errorf with no actual message

pkg/router/router.go
```diff
@@ -218,6 +218,8 @@ func registerEndpoints(r endpoints.Resource, container *restful.Container) {
 	wsv1.Route(wsv1.POST("/{namespace}/rerun").To(r.RerunPipelineRun))
 	wsv1.Route(wsv1.GET("/{namespace}/ingress").To(r.GetIngress))
 	wsv1.Route(wsv1.GET("/{namespace}/endpoints").To(r.GetEndpoints))
+	// wsv1.Route(wsv1.GET("/{namespace}/dashboardversion").To(r.GetDashboardVersion))
```
a-roberts: Comments

src/api/index.js
```diff
@@ -467,3 +467,8 @@ export function getEventListener({ name, namespace }) {
   const uri = getTektonAPI('eventlisteners', { name, namespace });
   return get(uri);
 }
+
+export async function fetchDashboardInfo() {
```
a-roberts: Function seems pointless, why not just await on getInstallProperties for the caller?

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+      }
+
+      return (
+        <>
+          {this.state.error !== '' && (
+            <InlineNotification
+              kind="error"
+              title={intl.formatMessage({
+                id: 'dashboard.displayVersion.error',
+                defaultMessage: 'Error Getting Version'
```
a-roberts: Dubious capitalisation here 👀

src/containers/About/About.test.js
```diff
@@ -0,0 +1,142 @@
...
+  );
+
+  await expect(API.fetchDashboardInfo).toHaveBeenCalledTimes(1);
+  expect(queryByText('Property')).toBeTruthy();
+  expect(getByText('Value')).toBeTruthy();
+  expect(getByText('Install Namespace')).toBeTruthy();
+  expect(getByText('Pipeline Version')).toBeTruthy();
+  expect(getByText('tekton-pipelines')).toBeTruthy();
+  expect(getByText('v0.10.0')).toBeTruthy();
+  expect(getByText('Error Getting Version')).toBeTruthy();
```
a-roberts: Case

src/api/index.js
CarolynMabbott: Have removed this now

a-roberts:
Changes look mostly OK now, @AlanGreene can we get your thoughts as well for this one again please?

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+    super(props);
+    this.state = initialState;
+  }
+
+  componentDidMount() {
+    this.fetchDashboardInfo();
+  }
+
+  setErrorState(errorsFound) {
+    let errorMessage = 'Error getting version of ';
```
AlanGreene: we need to use react-intl for user-facing strings, also can't use concatenation to build these strings due to ordering / syntax differences in many languages

for building the list we can't manually build comma-separated strings, instead we should use react-intl's formatted list support, see https://github.com/formatjs/react-intl/blob/master/docs/Components.md#formattedlist

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  render() {
+    const { intl, loading } = this.props;
+    if (this.state.dashboardInfo !== null && this.state.isLoaded === true) {
+      const initialHeaders = [
+        {
+          key: 'property',
+          header: intl.formatMessage({
+            id: 'dashboard.tableHeader.Property',
```
AlanGreene:
```suggestion
            id: 'dashboard.tableHeader.property',
```

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+            defaultMessage: 'Property'
+          })
+        },
+        {
+          key: 'value',
+          header: intl.formatMessage({
+            id: 'dashboard.tableHeader.Value',
```
AlanGreene:
```suggestion
            id: 'dashboard.tableHeader.value',
```

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { injectIntl } from 'react-intl';
+import { Table } from '@tektoncd/dashboard-components';
+import { InlineNotification } from 'carbon-components-react';
+import { getInstallProperties } from '../../api';
+import { isWebSocketConnected } from '../../reducers';
+
+const initialState = {
+  dashboardInfo: [],
```
AlanGreene: this is being used as an object later, so we should be consistent and use an empty object for the initialState

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+        }
+      ];
+
+      const initialRows = [];
+      let data = '';
+      const errorsToDisplay = [];
+
+      if (
+        this.state.dashboardInfo.InstallNamespace !== '' &&
+        this.state.dashboardInfo.InstallNamespace !== 'undefined'
```
AlanGreene: What's the scenario where the properties API would return a string 'undefined'? Seems unexpected. Can we change this value so it's easier to consume? `null` or an empty string would both work.

side: 'undefined' is a valid name for a namespace

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+      ) {
+        data = {
+          id: `InstallNamespace`,
+          property: 'Install Namespace',
```
AlanGreene: react-intl, same for the version labels below

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+          value: this.state.dashboardInfo.InstallNamespace
+        };
+        initialRows.push(data);
+      } else {
+        errorsToDisplay.push('Install Namespace');
+      }
+
+      if (
```
AlanGreene: Can we reduce the duplication here? We have essentially the same block repeated 3 times, once for each of the properties.

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+              kind="error"
+              title={intl.formatMessage({
+                id: 'dashboard.displayVersion.error',
+                defaultMessage: 'Error getting version'
+              })}
+              subtitle={this.state.error}
+              lowContrast
+            />
+          )}
+          <h1>About</h1>
```
AlanGreene: react-intl

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+            loading={loading}
+          />
+        </>
+      );
+    }
+
+    return (
+      <>
+        <h1>About</h1>
+      </>
```
AlanGreene: we should have a loading state here instead of a blank page with just the title

can we show the table with inline loading indicators for the values? or even just a blank table loading state as we have on other pages

src/containers/SideNav/SideNav.js
```diff
@@ -212,6 +212,9 @@ class SideNav extends Component {
           >
             &nbsp;
           </SideNavMenuItem>
+          <SideNavLink element={NavLink} icon={<span />} to={urls.about()}>
+            About
```
AlanGreene: react-intl

src/nls/messages_en.json
```diff
@@ -59,6 +59,7 @@
     "dashboard.deletePipelineRun.heading": "Delete PipelineRun",
     "dashboard.deleteTaskRun.body": "Are you sure you would like to delete TaskRun {name}?",
     "dashboard.deleteTaskRun.heading": "Delete TaskRun",
+    "dashboard.displayVersion.error": "Error Getting Version",
```
AlanGreene: `npm run i18n:extract` to update this to reflect the change you made in code

src/containers/About/About.js
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+function mapStateToProps(state) {
+  return {
+    webSocketConnected: isWebSocketConnected(state)
+  };
+}
+
+const mapDispatchToProps = {
```
AlanGreene: `getInstallProperties` is an API call, not an action that's dispatched through the store

`mapDispatchToProps` can be removed here as it's not needed / used (no calls to `this.props.getInstallProperties()` in the code)

src/containers/About/About.js
CarolynMabbott: Returning an empty string / null from the cluster.go causes an error in the API in the checkstatus which causes non of the fields to render
Had to change it to return a non empty string to stop the unresolved promise error due to check status

AlanGreene: That's definitely unexpected, are you talking about this `checkStatus`? https://github.com/tektoncd/dashboard/blob/134bc627651c9e2d5fd139ae9e3933440e260962/src/api/comms.js#L35

It should only care about the status code, and if the content type is JSON that it's well-formed

CarolynMabbott: I have apparently done a fix to stop that from happening since I last saw it. Cant get it to happen again so it may have been a bad reaction to a return from somewhere else  
Will take out undefined and replace it with empty strings

src/containers/About/About.js
CarolynMabbott: I have changed it to displaying the table headers
Once it gets the info it does update the table and display it

pkg/endpoints/cluster.go
```diff
@@ -133,6 +135,10 @@ func (r Resource) GetEndpoints(request *restful.Request, response *restful.Respo
 
 // GetProperties is used to get the installed namespace only so far
```
a-roberts: Not true now, also what if properties isn't set?

DEVELOPMENT.md
```diff
@@ -255,6 +255,20 @@ Returns HTTP code 500 if an error occurred getting the extensions
 Returns HTTP code 200 and the given extensions in the given namespace if found, otherwise an empty list is returned
 ```
 
+__Dashboard Properties__
+```
+GET /v1/properties
+Get the properties of the tekton-dashboard which are install namespace, dashboard version and tekton pipelines version
```
a-roberts: Nit: 

Gets the properties of the Tekton Dashboard which includes the:

- Installation namespace
- Tekton Dashboard version
- Tekton Pipelines version

Everything else is looking good now 👍

CarolynMabbott: Updated now

CarolynMabbott:
I have done all the intl as much as can be done however as I have a dynamic error message this seems to cause a problem with `npm run i18n:extract` command 

```
SyntaxError: /Users/carolyn.mabbott1ibm.com/go/src/github.com/tektoncd/dashboard/src/containers/About/About.js: [React Intl] Messages must be statically evaluate-able for extraction.
  123 |               subtitle={intl.formatMessage({
  124 |                 id: 'dashboard.displayVersion.errorVersions',
> 125 |                 defaultMessage: this.state.error
      |                                 ^
  126 |               })}
  127 |               lowContrast
```

src/containers/About/About.js
```diff
@@ -0,0 +1,161 @@
...
+          {this.state.error !== '' && (
+            <InlineNotification
+              kind="error"
+              title={intl.formatMessage({
+                id: 'dashboard.displayVersion.error',
+                defaultMessage: 'Error getting version'
+              })}
+              subtitle={intl.formatMessage({
+                id: 'dashboard.displayVersion.errorVersions',
+                defaultMessage: this.state.error
```
AlanGreene: Can't use variables here, it must be a string literal otherwise there's no way to know what value should be extracted to the messages bundle for translation.

If we're generating `this.state.error` ourselves we should use `react-intl` at the point the message is created, and just pass it as a prop here. If it's an API response, we can just display it as-is.

CarolynMabbott: `this.state.error` is dynamically set based upon which ones failed in the setErrorState(errorsFound) function 
The errors message appears like `DashboardVersion,PipelineVersion cannot be found` 
DashboardVersion and PipelineVersion are the properties of the api result to getProperties function that you suggseted further up
{
  installNamespace: 'tekton-pipelines',
  dashbaordVersion: '0.5.0',
    pipelineVersion: '0.10.1'
  }
}

I have tried putting in into intl-react when its created like
```
const errorsFoundList = intl.formatMessage({
      id: 'dashboard.error.error',
      defaultMessage: `${errorsFound} cannot be found`
    });
```

Also tried doing react-intl in 
```
this.setState({
        error: intl.formatMessage({
          id: 'dashboard.error.error',
          defaultMessage: errrosFoundList
        }),
```
When i run the `npm run i18n:extract` it fails with `React Intl] Messages must be statically evaluate-able for extraction.`

AlanGreene: yes that 'Messages must be statically evaluate-able for extraction.' means you can't use a variable as the value of the defaultMessage, however you can use placeholders within the string as we do in other places in the UI.

Taking your first example and updating it to use a placeholder:
```diff
const errorsFoundList = intl.formatMessage({
  id: 'dashboard.error.error',
-  defaultMessage: `${errorsFound} cannot be found`
+  defaultMessage: '{errorsFound} cannot be found'
+}, {
+  errorsFound
});
```

a-roberts:
Also see https://github.com/tektoncd/dashboard/issues/1034

a-roberts:
/lgtm

Spoke with @CarolynMabbott and we can look into augmenting this for openshift-pipelines next (could be as easy as looking in the openshift-pipelines namespace) as well. I've verified the {errorMessage} i18n is working as expected too

a-roberts:
New flaky spotted

```
I0210 11:00:43.194] --- FAIL: TestWebsocketResources (5.10s)
I0210 11:00:43.194]     websocket_test.go:86: Enter TestLogWebsocket...
I0210 11:00:43.194]     websocket_test.go:262: Creating pipelineresource
I0210 11:00:43.194]     websocket_test.go:270: Updating pipelineresource
I0210 11:00:43.194]     websocket_test.go:276: Deleting pipelineresource
I0210 11:00:43.195]     websocket_test.go:293: Creating pipeline
I0210 11:00:43.195]     websocket_test.go:301: Updating pipeline
... skipping 139 lines ...
I0210 11:00:43.239]     websocket_test.go:178: &{TaskRun 5 5 5}
I0210 11:00:43.239]     websocket_test.go:178: &{Extension 5 5 5}
I0210 11:00:43.240]     websocket_test.go:178: &{PipelineResource 5 5 5}
I0210 11:00:43.240]     websocket_test.go:178: &{PipelineRun 5 5 5}
I0210 11:00:43.240]     websocket_test.go:178: &{Task 0 0 0}
I0210 11:00:43.240]     websocket_test.go:189: CUD informer Task creates[0], updates[0] and deletes[0] not equal to expected value: 5
I0210 11:00:43.240] FAIL
I0210 11:00:43.240] FAIL	github.com/tektoncd/dashboard/pkg/endpoints	5.655s
```

Issue #974, opened 2020-01-29
Username in secrets table are displayed encoded


CarolynMabbott:
# Expected behavior
Username is correct 

# Actual behavior
Username is wrong 
Its the encoded name 

# Steps to reproduce the problem

1. Create secret
2. Look at secret on dashboard 
3.

# Environment
- Kubernetes Platform:

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

master 

# Additional Info

a-roberts:
I really don't see this as desirable behaviour - if we wanted it to be obfuscated it would be properly done (encoding/decoding is trivial) so marking this as a bug

a-roberts:
Have a two line fix for this, adding a test now

a-roberts:
Fix merged so closing

PR #988, opened 2020-02-03
Display decoded secret usernames instead of base64 form

a-roberts:
For https://github.com/tektoncd/dashboard/issues/974

Note that I did attempt this in the action but then had massive difficulty in writing a test for it - so do so when displaying the data instead

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/Secrets/Secrets.js
+++ b/src/containers/Secrets/Secrets.js
@@ -226,6 +226,7 @@ export /* istanbul ignore next */ class Secrets extends Component {
         });
       });
       const serviceAccountsString = serviceAccountsWithSecret.join(', ');
+      const secretUsernameToDisplay = atob(secret.username);
       const formattedSecret = {
         annotations: <span title={annotations}>{annotations}</span>,
         id: `${secret.namespace}:${secret.name}`,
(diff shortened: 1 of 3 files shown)


src/containers/Secrets/Secrets.js
```diff
@@ -226,6 +226,8 @@ export /* istanbul ignore next */ class Secrets extends Component {
         });
       });
       const serviceAccountsString = serviceAccountsWithSecret.join(', ');
+      const secretAsBase64 = Buffer.from(secret.username, 'base64');
```
AlanGreene: `Buffer` is not a standard browser type, it'll work in unit tests as it's provided by Node.js but it's unlikely to work in many browsers.

There's a well-supported (IE10+) browser native API for decoding base64 strings: `window.atob`

`const secretUsernameToDisplay = atob(secret.username);`

a-roberts: Excellent, thanks I'll give that a go

a-roberts: Done

a-roberts:
Hmm I'm wondering if this (and also https://github.com/tektoncd/dashboard/issues/990) is actually because *our* knowledge of `secrets` are only those with our annotation (those created with the dashboard). If I print the secrets we use in our `find` function, we only see ones I've made - so I think this is a bigger issue actually.

If this is the case we could make it clearer we only display tekton.dev annotated secret information on ServiceAccounts, and in fact I'd go one step further and make it a `Tekton Secrets` titled page instead of just `Secrets.

@AlanGreene @carlos-logro what are your thoughts?

Issue #981, opened 2020-01-30
Clicking on service account brings up error


CarolynMabbott:
# Expected behavior
I can click on a service account and it tells me info 

# Actual behavior
Get an error 
<img width="1149" alt="Screenshot 2020-01-30 at 11 58 22" src="https://user-images.githubusercontent.com/32342288/73448115-13683180-4358-11ea-8064-69282720a3be.png">

# Steps to reproduce the problem

1. Go to service account page 
2. Click on service account 
3.

# Environment
- Kubernetes Platform:

Openshift read only  

- Tekton Pipelines version:

0.10.0

- Tekton Triggers version:

- Tekton Dashboard version:

master 

# Additional Info

a-roberts:
Yea, reproduced this on our 0.4.1 install too

AlanGreene:
Is this only on specific service accounts or is it affecting all? I'm not seeing it in my dev environment at the moment

a-roberts:
@AlanGreene in my case it's for anything and everything, I'm using OCP 4.2 (happy to share a cluster if you'd like to look into this one closer)

a-roberts:
On OpenShift I've got the following secret patched to the SA default in kube-system:

```
(base) adams-mbp:dashboard aroberts$ k get secret default-dockercfg-2lk5b -o yaml -n kube-system
```
...
```
(base) adams-mbp:dashboard aroberts$ k get secret default-dockercfg-2lk5b -o yaml -n kube-system
apiVersion: v1
data:
  .dockercfg: omitted
kind: Secret
metadata:
  annotations:
    kubernetes.io/service-account.name: default
    kubernetes.io/service-account.uid: b804a234-3c47-11ea-b749-00000a1012bb
    openshift.io/token-secret.name: default-token-q6v49
    openshift.io/token-secret.value: omitted
  creationTimestamp: "2020-01-21T12:24:04Z"
  name: default-dockercfg-2lk5b
  namespace: kube-system
  resourceVersion: "15210"
  selfLink: /api/v1/namespaces/kube-system/secrets/default-dockercfg-2lk5b
  uid: e9cdacc4-3c48-11ea-8b2a-00000a101ddb
type: kubernetes.io/dockercfg
```

notice the type.

If I add hardening and debug:

```
 if (serviceAccount.imagePullSecrets && !loading) {
      rowsForImgPull = serviceAccount.imagePullSecrets.map(({ name }) => {
[...]

PR #989, opened 2020-02-03
[WIP] If type isn't known, don't error for individual SA page

a-roberts:
For https://github.com/tektoncd/dashboard/issues/981

# Changes

This is the "bare minimum" PR that allows the page to load instead of the fatal error we're seeing

I'm going to try and get it displaying the correct type next

Interestingly I don't see any data for the secrets either (regardless of this change):

image

The YAML for this particular ServiceAccount is

```
kind: ServiceAccount
apiVersion: v1
metadata:
  name: default
  namespace: kube-system
  selfLink: /api/v1/namespaces/kube-system/serviceaccounts/default
  uid: b804a234-3c47-11ea-b749-00000a1012bb
  resourceVersion: '11010'
  creationTimestamp: '2020-01-21T12:15:31Z'
secrets:
  - name: default-token-jmjd2
  - name: default-dockercfg-2lk5b
imagePullSecrets:
  - name: default-dockercfg-2lk5b
```

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/ServiceAccount/ServiceAccount.js
+++ b/src/containers/ServiceAccount/ServiceAccount.js
@@ -209,7 +209,9 @@ export /* istanbul ignore next */ class ServiceAccountContainer extends Componen
     let rowsForImgPull = [];
     if (serviceAccount.imagePullSecrets && !loading) {
       rowsForImgPull = serviceAccount.imagePullSecrets.map(({ name }) => {
-        const { type } = secrets.find(secret => secret.name === name);
+        const { type } = secrets.find(secret => secret.name === name) || {
+          type: 'unknown-type'
+        };
         return {
           id: name,
           name,


src/containers/ServiceAccount/ServiceAccount.js
```diff
@@ -209,7 +209,13 @@ export /* istanbul ignore next */ class ServiceAccountContainer extends Componen
     let rowsForImgPull = [];
     if (serviceAccount.imagePullSecrets && !loading) {
       rowsForImgPull = serviceAccount.imagePullSecrets.map(({ name }) => {
-        const { type } = secrets.find(secret => secret.name === name);
+        const foundSecret = secrets.find(secret => secret.name === name);
```
AlanGreene: We could also write this as:
`const { type } = secrets.find(secret => secret.name === name) || { type: 'unknown-type' };`

src/containers/ServiceAccount/ServiceAccount.js
```diff
@@ -209,7 +209,13 @@ export /* istanbul ignore next */ class ServiceAccountContainer extends Componen
...
+        let type = '';
+        if (!foundSecret) {
+          type = 'unknown-type';
+        } else {
+          type = { type };
```
AlanGreene: should this be `({ type } = foundSecret);` to set it to the value of the type field from the found secret instead of an object containing the type?

see previous comment for an alternative that avoids this awkward syntax for destructuring to a previously declared variable

src/containers/ServiceAccount/ServiceAccount.js
a-roberts: Yeah, this is nicer 👍

src/containers/ServiceAccount/ServiceAccount.js
a-roberts: Done, that's so much neater - figuring out the root cause now (and helpfully this branch unlocks Megan to fix the other issue I spotted)

a-roberts:
Hmm I'm wondering if this (and also https://github.com/tektoncd/dashboard/issues/990) is actually because *our* knowledge of `secrets` are only those with our annotation (those created with the dashboard). 

If I print the secrets we use in our `find` function, we only see ones I've made - so I think this is a bigger issue actually.

Multiple options.

Option 1: If this is the case we could make it clearer we only display tekton.dev annotated secret information on ServiceAccounts, and in fact I'd go one step further and make it a `Tekton Secrets` titled page instead of just `Secrets.

Option 2: How about displaying only the names (because that's all we actually know) for the secrets not annotated, in a separate table, and we'd have the ones we do have the data for (Tekton ones) in another? 

Option 3: Or, folks can go to the secrets page for those and we just display the names here?**

@AlanGreene @carlos-logro what are your thoughts? I'm a fan of Option 3 personally.

AlanGreene:
Option 3 sounds fair to me 👍

a-roberts:
PR for option 3 here https://github.com/tektoncd/dashboard/pull/991 so closing this in favour of that

PR #994, opened 2020-02-04
Support Triggers 0.2 features in the Dashboard

### LINKED ISSUE #944: Support Triggers 0.2 features in the dashboard
With the assessment done, under this task go ahead and do it, so specifically we want to:

1) Install Triggers 0.2
2) Read through https://github.com/tektoncd/dashboard/issues/918 and https://github.com/tektoncd/dashboard/issues/937
3) Make the changes and test/contribute

The assessment task (#918) should include all of the changes we'll have to make. If it does turn into something many folks can work on at once it'll be great to add more issues so we can get this done asap 😸 

Obvious things: multiple bindings and interceptors

ncskier:
# Changes

Fixes #944

Updates the EventListener details page with the following:
- Multiple Interceptors
- Multiple TriggerBindings
- No more params
- Display all 4 built-in types of Interceptors (Webhook, GitHub, GitLab,
and CEL)

Here is a screenshot of the updated EventListener component:
Screen Shot 2020-02-04 at 11 39 26 PM

/cc @AlanGreene 
/cc @a-roberts 

I still need to update/add tests and probably clean up the code a little bit. However, any initial feedback would be very helpful 🙂 

# Submitter Checklist

review them:

for more details._


packages/components/src/components/Trigger/Trigger.js
```diff
@@ -37,120 +38,215 @@ const Trigger = ({ intl, eventListenerNamespace, trigger }) => {
...
+                      <code className="bx--snippet--multi interceptor--cel-filter">
+                        {interceptor.cel.filter}
+                      </code>
+                    </>
+                  );
+                } else {
+                  return null;
+                }
+                return (
+                  <AccordionItem key={title} title={`${index + 1}. ${title}`}>
```
AlanGreene: we can't concatenate strings like this as the order may be different in some languages (e.g. RTL languages such as Arabic / Hebrew)

instead we should use a placeholder in the string for the value we're going to substitute

e.g.

```js
intl.formatMessage(
  {
    id: 'dashboard.triggerDetails.celInterceptorTitle',
    defaultMessage: '(CEL) {interceptorName}'
  },
  {
    interceptorName
  }
)
```

src/scss/Triggers.scss
```diff
@@ -31,3 +31,9 @@ limitations under the License.
     margin-top: $spacing-03
   }
 }
+
+.triggers--detail-block {
+  margin-bottom: $layout-01;
+  padding: $layout-01;
+  background-color: #ffffff;
```
AlanGreene: is there a suitable Carbon token we could use here? https://www.carbondesignsystem.com/guidelines/color/usage/
maybe `$ui-02` container background?

ncskier:
Thanks for the review @AlanGreene! I've addressed all of your change requests, so please take another look over the PR when you get the chance 🙂

packages/components/src/components/Trigger/Trigger.stories.js
```diff
@@ -0,0 +1,145 @@
...
+      <IntlProvider locale="en" defaultLocale="en" messages={{}}>
+        <Trigger {...props} />
+      </IntlProvider>
+    </Router>
+  ))
+  .add('no name', () => (
+    <Router history={history}>
+      <IntlProvider locale="en" defaultLocale="en" messages={{}}>
+        <Trigger
+          {...{ ...props, trigger: { ...props.trigger, name: undefined } }}
```
AlanGreene: The extra spread operator `{... }` makes this look odd and could cause confusion for others reading it. Maybe we could split the props to avoid it?

`<Trigger {...props} trigger={{ ...props.trigger, name: undefined }} />`

What do you think?

packages/components/src/components/Trigger/Trigger.js
```diff
@@ -37,120 +38,229 @@ const Trigger = ({ intl, eventListenerNamespace, trigger }) => {
...
-                  id: 'dashboard.trigger.noHeaders',
-                  defaultMessage: 'No headers found for this interceptor.'
-                })}
-                isSortable={false}
-              />
-            </>
-          )}
-        </div>
+      <div className="trigger--interceptors">
+        {trigger.interceptors && trigger.interceptors.length !== 0 && (
```
AlanGreene: do we have / need an empty state for this?

packages/components/src/components/Trigger/Trigger.test.js
```diff
@@ -16,131 +16,234 @@ import Trigger from './Trigger';
...
+    expect(queryByText(/(webhook)/i)).toBeTruthy();
+    expect(queryByText(/webhook-service-name/i)).toBeTruthy();
+    expect(queryByText(/Header/i)).toBeTruthy();
+    expect(queryByText(/webhook-service-namespace/i)).toBeTruthy();
     expect(queryByText(/Wext-Repository-Url/i)).toBeTruthy();
+    expect(queryByText(/https:\/\/github.com\/myorg\/myrepo/i)).toBeTruthy();
     expect(queryByText(/Wext-Incoming-Event/i)).toBeTruthy();
+    expect(queryByText(/wext-incoming-event-value/i)).toBeTruthy();
     expect(queryByText(/Wext-Secret-Name/i)).toBeTruthy();
+    expect(queryByText(/wext-secret-name-value/i)).toBeTruthy();
```
AlanGreene: do we need to explicitly check for all of these or can we simplify the test?

src/containers/EventListener/EventListener.js
```diff
@@ -32,9 +30,11 @@ import {
   getSelectedNamespace,
   isWebSocketConnected
 } from '../../reducers';
-
 import { fetchEventListener } from '../../actions/eventListeners';
 
+import '../../scss/Triggers.scss';
```
AlanGreene: Can we have the EventListener.scss import this instead?

src/containers/EventListener/EventListener.stories.js
```diff
@@ -0,0 +1,192 @@
...
+const props = {
+  fetchEventListener: () => Promise.resolve(eventListener),
+  eventListener,
+  match,
+  intl
+};
+
+storiesOf('EventListenerContainer', module)
+  .add('default', () => (
+    <Router history={history}>
```
AlanGreene: we should use the storybook router instead of using `react-router-dom` directly to avoid a number of issues with routed components in storybook

`import StoryRouter from 'storybook-react-router';`

See https://github.com/tektoncd/dashboard/blob/master/packages/components/src/components/Header/Header.stories.js for example

Also the Container component we use for rendering stories already has an [IntlProvider configured](https://github.com/tektoncd/dashboard/blob/master/.storybook/Container.js#L23) so you shouldn't need to add your own here.

packages/components/src/components/Trigger/Trigger.js
ncskier: I don't think we need an empty state, since a Trigger does not need to have any Interceptors

packages/components/src/components/Trigger/Trigger.test.js
ncskier: The previous test explicitly checked for all of them... but I could simplify it by looping over the Header names & values?

packages/components/src/components/Trigger/Trigger.stories.js
ncskier: Ok, yeah, that looks better 👍

ncskier:
@AlanGreene thanks for the review! I just updated the code to address all of your requests; let me know what you think! 👍

Issue #999, opened 2020-02-04
Created secrets don't display until page refresh


a-roberts:
# Expected behavior
They should!

# Actual behavior
I've seen this reliably with several already created secrets

# Steps to reproduce the problem

1. Repeatedly create any secret in the dashboard

# Environment
- Kubernetes Platform:

Any

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

# Additional info:

Potential clue, we think the secrets modal creates the secret, that's not finishing before the secrets is done rendering (out of sync). So the secrets page fetches before the modal is done.

a-roberts:
@AlanGreene's PR [here](https://github.com/tektoncd/dashboard/pull/1042) has fixed this one well, thanks for the PR Megan although I think Al's is what we want, I'll close that PR too 😸

Issue #998, opened 2020-02-04
Deleting multiple secrets doesn't correctly rerender the page without said secrets


a-roberts:
# Expected behavior
Say they were deleted successfully and not display the deleted secrets

# Actual behavior
Secrets remain in the table

# Steps to reproduce the problem

1. Make several secrets, I used 15
2. Observe the notification they were deleted successfully yet they're still sitting happily in the table
3. Refresh and they're gone

# Environment
- Kubernetes Platform:

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

# Additional Info

a-roberts:
Closing in favour of the work @AlanGreene has done and is merged, tested this works fine now 🎂

Issue #996, opened 2020-02-04
Secret creation results in error notification


Megan-Wright:
# Expected behavior
 When I create a secret I should be able to see the secret added to the table and the success notification "Secret created successfully"

# Actual behavior
When I create a secret, I see this notification: 
![Screenshot 2020-02-04 at 10 51 44](https://user-images.githubusercontent.com/52741262/73738626-b21bd600-473c-11ea-94ab-6985c50e6bee.png)
However, on refreshing the page, the secret has still been created successfully.

# Steps to reproduce the problem

1. Navigate to the secrets panel
2. Create a secret 
3. Error appears

# Environment
- Kubernetes Platform:

OpenShift

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

# Additional Info

a-roberts:
Yeah seeing this too

```

printWarning @ checkPropTypes.js?a15c:20
Secrets.js?510b:230 Uncaught DOMException: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.
    at eval (webpack-internal:///./src/containers/Secrets/Secrets.js:272:39)
    at Array.map (<anonymous>)
    at Secrets.render (webpack-internal:///./src/containers/Secrets/Secrets.js:252:38)
    at finishClassComponent (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:17039:31)
    at updateClassComponent (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:16994:24)
    at beginWork$1 (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:18505:16)
    at HTMLUnknownElement.callCallback (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:347:14)
    at Object.invokeGuardedCallbackDev (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:397:16)
    at invokeGuardedCallback (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:454:31)
    at beginWork$$1 (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:23217:7)
ErrorBoundary.js?9c7c:70 ErrorBoundary Object
```

a-roberts:
We've got the secret created fine in-cluster but if we print the secret object in our map, sometimes it's without the username field (so secret.username) is defined. Working on this now with Megan

[...]

PR #1008, opened 2020-02-05
Secrets table displays all data for all secrets

Megan-Wright:
# Changes

For: https://github.com/tektoncd/dashboard/issues/996, https://github.com/tektoncd/dashboard/issues/998, https://github.com/tektoncd/dashboard/issues/999

- Adds a check to make sure all secret data is populated for a newly created secret before that secret is displayed in the table
- Adds a check on `componentDidUpdate` to check for change in secrets and automatically rerender the page if any are detected, (meaning that the page refreshes on both creation and deletion to reflect these)

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/Secrets/Secrets.js
+++ b/src/containers/Secrets/Secrets.js
@@ -66,16 +66,18 @@ export /* istanbul ignore next */ class Secrets extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { filters, namespace, webSocketConnected } = this.props;
+    const { filters, namespace, webSocketConnected, secrets } = this.props;
     const {
       filters: prevFilters,
       namespace: prevNamespace,
-      webSocketConnected: prevWebSocketConnected
+      webSocketConnected: prevWebSocketConnected,
+      secrets: prevSecrets
     } = prevProps;
 
     if (
       !isEqual(filters, prevFilters) ||
       namespace !== prevNamespace ||
+      !isEqual(secrets, prevSecrets) ||
       (webSocketConnected && prevWebSocketConnected === false)
     ) {
       this.fetchData();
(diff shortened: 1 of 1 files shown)


src/containers/Secrets/Secrets.js
```diff
@@ -66,16 +66,18 @@ export /* istanbul ignore next */ class Secrets extends Component {
...
       namespace: prevNamespace,
-      webSocketConnected: prevWebSocketConnected
+      webSocketConnected: prevWebSocketConnected,
+      secrets: prevSecrets
     } = prevProps;
 
     if (
       !isEqual(filters, prevFilters) ||
       namespace !== prevNamespace ||
+      !isEqual(secrets, prevSecrets) ||
```
AlanGreene: seems a little odd, if the secrets have changed, call the API again...

is the problem that we're only extracting a subset of data from the secret and missing some fields, or are we actually receiving an incomplete object over the websocket or something?

Megan-Wright: the problem is that the `createSecret/deleteSecret`  functions haven't always finished executing before this gets called, and so we were initially receiving fragments of secrets half way through their creation and that was what was being put in the table (or in the delete case they would stay in the table after being deleted). With this change it checks for when the secret is finished creating/deleted  then refreshes to show this in the table.

AlanGreene: Can we not wait for the API calls to complete?

a-roberts: That was our initial attempt, @Megan-Wright can go into more details and provide clarifications - i believe we've got some nice "lifecycle/flow" logs to share, and fetchData's `secrets` weren't *full* in terms of having all the fields (type and username were particularly problematic)

Megan-Wright: @AlanGreene yeah we tried this but the SecretsModal is the one doing `createSecret` and this works completely independently of the Secrets page.... I basically ended up sticking an`await` before every single function within secret creation and the Secrets page would still refresh in the background...

AlanGreene: I think a lot of the problems we're having with secrets stem from the fact that we have inconsistent treatment of them in different parts of the app. We're not storing the full kubernetes resource, instead using a custom structure with a subset of the fields, and we end up with slight differences depending on how the secrets are loaded (i.e. created via the UI, received over websocket, or via API call)

Here's an attempt at storing the full kubernetes resource as we do with other types, eliminating the custom object. https://github.com/AlanGreene/dashboard/commit/8fe96da2ecd2d5372ac97545afd9bac39ac4b98a

I think it would make sense to deliver something like that, and then build on it to resolve the loading issues you're seeing. We really shouldn't need to re-request the secrets in componentDidUpdate and there's a chance it could get called multiple times until the stored secrets stabilise.

Was the problem only reproduced on OpenShift or have you seen it on other platforms too? (e.g. docker desktop). If we handle the errors / incomplete data does it eventually load correctly or does it stay in a partially loaded state?

I'll take a look and see if I can figure out why things aren't behaving.

a-roberts: > I think a lot of the problems we're having with secrets stem from the fact that we have inconsistent treatment of them in different parts of the app. We're not storing the full kubernetes resource, instead using a custom structure with a subset of the fields, and we end up with slight differences depending on how the secrets are loaded (i.e. created via the UI, received over websocket, or via API call)
> 
> Here's an attempt at storing the full kubernetes resource as we do with other types, eliminating the custom object. [AlanGreene@8fe96da](https://github.com/AlanGreene/dashboard/commit/8fe96da2ecd2d5372ac97545afd9bac39ac4b98a)
> 
> I think it would make sense to deliver something like that, and then build on it to resolve the loading issues you're seeing. We really shouldn't need to re-request the secrets in componentDidUpdate and there's a chance it could get called multiple times until the stored secrets stabilise.
> 
> Was the problem only reproduced on OpenShift or have you seen it on other platforms too? (e.g. docker desktop). If we handle the errors / incomplete data does it eventually load correctly or does it stay in a partially loaded state?
> 
> I'll take a look and see if I can figure out why things aren't behaving.

Yeah, big +1 on this, let's look into this as part of the 0.5.1 milestone, we saw this most reliably on OpenShift (the secrets page itself takes several seconds to render in this particular environment), we'll get back to you with regards to Docker Desktop

AlanGreene: PR based on my initial testing mentioned in the comment above: https://github.com/tektoncd/dashboard/pull/1042

This should hopefully simplify things a bit, and make it easier to get to the bottom of any remaining issues.

a-roberts:
Closing in favour of https://github.com/tektoncd/dashboard/pull/1042

Issue #1013, opened 2020-02-05
Notifications errors for the dashboard only appear once


Megan-Wright:
# Expected behavior

If I try to create/delete something in the read-only version of the dashboard, I should get an error message displayed consistently.

# Actual behavior

If I try to create a PipelineRun or Resource etc, then close the error notification, when I press `submit` a second time I get no error displayed in the dashboard.

# Steps to reproduce the problem

1. Easiest way: Install read-only dashboard
2. Try to create a PipelineRun and observe error notification
3. Close the notification and try to create it again, observe no error

# Environment
- Kubernetes Platform:

OpenShift

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

# Additional Info

a-roberts:
Here's the weird thing

![image](https://user-images.githubusercontent.com/12814972/73864750-14a6cc00-483a-11ea-90f7-2e60598aa7c7.png)


so it does enter the action, and it does have the message, so therefore it should be setting errorMessage state which is the controller of the notification being visible

If I augment render() to print the errorMessage prop I see it's always set accurately, but...


```
    return (
      <>
        {createSuccess &&
```

that deleteSuccess check is inside a createSuccess check 👀

a-roberts:
Merged so closing

PR #1016, opened 2020-02-06
Error notifications appear consistently for all create functions

Megan-Wright:
# Changes

For: https://github.com/tektoncd/dashboard/issues/1013

Error notifications now appear every time create functions are triggered, instead of only on the first time.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/CreatePipelineRun/CreatePipelineRun.js
+++ b/src/containers/CreatePipelineRun/CreatePipelineRun.js
@@ -126,6 +126,10 @@ class CreatePipelineRun extends React.Component {
     }
   }
 
+  resetError = () => {
+    this.setState({ submitError: '' });
+  };
+
   getPipelineInfo = (key, { state = this.state, props = this.props } = {}) => {
     if (key === DISABLED) {
       // Disable pipeline info selection when pipelineRef is supplied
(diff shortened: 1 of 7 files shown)


a-roberts:
Yeah, I worked on this with Megan and determined we weren't actually passing clearNotification correctly in the case of using actions, and no rerender was occurring for the PipelineRun/PipelineResource creation pages and so by manipulating state again (doing a reset) causes a proper render again which allows our notification to pop up.

I'll test this further now to check we haven't missed any edge cases, otherwise let's get it in for 0.5.0.

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -105,11 +105,16 @@ const isValidLabel = (type, value) => {
   return regex.test(value);
 };
 
+function resetError() {
```
AlanGreene: Is there a reason this is defined outside the component class?

Normally this would be defined like this:

```js
class CreatePipelineRun extends Component {
  constructor(props) {
    super(props);
    this.state = this.initialState(props);
  }

  resetError = () => {
    this.setState({ submitError: '' });
  }

  ...
}
```

Megan-Wright: ah, nope, I've fixed that now for all of them

a-roberts:
👍  We've got one more change for this PR to do with the success scenario too (so when we create, delete, create, we should get the second create notification), then unless anything's obviously odd I plan to take a look and get it merged for our 0.5.0 release today

Issue #966, opened 2020-01-27
Misc. Front End Development Specs for Dashboard Designs


kimholmes:
This issue should reside under the following Epic:
https://github.com/tektoncd/dashboard/issues/965

- [x] 1. ~~Remove blue background~~ Tracked in https://github.com/tektoncd/dashboard/issues/803
  <details>

  **Expected behavior** - no blue background
  ![NOblueBackground](https://user-images.githubusercontent.com/21316329/73215862-e41caf00-4122-11ea-8b83-bbb37d4662af.png)

  **Actual behavior**
  ![blueBackground](https://user-images.githubusercontent.com/21316329/73216011-3231b280-4123-11ea-946f-73dd7a780c47.png)
  </details>

- [ ] 2. Filter pill appearing when linking to PLRs filtered by Pipeline
  <details>

  **Expected behavior**
  The filter remains empty since the user never engaged with it.

  **Actual behavior**
  A search term gets added to the filter without the user's interaction.
  ![Filter_pipeline](https://user-images.githubusercontent.com/21316329/73216390-eb908800-4123-11ea-8339-b6df8a694a90.png)
  </details>

- [ ] 3. Make tables consistent per design. Remove ellipse action menu and move actions into header per Carbon table component 
  <details>

  **Expected behavior**
  Actions are stored in the header and activated when a row is selected
  ![tableCorrect](https://user-images.githubusercontent.com/21316329/73217212-7e7df200-4125-11ea-8a83-494359ea8cff.png)

  **Actual behavior**
  The table stores actions in an "actions ellipse" at the end of the rows instead of using the Carbon header that utilizes checkboxes and buttons.
  ![table_inconsistent](https://user-images.githubusercontent.com/21316329/73216827-ba648780-4124-11ea-9009-10f4b2965a3f.png)

  ~~The Create buttons on the PipelineRuns and PipelineResources pages should also be moved to the Table header as on the Secrets page.~~ Done
  </details>

- [x] 4. Update Filter design throughout
  <details>

  **Expected behavior**
  ![FilterSpec1](https://user-images.githubusercontent.com/21316329/73757180-d4aaf080-4736-11ea-9747-e1c3df7e0443.png)
  ![FilterSpec2](https://user-images.githubusercontent.com/21316329/73757186-d7a5e100-4736-11ea-8e49-6a7e9a8a6fb6.png)

  **Actual behavior**
[...]

PR #1023, opened 2020-02-06
Update LabelFilter component design

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/966

Move 'Create' buttons to the DataTable header and remove
`additionalButton` prop from LabelFilter.

Also remove style overrides and 'Filters:' label from the tags.

Before:
image

After:
image

There are more changes to come in a later PR, including the addition of a 'clear all' button for the filters.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/LabelFilter/LabelFilter.js
+++ b/packages/components/src/components/LabelFilter/LabelFilter.js
@@ -14,15 +14,12 @@ limitations under the License.
 import React, { Component } from 'react';
 import { injectIntl } from 'react-intl';
 import {
-  Button,
   Form,
   InlineNotification,
   Link,
   Search,
   Tag
 } from 'carbon-components-react';
-
-import { Add16 as Add } from '@carbon/icons-react';
 
 import './LabelFilter.scss';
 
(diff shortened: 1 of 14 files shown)


steveodonovan:
So just hit enter to add the label ? That type of search bar does tend to lend itself to typeahead more than this, seems a little less intuitive

AlanGreene:
@kimholmes see feedback from Steve above

kimholmes:
Thanks. Yes. We've discussed the type ahead feature. Please ignore the design in this mock. Just focus on the type ahead dropdown. A type ahead feature like this one, especially one that can categorize the search term was put on the backburner because of the additional labeling that would need to be done on the backend. But, yes, I agree. A type ahead would be optimal. It's a matter of prioritization of the dev work.
TektonFilter4

AlanGreene:
So should we go ahead with the change in this PR (press enter to add filter) or restore the 'Add filter' button until we have the typeahead functionality?

kimholmes:
The design calls for the removal of the "Add filter" button. I think that's an older pattern and users know to press enter now.

AlanGreene:
@steveodonovan back to you, happy enough with ^^^?

PR #1041, opened 2020-02-11
Refactor About container

AlanGreene:
# Changes

- Remove redux as it wasn't necessary
- Remove duplication for error state
- Remove router + redux from tests as they were not necessary

# Submitter Checklist

review them:

for more details._


AlanGreene:
@a-roberts @CarolynMabbott I've rebased and updated this to reflect the changes in https://github.com/tektoncd/dashboard/pull/1048 with some minor tweaks:

- `IsOpenshift` -> `IsOpenShift` since the platform is called OpenShift
- only display the `IsOpenShift` row when it's true, but include it in the error message if it's missing from the API response

Let me know what you think.

CarolynMabbott:
@AlanGreene Currently running your changes and just noticed that when I run it on Docker that an error comes up `Error getting data IsOpenshift cannot be found` 
Is this intentional behaviour? 
Do we want the error for IsOpenShift to come up all the time if it cannot be found?
What happens if a user is running it on Docker / any other platform thats not OpenShift, this might worry users that the dashboard is not going to work for them

CarolynMabbott:
@AlanGreene Also when I'm running this on Openshift I'm still getting the `Error getting data IsOpenshift cannot be found`
I think `checkMissingProperties` might not be realising that there is a true or false value is there

AlanGreene:
@CarolynMabbott are you running this against the updated back end (changes also in this PR)? It shouldn't be displaying any error or mention of OpenShift on Docker Desktop. I change the case of the property, so if the front and back end code don't match you would get some errors alright:

```diff
-IsOpenshift
+IsOpenShift
```

CarolynMabbott:
@AlanGreene I rebuilt it again and it works perfectly 
The new error message format looks much better than before

Issue #969, opened 2020-01-28
PipelineResources - use a consistent pattern for viewing details and YAML


kimholmes:
# Expected behavior
The PipelineResources should follow the same flow as the other resources in the dashboard.
**These are examples: Main View** (should follow Carbon table spec. Remove ellipse action dropdown on each row and move the actions to the header)
![PLRs_markup](https://user-images.githubusercontent.com/21316329/73276366-90f43c00-41b6-11ea-8107-9cf1199c0050.png)

Details View:
![Details_example](https://user-images.githubusercontent.com/21316329/73277003-90a87080-41b7-11ea-9d0d-1c9181b9a69f.png)

# Actual behavior
There is no way to access the YAML. 
Incorrect usage of the Carbon table
![PLresources_main](https://user-images.githubusercontent.com/21316329/73275650-71a8df00-41b5-11ea-9948-ec956f7a8918.png)
![PLresources_details](https://user-images.githubusercontent.com/21316329/73275656-753c6600-41b5-11ea-8308-c64f65661e27.png)

PR #1054, opened 2020-02-14
Update PipelineResource details view to match common layout

AlanGreene:
# Changes

Add tabs for details / YAML on the PipelineResource details
page to match the common layout being introduced for all
resource types.

https://github.com/tektoncd/dashboard/issues/969

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/PipelineResource/PipelineResource.js
+++ b/src/containers/PipelineResource/PipelineResource.js
@@ -18,9 +18,13 @@ import { connect } from 'react-redux';
 import {
   DataTable,
   DataTableSkeleton,
-  InlineNotification
+  InlineNotification,
+  Tab,
+  Tabs,
+  Tag
 } from 'carbon-components-react';
-import { getErrorMessage } from '@tektoncd/dashboard-utils';
+import { FormattedDate, ViewYAML } from '@tektoncd/dashboard-components';
+import { formatLabels, getErrorMessage } from '@tektoncd/dashboard-utils';
 
 import {
   getPipelineResource,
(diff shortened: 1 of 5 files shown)


src/containers/PipelineResource/PipelineResource.js
```diff
@@ -138,129 +142,199 @@ export /* istanbul ignore next */ class PipelineResourceContainer extends Compon
...
+                  />
+                </p>
+                <p>
+                  <span>
+                    {intl.formatMessage({
+                      id: 'dashboard.metadata.labels',
+                      defaultMessage: 'Labels:'
+                    })}
+                  </span>
+                  {formattedLabelsToRender.length === 0
```
steveodonovan: Is the equals 0 check needed here ?

src/containers/PipelineResource/PipelineResource.js
```diff
@@ -138,129 +142,199 @@ export /* istanbul ignore next */ class PipelineResourceContainer extends Compon
...
-                      </TableHeader>
-                    ))}
-                  </TableRow>
-                </TableHead>
-                <TableBody>
-                  {rows.map(row => (
-                    <TableRow {...getRowProps({ row })}>
-                      {row.cells.map(cell => (
-                        <TableCell key={cell.id}>{cell.value}</TableCell>
+        <Tabs selected={0}>
```
steveodonovan: Just a default value ? What does this prop do ?

src/containers/PipelineResource/PipelineResource.js
AlanGreene: unfortunately yes, otherwise we end up rendering a '0' in the UI

AlanGreene: hrmm I guess not

src/containers/PipelineResource/PipelineResource.js
AlanGreene: selects the first tab by default so its content is rendered on load
after it's mounted Carbon manages the selected tab via internal state and this prop no longer applies

src/containers/PipelineResource/PipelineResource.js
AlanGreene: updated

AlanGreene:
> Didn't do a full comparison of each field but I assume no new values for the translatable strings needed here ?

That's right, using the existing standard strings for the details / YAML tabs to ensure consistency across all types.

Issue #1045, opened 2020-02-12
Use X-Frame-Options header to prevent Clickjacking


mnuttall:
Dashboard should set the X-Frame-Options header in order to prevent "Clickjacking", also known as a "UI redress attack", in which an attacker uses multiple transparent or opaque layers to trick a user into clicking on a button or link on another page when they were intending to click on the top-level page. 

IBM staff can find more details via https://github.ibm.com/IBMCloudPak4Apps/icpa-security/issues/3.

AlanGreene:
This header should only need to be set on the root URL (that serves index.html) since all client side routes are just fragments added to that URL, e.g. `<path_to_root>/#/pipelines`, `<path_to_root>/#/about`

Unless we expect the UI to be embedded in an iframe anywhere we should just set the value to "deny"

PR #1072, opened 2020-02-19
clickjacking fix

CarolynMabbott:
# Changes

Possibly fixes clickjacking 

@AlanGreene Is this the fix you mean on issue #1045 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -215,6 +215,7 @@ func registerWeb(container *restful.Container) {
 			// Static resources are immutable and have a content hash in their URL
 			w.Header().Set("Cache-Control", "public, max-age=31536000, immutable")
 		}
+		w.Header().Set("X-Frame-Options", "deny")
 		fs.ServeHTTP(w, r)
 	}))
 }


AlanGreene:
No this would need to be applied to the Go backend that serves up the index.html file.
There might be another/better place but see here for an example of how we're adding cache headers on the response for the JS / CSS bundles: https://github.com/tektoncd/dashboard/blob/f57c87b939dbf8199a25669c2314e63c1a98eac1/pkg/router/router.go#L195

pkg/router/router.go
```diff
@@ -200,6 +200,7 @@ func registerWeb(container *restful.Container) {
 		if webResourcesStaticPattern.Match([]byte(r.URL.Path)) && !webResourcesStaticExcludePattern.Match([]byte(r.URL.Path)) {
 			// Static resources are immutable and have a content hash in their URL
 			w.Header().Set("Cache-Control", "public, max-age=31536000, immutable")
+			w.Header().Set("X-Frame-Options", "deny")
```
AlanGreene: As mentioned on the working group call this needs to be outside the conditional since the conditional only matches the JS, CSS, and image resources.

Ideally we'd only add the `X-Frame-Options` response header on requests for the landing page (`/index.html` or just `/`), but shouldn't have any negative impact just adding it to everything if that proves awkward.

Issue #932, opened 2020-01-21
Tasks should be sorted in Pipeline spec order not alphabetically


a-roberts:
# Expected behavior
Tasks should be sorted by the order they are defined in the Pipeline, not alphabetically

# Actual behavior
They're sorted alphabetically

# Steps to reproduce the problem

1. Create a PipelineRun with at least three tasks
2. Refresh the page once all tasks complete and see they're ordered alphabetically

# Environment (e.g. Docker Desktop, OpenShift)
All

# Tekton Pipelines version (e.g. 0.9.2, 0.8.0, master)
Any

# Tekton Triggers version (e.g. 0.2, 0.1, master)
Any

# Tekton Dashboard version (e.g. 0.4.0, 0.3.0, master)
All

a-roberts:
@AlanGreene do you think this'd be particularly hard to do? I've noticed we've got a sorting by time already on the TaskRuns page, for full disclosure it's being asked for as a useful feature to have for the upcoming release of the Kabanero project as there's a validate task that actually happens first, but it's displayed at the bottom

AlanGreene:
hrmm displaying them alphabetically (the current mystery behaviour we have no explicit code for...) is definitely not right, but I'm not sure if sorting by timestamp is right either as it won't be consistent across multiple PipelineRuns for the same Pipeline where some Tasks are run in parallel.

Ideally we'd display them in topologically sorted order (walk the DAG essentially), but this would take a bit of work considering the current approach to processing the data.

Maybe we could display them in the order they're defined in the Pipeline definition? That would be relatively simple and we already do it for steps so could take a similar approach.

If we do want to sort by timestamp it would be simple enough to do in either the PipelineRun or TaskTree component assuming the metadata including timestamps is available at those points.

a-roberts:
> hrmm displaying them alphabetically (the current mystery behaviour we have no explicit code for...) is definitely not right, but I'm not sure if sorting by timestamp is right either as it won't be consistent across multiple PipelineRuns for the same Pipeline where some Tasks are run in parallel.
> 
> Ideally we'd display them in topologically sorted order (walk the DAG essentially), but this would take a bit of work considering the current approach to processing the data.
> 
> Maybe we could display them in the order they're defined in the Pipeline definition? That would be relatively simple and we already do it for steps so could take a similar approach.
> 
> If we do want to sort by timestamp it would be simple enough to do in either the PipelineRun or TaskTree component assuming the metadata including timestamps is available at those points.

[...]

PR #1077, opened 2020-02-20
Sort Tasks for PipelineRuns by their Task's earliest startedAt time

a-roberts:
For https://github.com/tektoncd/dashboard/issues/932 which definitely requires a good read through, particularly @AlanGreene's comment:

This approach does satisfy the requirement from Kabanero (in that the `validate` step will appear first as that starts first).

This is also missing unit tests and I want to check this is the approach we want to take, doing things via their order in the Pipeline definition itself proved non-trivial (when we get the data using `getTektonAPI` for example, it's mysteriously sorted by Task name).

More notes:
1) I want to make this code much neater
2) Note that `moment` has been introduced too (MIT license)

3) I'm using `steps[0]` - although there may be multiple steps for a Task.

If one of the steps much sooner than the very first in that array; our sorting will be incorrect so this is a bug. 

Perhaps we advocate for a `startTime` in the Task field for each TaskRun that's not under `running` or `completed`. Alternatively we could implement a `getEarliestStep` method and use that in the sort, sorting against the earliest step for the other tasks.

@AlanGreene what do you think? Perhaps it is easier to display them as defined in the Pipeline but I couldn't figure it out.

### DIFF
--- a/packages/components/src/components/PipelineRun/PipelineRun.js
+++ b/packages/components/src/components/PipelineRun/PipelineRun.js
@@ -62,6 +62,34 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
     };
   };
 
+  sortTaskRuns = taskRuns => {
+    const toReturn = taskRuns.sort((taskRunA, taskRunB) => {
+      if (!taskRunA || !taskRunB) {
+        return 0;
+      }
+
+      const firstStepA = taskRunA.steps[0];
+      const firstStepB = taskRunB.steps[0];
+
+      if (!firstStepA || !firstStepB) {
+        // shouldn't be able to reach this state (both tasks requiring at least one step)
+        return 0;
+      }
+
+      const { startedAt: startedAtA } =
+        firstStepA.terminated || firstStepA.running || {};
+      const { startedAt: startedAtB } =
+        firstStepB.terminated || firstStepB.running || {};
+
+      if (!startedAtA || !startedAtB) {
+        return 0;
+      }
+
+      return new Date(startedAtA).getTime() - new Date(startedAtB).getTime();
+    });
+    return toReturn;
+  };
+
   loadTaskRuns = pipelineRun => {
     if (!pipelineRun || !pipelineRun.status || !pipelineRun.status.taskRuns) {
       return [];
@@ -249,6 +277,8 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
       );
     }
     const taskRuns = this.loadTaskRuns(pipelineRun, taskRunNames);
+
+    this.sortTaskRuns(taskRuns);
 
     if (taskRuns.length === 0) {
       return (
--- a/testdata/ordering/pipeline.yaml
+++ b/testdata/ordering/pipeline.yaml
@@ -1,0 +1,20 @@
+---
+apiVersion: tekton.dev/v1alpha1
+kind: Pipeline
+metadata:
+  name: test-pipeline
+  namespace: tekton-pipelines
+spec:
+  tasks:
+    - name: b-task
+      taskRef:
+        kind: Task
+        name: step-b-task
+    - name: c-task
+      taskRef:
+        kind: Task
+        name: step-c-task
+    - name: a-task
+      taskRef:
+        kind: Task
+        name: step-a-task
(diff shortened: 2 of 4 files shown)


a-roberts:
@AlanGreene updated this PR to include finding the earliest startedAt times for steps, to be used for ordering the tasks that are displayed, would be useful to discuss this one before proceeding I think because of the concerns you raised (I personally think it's a good starting point though, perhaps the implementation can be tidied/optimised?)

Once we're happy with this approach I can work on the unit tests - so far I'm applying the PIpeline/PipelineRun/Task from `testdata/ordering`, and then looking at all of the `startedAt` fields for the steps in the tasks.

AlanGreene:
Can we not use the `taskRun.status.startTime`?

Also instead of introducing moment we should be able to get time in milliseconds by `new Date(startTime).getTime()` and compare using that.

a-roberts:
> Can we not use the `taskRun.status.startTime`?
> 
> Also instead of introducing moment we should be able to get time in milliseconds by `new Date(startTime).getTime()` and compare using that.
> 
> I'll try make some time this evening to look at the definition ordering to see if that'll be any easier.

Ah yeah, that does look better than startedAt vs terminatedAt, will give it a go 😅 thanks, yep lemme look at taking out moment too

a-roberts:
Refactoring done, ready for another look please 😸

a-roberts:
Had plenty of lint errors so fixed and pushed 🤞

package.json
```diff
@@ -41,6 +41,7 @@
     "lodash.keyby": "^4.6.0",
     "lodash.merge": "^4.6.2",
     "lodash.snakecase": "^4.1.1",
+    "moment": "^2.24.0",
```
AlanGreene: no longer needed, same for the package-lock change

packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -133,6 +133,7 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
           pod: taskRun.status.podName,
           reason,
           steps,
+          status: taskRun.status,
```
AlanGreene: This isn't used anymore right? Should remove it to avoid potential confusion

packages/components/src/components/PipelineRun/PipelineRun.js
```diff
@@ -248,7 +249,46 @@ export /* istanbul ignore next */ class PipelineRunContainer extends Component {
...
+            const startedAtB = new Date(
+              foundEarliestTaskBStep.running.startedAt
+            ).getTime();
+            return startedAtA - startedAtB;
+          }
+        }
+        return 0;
+      }
+      return 0;
+    });
```
AlanGreene: Can we extract this to a function to keep `render` a bit more manageable? It can live in the same file, outside the component.

AlanGreene: I'm also not 100% sure when it should sort vs. leaving the order unchanged. The code here will only sort them when both steps are in the same state (i.e. both terminated, or both running). It leaves them as they are (`return 0`) in the case that either of them are in the 'waiting' state, or if they're in different states. Is this desired?

Early returns could be used to reduce some of the indentation here and (IMO) make it easier to read.

Depending on exactly when we want to sort the tasks, something like this might work:

```js
taskRuns = taskRuns.sort((taskRunA, taskRunB) => {
  if (!taskRunA || !taskRunB) {
    return 0;
  }

  const firstStepA = taskRunA.steps[0];
  const firstStepB = taskRunB.steps[0];

  if (!firstStepA || !firstStepB) {
    // can we actually reach this state?
    return 0;
  }

  const { startedAt: startedAtA } = firstStepA.terminated || firstStepA.running || {};
  const { startedAt: startedAtB } = firstStepB.terminated || firstStepB.running || {};

  if (!startedAtA || !startedAtB) {
    return 0;
  }

  return new Date(startedAtA).getTime() - new Date(startedAtB).getTime();
});
```

a-roberts: Yeah this looks so much neater, thanks, so for

```
The code here will only sort them when both steps are in the same state (i.e. both terminated, or both running). It leaves them as they are (return 0) in the case that either of them are in the 'waiting' state, or if they're in different states. Is this desired?
```

you're right, I'm willing to live with this though - and the sorting happens often, on clicking the tasks or steps for example. I figured this would cause undesired behaviour (e.g. clicking on what you think is Step B might end up being Step A, but actually things are fine), I've addressed your other comments too 👀

Issue #1078, opened 2020-02-24
Make pipelineRun batch actions configurable


steveodonovan:
# Expected behavior

In the pipelineRuns component you should be able to pass batch actions to the Table component

# Actual behavior

Its is not possible at the moment limiting use of the batch actions.

# Environment
- Kubernetes Platform:

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

# Additional Info

PR #1079, opened 2020-02-24
Add multiple action parameters to pipelineRuns

steveodonovan:
https://github.com/tektoncd/dashboard/issues/1078

/cc @AlanGreene 

# Changes

Allows more configuration around the pipelineruns component. Specifically allows a consumer to pass the batch actions

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineRuns/PipelineRuns.js
+++ b/packages/components/src/components/PipelineRuns/PipelineRuns.js
@@ -22,6 +22,7 @@ import { FormattedDate, FormattedDuration, RunDropdown } from '..';
 import './PipelineRuns.scss';
 
 const PipelineRuns = ({
+  batchActionButtons = [],
   createPipelineRunURL = urls.pipelineRuns.byName,
   createPipelineRunDisplayName = ({ pipelineRunMetadata }) =>
     pipelineRunMetadata.name,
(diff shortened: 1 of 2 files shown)


packages/components/src/components/PipelineRuns/PipelineRuns.js
```diff
@@ -50,6 +51,8 @@ const PipelineRuns = ({
   },
   getPipelineRunCreatedTime = pipelineRun =>
     pipelineRun.metadata.creationTimestamp,
+  getPipelineRunId = pipelineRun =>
+    `${pipelineRun.metadata.namespace}:${pipelineRun.metadata.name}`,
```
AlanGreene: since we're changing this anyway can we just make it `pipelineRun.metadata.uid`?

steveodonovan: I was going to but for the sake of our limbo state I decided against. uid is generated by k8s after the resource has been applied. namespace and name are expected even for pipelineRuns that have not been. In practice it should never break the dashboard but this is the safer option imo

AlanGreene: fair enough

Issue #1063, opened 2020-02-17
Tekton Pipelines beta support


a-roberts:
There are changes at least [here](https://github.com/tektoncd/pipeline/pull/2035) for Tekton Pipelines itself to move up to beta, it'd be useful for the Tekton Dashboard to work with any features exclusive to the beta and to not break. 

Have a look at what needs to be done (so I'd build against master, check everything continues to work fine, and then build Pipelines using the referenced commit). It could be that there are no breaking changes.

AlanGreene:
We'll need to update `getTektonAPI` in src/api/index.js to accept the version, defaulting to `v1alpha2` if none provided and only pass the override when required (e.g. PipelineResource APIs, maybe triggers?)

https://github.com/tektoncd/dashboard/blob/24eb823bfd4472d3fb1239dcf811d89b92f3f0a8/src/api/index.js#L91

anneqm:
on beta community call, there was discussion about move to v1alpha2 or maybe to v1beta1-rc1. 
See https://docs.google.com/document/d/1JmRd0vbh2Q6Emu2DOk-LwLHfxd2V0zrN4WNawGzoNXw/edit#

anneqm:
@a-roberts do we need a similar issue for the move up to pipelines 0.11 ? That's targeted for Feb 26, with beta then in March (March 13 as of last week beta call).

a-roberts:
> @a-roberts do we need a similar issue for the move up to pipelines 0.11 ? That's targeted for Feb 26, with beta then in March (March 13 as of last week beta call).

Yep I think that's a good idea

a-roberts:
I'll pick this up, will try out @vdemeester's super PR shortly

a-roberts:
Upgraded my Docker Desktop version to pick up Kube 1.15.x, built and deployed perfectly with Vincent's latest commit

Checked out master of Dashboard and did a clean build, kicked the tires. Testing to do (comment will be updated with results):

1) Import resources (creating a PIpelineRun, then importing known resources in the PIpeline Hotel)
2) Creating and deleting Tekton secrets
3) Testing manual PipelineRun creation
4) Checking things render OK for Tekton resources (Tasks, Pipelines, PipelineRuns, PIpelineResources)
5) About page
Not 6) Webhooks extension - not testing this just yet

Then, repeat with the change for v1alpha2 in api/index.js:

1) Import resources (creating a PIpelineRun, then importing known resources in the PIpeline Hotel)
[...]

PR #1089, opened 2020-02-26
[WIP] Update to use latest Pipelines Beta code

a-roberts:
https://github.com/tektoncd/dashboard/issues/1063

Tested I can get and create resources fine (including import, PipelineRuns, reruns), and viewing Trigger resources

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -6,10 +6,10 @@ required = [
 
 [[constraint]]
   name = "github.com/tektoncd/pipeline"
-  revision = "558b37edfffa1d8a33428fb9bd50a8ab425e3e83"
+  revision = "v0.11.0-rc2"
 
 [[override]]
-  name = "github.com/knative/pkg"
+  name = "knative.dev/pkg"
   branch = "release-0.12"
 
 [[override]]
@@ -22,7 +22,7 @@ required = [
 
 [[override]]
   name = "k8s.io/api"
-    version = "v0.16.5"
+  version = "v0.16.5"
 
 [[override]]
   name = "k8s.io/code-generator"
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -88,11 +88,16 @@ export function getResourcesAPI(
   ].join('');
 }
 
-export function getTektonAPI(type, { name = '', namespace } = {}, queryParams) {
-  return getResourcesAPI(
-    { group: 'tekton.dev', version: 'v1alpha1', type, name, namespace },
+export function getTektonAPI(
+  type,
+  { name = '', namespace, version = 'v1beta1' } = {},
+  queryParams
+) {
+  const resourceAPI = getResourcesAPI(
+    { group: 'tekton.dev', type, version, name, namespace },
     queryParams
   );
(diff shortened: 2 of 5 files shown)


a-roberts:
Adding @AlanGreene and @vdemeester as reviewers to keep an eye on this one 😸 (the code could be cleaner, but this works - see description for testing and obvious todo item)

src/api/index.js
```diff
@@ -88,11 +93,22 @@ export function getResourcesAPI(
...
+  version
+) {
+  let versionToUse = 'v1beta1';
+  if (version) {
+    versionToUse = version;
+  }
+  const resourceAPI = getResourcesAPI(
+    { group: 'tekton.dev', type, name, namespace },
+    queryParams,
+    versionToUse
```
```suggestion
export function getTektonAPI(
  type,
  { name = '', namespace, version = 'v1beta1' } = {},
  queryParams
) {
  const resourceAPI = getResourcesAPI(
    { group: 'tekton.dev', type, version, name, namespace },
    queryParams
```

we can set the default version in the first param and avoid the extra juggling here
then just pass it down to `getResourcesAPI` which would remain unchanged (as-is on master)

a-roberts:
Revisited this today as there's an rc cut for PIpelines beta, everything working ok apart from the version detection which is in https://github.com/tektoncd/dashboard/pull/1104

Gopkg.toml
```diff
@@ -6,10 +6,10 @@ required = [
 
 [[constraint]]
   name = "github.com/tektoncd/pipeline"
-  revision = "558b37edfffa1d8a33428fb9bd50a8ab425e3e83"
+  revision = "60d76bd5526e815cdbf5639ff3fa5b9eb09ce321"
```
vdemeester: why not using version now ? `version = "v0.11.0-rc1"`

a-roberts: Fixed 🤗

a-roberts:
Looking pretty nice now I reckon and tests are green, for awareness (this shouldn't be news to @AlanGreene and @vdemeester from our working group calls, lemme know if it is) the plan is to do a 0.5.3 first (our last "non-beta" Dashboard release) and then a 0.6.0 for beta and that's when this pretty small but important PR can go in

anneqm:
@a-roberts I prefer the 0.5.3 first so we have options for an update for Kabanero before the move to beta.

a-roberts:
Modified this to use a new tekton-beta branch, my plan is to, once the 0.5.3 release is done today, get this published somewhere for folks to try (potentially @afrittoli in the robocat cluster), I'll rebase if necessary and pick up the latest published RC too

I'm unavailable after this Thursday but I'd like to...
1) have the tekton-beta branch done by then to the best of our ability and published
2) have a working release by then that folks can play with (referencing builds of the above ^^)
3) then once we know that's all great and tested and beta goes live, we can merge this into master and update our docs

a-roberts:
I'm gonna merge this into pipelines-beta so we have a stream we can deliver any beta specific fixes into, and then get folks to review this for master when we're ready (this is instead of it sitting at my fork)

Issue #1094, opened 2020-02-27
Include installed Triggers version for About page


a-roberts:
Info at https://github.com/tektoncd/dashboard/issues/1087 - detect 0.1, 0.2.0, 0.2.1, 0.3.0 and master (e.g. `development`)

AlanGreene:
Merged https://github.com/tektoncd/dashboard/pull/1212

PR #1104, opened 2020-03-03
Add triggers to about page

CarolynMabbott:
References issue #1094 
# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/endpoints/cluster.go
+++ b/pkg/endpoints/cluster.go
@@ -32,6 +32,7 @@ type Properties struct {
 	InstallNamespace string
 	DashboardVersion string
 	PipelineVersion  string
+	TriggersVersion  string
 	IsOpenShift      bool
 	ReadOnly         bool
 }
@@ -170,6 +171,13 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
 	isOpenShift := IsOpenShift(r, installedNamespace)
 	isReadOnly := IsReadOnly()
 	pipelineVersion := GetPipelineVersion(r, isOpenShift)
+	isTriggersInstalled := IsTriggersInstalled(r, isOpenShift)
+	triggersVersion := ""
+	if isTriggersInstalled == false {
+		triggersVersion = "Not Installed"
+	} else {
+		triggersVersion = GetTriggersVersion(r, isOpenShift)
+	}
 
 	properties := Properties{
 		InstallNamespace: os.Getenv("INSTALLED_NAMESPACE"),
@@ -177,6 +185,7 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
 		PipelineVersion:  pipelineVersion,
 		IsOpenShift:      isOpenShift,
 		ReadOnly:         isReadOnly,
+		TriggersVersion:  triggersVersion,
 	}
 	logging.Log.Debugf("Writing install properties: %s", properties)
 	response.WriteEntity(properties)
--- a/pkg/endpoints/dashboard.go
+++ b/pkg/endpoints/dashboard.go
@@ -55,30 +55,6 @@ func GetDashboardVersion(r Resource, installedNamespace string) string {
 	return version
 }
 
-// Get pipelines version
-func GetPipelineVersion(r Resource, isOpenShift bool) string {
-	version := ""
-	namespacesToCheck := []string{"tekton-pipelines"}
-
-	if isOpenShift {
-		namespacesToCheck = append(namespacesToCheck, "openshift-pipelines")
-	}
-
-	// Go through possible namespaces Tekton Pipelines is installed in
-	// For each namespace if the version is empty then it goes through to get the deployments within that namespace to try and locate the Tekton Pipelines deployment to get the version
-	for _, versionAttempt := range namespacesToCheck {
-		if version == "" {
-			version = getPipelineDeployments(r, versionAttempt)
-		}
-	}
-
-	if version == "" {
-		logging.Log.Error("Error getting the Tekton Pipelines deployment version. Version is unknown")
-		return ""
-	}
-	return version
-}
-
 // IsReadOnly determines whether the Dashboard is running in read-only mode or not
 func IsReadOnly() bool {
 	asBool, err := strconv.ParseBool(os.Getenv("READ_ONLY"))
@@ -91,6 +67,7 @@ func IsReadOnly() bool {
 }
 
 // IsOpenShift determines whether the Dashboard is running on OpenShift or not
+// Get whether running on openshift or not
 func IsOpenShift(r Resource, installedNamespace string) bool {
 	namespaces, err := r.K8sClient.CoreV1().Namespaces().List(v1.ListOptions{})
 	if err != nil {
@@ -105,10 +82,10 @@ func IsOpenShift(r Resource, installedNamespace string) bool {
 		namespaceName := namespace.GetName()
 
 		if namespaceName == "openshift-pipelines" {
-			openshiftPipelineFound = SearchForPipelineDeployments(r, "openshift-pipelines")
+			openshiftPipelineFound = searchForDeployment(r, "pipelines", "openshift-pipelines")
 		}
 		if namespaceName == "tekton-pipelines" {
-			tektonPipelinesFound = SearchForPipelineDeployments(r, "tekton-pipelines")
+			tektonPipelinesFound = searchForDeployment(r, "pipelines", "tekton-pipelines")
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


pkg/endpoints/cluster.go
```diff
@@ -32,7 +32,9 @@ type Properties struct {
 	InstallNamespace string
 	DashboardVersion string
 	PipelineVersion  string
-	IsOpenShift      bool
+	TriggersVersion  string
+	// IsTriggersInstalled bool
```
a-roberts: Can go

pkg/endpoints/dashboard.go
```diff
@@ -186,3 +186,118 @@ func getPipelineDeployments(r Resource, namespace string) string {
...
+	// Go through possible namespaces Tekton Triggers is installed in
+	// For each namespace if the version is empty then it goes through to get the deployments within that namespace to try and locate the Tekton Triggers deployment to get the version
+	for _, versionAttempt := range namespacesToCheck {
+		if version == "" {
+			version = getTriggersDeployments(r, versionAttempt)
+		}
+	}
+
+	if version == "" {
+		logging.Log.Error("Error getting the Tekton Trigggers deployment version. Version is unknown")
```
a-roberts: gggggg

pkg/endpoints/dashboard.go
```diff
@@ -186,3 +186,118 @@ func getPipelineDeployments(r Resource, namespace string) string {
 
 	return version
 }
+
+// Get triggers version
+func GetTriggersVersion(r Resource, isOpenShift bool) string {
```
a-roberts: Could we reduce duplicate code in this file? GetDeploymentVersion perhaps and you accept triggers or pipelines, then just change the version = line from getTriggersDeployments if it's Triggers, getPipelineDeployments if it's Pipelines?

pkg/endpoints/dashboard.go
```diff
@@ -186,3 +186,118 @@ func getPipelineDeployments(r Resource, namespace string) string {
...
+	}
+
+	if version == "" {
+		logging.Log.Error("Error getting the Tekton Trigggers deployment version. Version is unknown")
+		return ""
+	}
+	return version
+}
+
+// Go through pipeline deployments and find what version it is
```
a-roberts: Triggers, and see comment above about reducing duplicate code

pkg/endpoints/dashboard.go
```diff
@@ -186,3 +186,118 @@ func getPipelineDeployments(r Resource, namespace string) string {
...
+	deployments, err := r.K8sClient.AppsV1().Deployments(namespace).List(listOptions)
+	if err != nil {
+		logging.Log.Errorf("Error getting the Tekton Triggers deployment: %s", err.Error())
+		return ""
+	}
+
+	for _, deployment := range deployments.Items {
+		deploymentLabels := deployment.GetLabels()
+
+		logging.Log.Debug("triggers labels are %s ", deploymentLabels)
```
a-roberts: Can go, and isn't debugf'd if we did want it

pkg/endpoints/dashboard.go
```diff
@@ -186,3 +186,118 @@ func getPipelineDeployments(r Resource, namespace string) string {
...
+		if version == "" {
+			deploymentLabels := deployment.Spec.Template.GetLabels()
+			version = deploymentLabels["triggers.tekton.dev/release"]
+		}
+		if version == "" {
+			deploymentLabels := deployment.Spec.Template.GetLabels()
+			version = deploymentLabels["version"]
+		}
+
+		// For openshift pipelines operator triggers is at 0.2.1
```
a-roberts: Confusing comment

pkg/endpoints/dashboard.go
```diff
@@ -186,3 +186,118 @@ func getPipelineDeployments(r Resource, namespace string) string {
...
+
+	deployments, err := r.K8sClient.AppsV1().Deployments(namespace).List(listOptions)
+	if err != nil {
+		logging.Log.Errorf("Error getting the Tekton Triggers deployment: %s", err.Error())
+		return false
+	}
+
+	for _, deployment := range deployments.Items {
+		isTriggersInstalled = true
+		logging.Log.Debug("triggers deployment is %s", deployment)
```
a-roberts: Can go

src/containers/About/About.js
```diff
@@ -118,7 +119,7 @@ export /* istanbul ignore next */ class About extends Component {
     if (dashboardInfo && !loading) {
       propertiesToCheck.forEach(property => {
         const value = this.getDisplayValue(dashboardInfo[property]);
-        if (value) {
+        if (value && value !== 'Not Installed') {
```
AlanGreene: can we send an empty string / null back instead?

CarolynMabbott: An empty string implies that triggers is installed 
Which brings up an error saying that it cant find triggers version which I wanted to avoid if a user doesn't have triggers installed

AlanGreene: I'm not sure I follow, why would an empty string imply that it's installed? Are you referring to the current implementation or a user's expectations?

The API should not have to return a magic string to satisfy the current implementation in the front end, instead we should design the API in a way that makes sense (either returning null / empty string, or omitting the field entirely) and update the front end code accordingly.

pkg/endpoints/dashboard.go
CarolynMabbott: I've reduced the duplicate code now 
Combined the TriggersDeployments and GetDeployments and version stuff

src/containers/About/About.test.js
```diff
@@ -90,3 +95,22 @@ describe('About', () => {
     ).toBeTruthy();
   });
 });
+
+it('should not display TiggersVersion when value is Not Installed', async () => {
```
a-roberts: Not Installed is now gone so needs updating I assume

pkg/endpoints/dashboard.go
a-roberts: Still too many g's

src/containers/About/About.js
```diff
@@ -83,6 +83,11 @@ export /* istanbul ignore next */ class About extends Component {
   async fetchDashboardInfo() {
     try {
       const dashboardInfo = await getInstallProperties();
+      if (
+        Object.prototype.hasOwnProperty.call(dashboardInfo, 'TriggersVersion')
```
a-roberts: So if dashboardInfo.contains TriggersVersion basically?

src/containers/About/About.js
```diff
@@ -83,6 +83,11 @@ export /* istanbul ignore next */ class About extends Component {
   async fetchDashboardInfo() {
     try {
       const dashboardInfo = await getInstallProperties();
+      if (
```
AlanGreene: if we add other 'optional' properties in future this code will quickly get out of hand

what about the approach I shared on Slack last week? It would allow us to add new properties with minimal effort, and differentiate between those that are 'required' (i.e. should trigger an error notification) vs. optional.

image

pkg/endpoints/dashboard.go
```diff
@@ -130,72 +106,173 @@ func IsOpenShift(r Resource, installedNamespace string) bool {
...
+
+			// For Tekton Pipelines 0.9.0 - 0.9.2
+			if version == "" {
+				deploymentImage := deployment.Spec.Template.Spec.Containers[0].Image
+				if strings.Contains(deploymentImage, "pipeline/cmd/controller") && strings.Contains(deploymentImage, ":") && strings.Contains(deploymentImage, "@") {
+					s := strings.SplitAfter(deploymentImage, ":")
+					if strings.Contains(s[1], "@") {
+						t := strings.Split(s[1], "@")
+						version = t[0]
+					}
```
AlanGreene: @a-roberts what pipeline version do we need to support for our next dashboard patch release (0.5.3)? 0.5.0 supports 0.10, are we expecting to still support older versions?

pkg/endpoints/dashboard.go
```diff
@@ -130,72 +106,173 @@ func IsOpenShift(r Resource, installedNamespace string) bool {
...
+
+	if version == "" {
+		logging.Log.Error("Error getting the Tekton Trigggers deployment version. Version is unknown")
+		return ""
+	}
+	return version
+}
+
+// Check whether Tekton Triggers is installed
+func IsTriggersInstalled(r Resource, isOpenShift bool) bool {
```
AlanGreene: Is this needed? Can we not try to get the version, and if not found assume it's not installed?

pkg/endpoints/dashboard.go
a-roberts: Yeah, so the approach here should be supplementary for detecting versions and not break anything we used to pick up

src/containers/About/About.test.js
CarolynMabbott: yep updated now

src/containers/About/About.js
CarolynMabbott: Yep

pkg/endpoints/dashboard.go
CarolynMabbott: Yes it is needed
If the version is not found that could be because of multiple reasons including errors in the dashboard as well as Triggers changing their versioning code 
Could also indicate that a user is using a very old version of triggers if an error with version of triggers is displayed

src/containers/About/About.js
CarolynMabbott: I did try that last week after you sent it via Slack but I couldn't get it to work 

If a optional property is installed and does somehow return an empty string as the version then that should probably be displayed as it might indicate an error somewhere in dashboard
The two bottom properties should be required anyway shouldn't they? As they are true or false and the ReadOnly impacts what gets displayed

CarolynMabbott:
@a-roberts @AlanGreene 
I have pushed some changes to fix an error where TriggersVersion kept getting added everytime the About page was opened but to get the properties to update dynamically I have to either reset propertiesToCheck each time or I could have a function each time the page is opened to ensure that propertiesToCheck doesn't contain properties that have been removed (like deleting triggers)

Could I get your opinions on how best to do this?
I have just reset propertiesToCheck for now but what is the best way to do this so it doesnt display errors?

a-roberts:
> @a-roberts @AlanGreene
> I have pushed some changes to fix an error where TriggersVersion kept getting added everytime the About page was opened but to get the properties to update dynamically I have to either reset propertiesToCheck each time or I could have a function each time the page is opened to ensure that propertiesToCheck doesn't contain properties that have been removed (like deleting triggers)
> 
> Could I get your opinions on how best to do this?
> I have just reset propertiesToCheck for now but what is the best way to do this so it doesnt display errors?

I think the optional approach is pretty nice, with regards to surfacing errors - if it's an optional then we could display a warning dialog instead, which Carbon does support anyway https://www.carbondesignsystem.com/components/notification/code/

a-roberts:
@CarolynMabbott let's split this PR as well please, it'd be more useful to have one for getting the new Tekton Pipelines version and one for this specifically which is requiring more work, I'd like to at least get the Pipelines version detection for beta, in our Dashboard release candidate. You can keep everything you've done here apart from the Trigger version handling.

CarolynMabbott:
@AlanGreene I tried to add the string we talked about of `No Version Found` however this meant I had to dynamically change the properties to check in about.js which was very unreliable and didnt work when triggers was uninstalled

AlanGreene:
Taking another look at this now

AlanGreene:
This is what I was thinking: https://github.com/AlanGreene/dashboard/commit/2e7712ca58c85f6834f0d540ad99eb78741fd925

Based on https://github.com/tektoncd/dashboard/commit/fa128ccaa4285ad25f270a6585abb1473b632827 and https://github.com/tektoncd/dashboard/pull/1104#discussion_r389838675

* If Triggers is installed it will display either the found version or 'Unknown'.
* If Triggers is not installed it is ignored.
* If ReadOnly or OpenShift are true, we display the corresponding row. I don't think there's value in telling someone who's not running on OpenShift that they're not on OpenShift...

a-roberts:
Yep, so do wanna say thanks to @CarolynMabbott for working on this, thanks Alan for getting the final touches in and all of the review comments. We'd really like this in for beta so in the interest of time and Carolyn not being here atm...

/lgtm
/approve
/meow hats

🐈

a-roberts:
Closed in favour of https://github.com/tektoncd/dashboard/pull/1212 finishing this one off, thanks @CarolynMabbott for all of your work on this 🐱

Issue #1095, opened 2020-02-27
Display new Triggers 0.3 features


a-roberts:
See https://github.com/tektoncd/dashboard/issues/1087 for details

# Expected behavior
@ncskier mentions a few new additions (detailed above) so render these in the dashboard under the appropriate page

# Steps to reproduce the problem

1. Install Triggers 0.3 and the latest dashboard
2. Use the latest Triggers 0.3 features (https://github.com/tektoncd/triggers/pull/358 for ClusterTriggerBindings, https://github.com/tektoncd/triggers/pull/329 for overlays)
3. Observe no details are displayed

# Environment
- Kubernetes Platform:
Any

- Tekton Pipelines version:
Any

- Tekton Triggers version:
0.3.0

- Tekton Dashboard version:
master

a-roberts:
ClusterTriggerBindings are in so only the overlays work to go

Megan-Wright:
Overlays PR merged now so closing
/close

PR #1112, opened 2020-03-04
Display overlays for CEL-interceptors

Megan-Wright:
# Changes
For: https://github.com/tektoncd/dashboard/issues/1095

The EventListener view can now display overlays for CEL-interceptors, which was added in Tekton Triggers 0.3.0.

Screenshot 2020-03-04 at 09 47 57

# Submitter Checklist

review them:

for more details._


Megan-Wright:
@AlanGreene this is the best way that I could find of adding the overlays elements to the display output, but was wondering if you knew a better way?

packages/components/src/components/Trigger/Trigger.test.js
```diff
@@ -90,7 +90,13 @@ const fakeTrigger = {
     },
     {
       cel: {
-        filter: 'cel-filter'
+        filter: 'cel-filter',
+        overlays: [
```
AlanGreene: the tests should verify this new content is handled correctly, same for the others

packages/components/src/components/Trigger/Trigger.js
```diff
@@ -223,6 +223,15 @@ const Trigger = ({ intl, eventListenerNamespace, trigger }) => {
...
                 interceptorName = interceptor.cel.filter;
+                const overlays = [];
+                interceptor.cel.overlays.forEach(overlay => {
+                  overlays.push(
+                    <>
+                      <p> expression: {overlay.expression} </p>
+                      <p> key: {overlay.key} </p>
+                    </>
+                  );
+                });
```
AlanGreene: Items in an array must have a `key` prop, you should see a warning related to this in the browser console when in dev mode. To use this with a fragment we need to use the explicit `<React.Fragment>` instead of the shorthand `<>`

Prefer `.map` instead of `.forEach + .push`.

```js
const overlays = interceptor.cel.overlays.map(overlay => (
  <React.Fragment key={overlay}>
    <p>{expression}</p>
    <p>{key}</p>
  </React.Fragment>
));
```

you'll also need to use react-intl for the labels


I wonder if this is the best presentation of this data, it seems more like a list of some sort (a definition list or table). When there are multiple overlays the current presentation isn't particularly clear at expressing the relationship between the key and value. A structured approach like a table (Carbon `StructuredList` or `DataTable`) makes it more obvious that they're related / part of the same thing as each overlay is captured in a single row.

Megan-Wright: Thanks for the feedback, I've updated all of that now :)

Megan-Wright:
@AlanGreene I've had a go at putting the information in a DataTable instead but I can't seem to do it easily and I think it's quite important we get the triggers 0.3.0 stuff supported ASAP... we could maybe think about changing it later if it becomes an issue?

AlanGreene:
How about something like this?

```js
const headers = [
  {
    key: 'key',
    header: intl.formatMessage({
      id: 'dashboard.tableHeader.key',
      defaultMessage: 'Key'
    })
  },
  {
    key: 'expression',
    header: intl.formatMessage({
      id: 'dashboard.tableHeader.expression',
      defaultMessage: 'Expression'
    })
  }
];

const rows = overlays.map(overlay => ({
  id: overlay.key, // assuming key is unique
  key: overlay.key,
  expression: overlay.expression
}));

<Table
  headers={headers}
  rows={rows}
/>
```

src/containers/ClusterTriggerBindings/ClusterTriggerBindings.js
```diff
@@ -121,7 +121,7 @@ export /* istanbul ignore next */ class ClusterTriggerBindings extends Component
           emptyTextAllNamespaces={intl.formatMessage(
             {
               id: 'dashboard.emptyState.clusterTriggerBindings',
-              defaultMessage: 'No {kind}.'
+              defaultMessage: 'No {kind}'
```
AlanGreene: 👍

Megan-Wright:
@AlanGreene Updated that to a table now - thank you for the code :)

Issue #1009, opened 2020-02-05
Have known secrets clickable from ServiceAccounts page


a-roberts:
# Expected behavior
If it's a Tekton annotated secret (so one we have under the secrets page) it'd be helpful to have these linked to from the ServiceAccounts page where such secrets are displayed

# Actual behavior
As of March 2020:

No linking is done currently and we see a tiny amount of data
![image](https://user-images.githubusercontent.com/12814972/76203304-4ec70d00-61ee-11ea-979b-93c050a9d9da.png)

# Steps to reproduce the problem

1. Create a secret and have it patched to a ServiceAccount (e.g. make it through the Dashboard so it has the tekton.dev annotation)
2. Visit the ServiceAccounts page and select the ServiceAccount you patched said secret to
3. Realise it'd be nice to have a link back to view your full secret information

a-roberts:
@AlanGreene FYI, there's currently no full page details page for secrets so the plan is to introduce something basic for now (a copy of the existing detail pages, such as for PipelineResources), so that the cilckable link to get more details can be provided. 

We do the same kinda thing already for EventListeners and their TriggerBindings/Templates so I'm suggesting we apply the same style as well for the links here.

I'm curious if you think there's anything else we should be displaying on a secrets details page, perhaps one we can open up to the design team - let us know if you've got anything in mind

a-roberts:
Note that if we are to filter by name, we can use fieldSelector but IIRC there was a problem with it (perhaps not supported through the proxy, we've definitely tried something like this before) - we know labelSelector is good though and, worst case scenario, we could add labels to secrets we create as a not-ideal workaround.

AlanGreene:
> what value would this page offer that our secrets page doesn't already?

Right now very little, but I think that's OK because we've chosen to provide a simple experience around secrets that allows users to create them with the required annotation for Tekton, it's not intended to be a general purpose secrets editor / explorer.

There are other parts of the experience missing, such as the ability to edit a secret / patch it onto a ServiceAccount after exiting the create flow. These are covered in https://github.com/tektoncd/dashboard/issues/727 and https://github.com/tektoncd/dashboard/issues/733

a-roberts:
> > what value would this page offer that our secrets page doesn't already?
> 
> Right now very little, but I think that's OK because we've chosen to provide a simple experience around secrets that allows users to create them with the required annotation(s) for Tekton, it's not intended to be a general purpose secrets editor / explorer.
> 
[...]

PR #1138, opened 2020-03-12
Add secret details page

Megan-Wright:
# Changes

For: https://github.com/tektoncd/dashboard/issues/1009
- Adds a secret details page for individual secrets, showing basic data as well as a complete list of service accounts, as well as tests for this file
- Introduces links from individual service accounts to secrets and vice versa
- Updates tests to account for links and also removes a duplicate test for `getCredential`

Screenshot 2020-03-12 at 10 09 29

Screenshot 2020-03-12 at 10 09 36

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/utils/src/utils/router.js
+++ b/packages/utils/src/utils/router.js
@@ -86,6 +86,12 @@ export const paths = {
   secrets: {
     all() {
       return '/secrets';
+    },
+    byName() {
+      return byNamespace({ path: '/secrets/:secretName' });
+    },
+    byNamespace() {
+      return byNamespace({ path: '/secrets' });
     }
   },
   serviceAccounts: {
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -301,8 +301,8 @@ export function getAllCredentials(namespace) {
   return get(uri);
 }
 
-export function getCredential(id, namespace) {
-  const uri = getKubeAPI('secrets', { name: id, namespace });
+export function getCredential(name, namespace) {
+  const uri = getKubeAPI('secrets', name, namespace);
   return get(uri);
 }
 
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -48,6 +48,7 @@ import {
   PipelineRuns,
   Pipelines,
   ResourceList,
+  Secret,
   Secrets,
   ServiceAccount,
   ServiceAccounts,
@@ -205,6 +206,16 @@ export /* istanbul ignore next */ class App extends Component {
                   )}
 
                   <Route path={paths.secrets.all()} exact component={Secrets} />
+                  <Route
+                    path={paths.secrets.byName()}
+                    exact
+                    component={Secret}
+                  />
+                  <Route
+                    path={paths.secrets.byNamespace()}
+                    exact
+                    component={Secrets}
+                  />
                   <Route
                     path={paths.serviceAccounts.byName()}
                     exact
--- a/src/containers/Secret/Secret.js
+++ b/src/containers/Secret/Secret.js
@@ -1,0 +1,268 @@
+/*
+Copyright 2020 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React, { Component } from 'react';
+import '../../scss/Triggers.scss';
+import { injectIntl } from 'react-intl';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { Link } from 'react-router-dom';
+import { InlineNotification, Tag } from 'carbon-components-react';
+import { FormattedDate, Table } from '@tektoncd/dashboard-components';
+import { formatLabels, getErrorMessage, urls } from '@tektoncd/dashboard-utils';
+import { fetchServiceAccounts } from '../../actions/serviceAccounts';
+
+import {
+  getSecret,
+  getSecretsErrorMessage,
+  getServiceAccounts,
+  isFetchingSecrets,
+  isWebSocketConnected
+} from '../../reducers';
+
+import { fetchSecret } from '../../actions/secrets';
+
+export /* istanbul ignore next */ class SecretContainer extends Component {
+  componentDidMount() {
+    this.fetchData();
+  }
+
+  componentDidUpdate(prevProps) {
+    const { match, webSocketConnected } = this.props;
+    const { namespace, secretName } = match.params;
+    const {
+      match: prevMatch,
+      webSocketConnected: prevWebSocketConnected
+    } = prevProps;
+    const {
+      namespace: prevNamespace,
+      secretName: prevSecretName
+    } = prevMatch.params;
+
+    if (
+      namespace !== prevNamespace ||
+      secretName !== prevSecretName ||
+      (webSocketConnected && prevWebSocketConnected === false)
+    ) {
+      this.fetchData();
+    }
+  }
+
+  async fetchData() {
+    const { match } = this.props;
+    const { namespace, secretName } = match.params;
+    await this.props.fetchSecret({ name: secretName, namespace });
+    await this.props.fetchServiceAccounts();
+  }
+
+  render() {
+    const { error, intl, match, secret, loading, serviceAccounts } = this.props;
+    const { secretName, namespace } = match.params;
+    const serviceAccountsWithSecret = [];
+
+    if (serviceAccounts) {
+      serviceAccounts.forEach(serviceAccount => {
+        serviceAccount.secrets.forEach(secretInServiceAccount => {
+          if (secretInServiceAccount.name === secret.metadata.name) {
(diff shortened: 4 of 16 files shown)


CarolynMabbott:
Is it possible to change the username from being displayed in base64 to plain text?
As the username is already displayed in the secrets table in plain text it seems better to display them both in the same way

Megan-Wright:
> @Megan-Wright I've been thinking more about this, we should either
> 
> 1. actually, only show Tekton annotated secrets here, because those we care about
> 2. not display the yaml component, because if we show any old secret and they store sensitive stuff in a place other than data.password, it's gonna show up
> 
> You could say even with 1) they could do that, so not displaying the yaml component at all might actually be better going forward

Yep fair enough might be a security issue - removed the YAML view now

src/api/index.js
```diff
@@ -301,8 +301,8 @@ export function getAllCredentials(namespace) {
   return get(uri);
 }
 
-export function getCredential(id, namespace) {
-  const uri = getKubeAPI('secrets', { name: id, namespace });
+export function getCredential(name, namespace) {
+  const uri = getKubeAPI('secrets', name, namespace);
```
```suggestion
  const uri = getKubeAPI('secrets', { name, namespace });
```

I'm not sure how this is working since `getKubeAPI` expects an object containing the name and namespace as its second param...

The test should be update to ensure the values are being passed correctly to the API, currently fetchMock is just matching 'secrets' in the URL.

src/containers/Secret/Secret.js
```diff
@@ -0,0 +1,264 @@
...
+        serviceAccount.secrets.forEach(secretInServiceAccount => {
+          if (secretInServiceAccount.name === secret.metadata.name) {
+            serviceAccountsWithSecret.push(serviceAccount.metadata.name);
+          }
+        });
+      });
+    }
+
+    const emptyTextMessageSAs = intl.formatMessage({
+      id: 'dashboard.secret.noSAs',
```
AlanGreene:
```suggestion
      id: 'dashboard.secret.noServiceAccounts',
```

for clarity

src/containers/Secret/Secret.js
```diff
@@ -0,0 +1,264 @@
...
+        key: 'name',
+        header: intl.formatMessage({
+          id: 'dashboard.tableHeader.name',
+          defaultMessage: 'Name'
+        })
+      }
+    ];
+
+    let rowsForServiceAccounts = [];
+    rowsForServiceAccounts = serviceAccountsWithSecret.map(serviceAccount => {
```
AlanGreene:
```suggestion
    const rowsForServiceAccounts = serviceAccountsWithSecret.map(serviceAccount => {
```

src/containers/Secret/Secret.js
```diff
@@ -0,0 +1,264 @@
...
+                  id: 'dashboard.secret.type',
+                  defaultMessage: 'Type:'
+                })}
+              </span>
+              {type}
+            </p>
+          </div>
+          <Table
+            title={intl.formatMessage({
+              id: 'dashboard.secret.saTableTitle',
```
AlanGreene:
```suggestion
              id: 'dashboard.secret.serviceAccountTableTitle',
```

src/reducers/secrets.js
```diff
@@ -173,6 +173,15 @@ export function getSecrets(state, namespace) {
   return Object.values(state.byNamespace[namespace] || []);
 }
 
+export function getSecret(state, name, namespace) {
+  const resources = state.byNamespace[namespace] || {};
+  const secretToUse = resources[name];
+  if (secretToUse) {
+    return secretToUse;
+  }
+  return null;
```
AlanGreene: Can we simplify this?

```suggestion
  return resources[name];
```

if it's not found we'll just return `undefined` which should be fine

packages/utils/src/utils/router.js
```diff
@@ -86,6 +86,9 @@ export const paths = {
   secrets: {
     all() {
       return '/secrets';
+    },
+    byName() {
```
AlanGreene: the Secrets link in the SideNav is not selected when viewing a Secret details page

```diff
-to={urls.secrets.all()}
+to={this.getPath(urls.secrets.all())}
```
should resolve that
and you'll also need to add a Route for `urls.secrets.byNamespace` and an entry for that here

src/api/index.js
Megan-Wright: yeah so I thought it would be done the way you've suggested but passing the variables in like that gives you an encoded url back that doesn't lead anywhere - only works with passing the variables in how I've done it seemingly

Megan-Wright: Also I'm not sure I understand what you mean about changing the test? I just copied how the other functions in `index.js` have been tested

Megan-Wright:
> Can we add some space between the labels and values here like we have on the trigger pages? Should be able to use the same styles
> 
> image

Done this now :)

src/containers/Secret/Secret.js
Megan-Wright: Done 😸

src/containers/Secret/Secret.js
Megan-Wright: Done 😸

src/containers/Secret/Secret.js
Megan-Wright: Done 😸

src/reducers/secrets.js
Megan-Wright: yep good point, updated now 😸

packages/utils/src/utils/router.js
Megan-Wright: Done 😸

CarolynMabbott:
Slight issue I came across 
After i go on describe secret page i then click back on the secrets tab and i get this error 
`comms.js?cc45:56 GET http://localhost:8000/proxy/apis/namespaces/tekton-pipelines/namespaces/tekton-pipelines/secrets/`

AlanGreene:
> Slight issue I came across
> After i go on describe secret page i then click back on the secrets tab and i get this error
> `comms.js?cc45:56 GET http://localhost:8000/proxy/apis/namespaces/tekton-pipelines/namespaces/tekton-pipelines/secrets/`

That should be resolved by https://github.com/tektoncd/dashboard/pull/1138#discussion_r391568735
> and you'll also need to add a Route for urls.secrets.byNamespace

src/api/index.js
AlanGreene: ok so the reason this was working is that the `name` parameter being passed to `getCredential` was actually an object containing both the name and namespace

the function should be written like this:

```diff
-export function getAllCredentials(namespace) {
-  const uri = getKubeAPI('secrets', namespace);
-  return get(uri);
-}
-
-export function getCredential(name, namespace) {
-  const uri = getKubeAPI('secrets', name, namespace);
+export function getCredential({ name, namespace }) {
+  const uri = getKubeAPI('secrets', { name, namespace });
   return get(uri);
 }
```

removed `getAllCredentials` as it's not used anywhere, `/src/api/index.test.js` contains a test that can be deleted, and `/src/actions/secrets.test.js` contains a reference to this function that can also be removed without affecting the test (it wasn't doing anything anyway...)

as for the test for `getCredential`, it didn't catch this error so obviously could do with being updated to properly check the  correct values are being passed as expected. We should also review the other tests to ensure we don't have similar gaps but that's beyond the scope of this PR.

AlanGreene:
Also don't forget to  `npm run i18n:extract` to update messages_en.json with the new strings.

src/api/index.js
Megan-Wright: thanks for explaining this to me, I'll have a go at updating it now

Megan-Wright:
> > Slight issue I came across
> > After i go on describe secret page i then click back on the secrets tab and i get this error
> > `comms.js?cc45:56 GET http://localhost:8000/proxy/apis/namespaces/tekton-pipelines/namespaces/tekton-pipelines/secrets/`
> 
> That should be resolved by [#1138 (comment)](https://github.com/tektoncd/dashboard/pull/1138#discussion_r391568735)
> 
> > and you'll also need to add a Route for urls.secrets.byNamespace

Hi Alan, the behaviour I'm seeing is that without your fix (referenced above), the sidebar isn't highlighted but the page is routed correctly. However, with your fix, the sidebar is highlighted but the routing breaks, causing what Carolyn is seeing where clicking on an individual secret then back to the secrets page causes it to break... I'm unsure of why this is happening or how to fix it?

AlanGreene:
@Megan-Wright can you push your latest changes and I'll take a look?

Megan-Wright:
@AlanGreene the code currently on this branch has your fix for https://github.com/tektoncd/dashboard/pull/1138#discussion_r391568735, and is also where Carolyn saw the issue with rerouting.

AlanGreene:
I don't see a URL for `secrets.byNamespace` in the router utils, or a matching `Route` in App.js
Are you sure you've pushed your latest changes?

Megan-Wright:
> I don't see a URL for `secrets.byNamespace` in the router utils, or a matching `Route` in App.js
> Are you sure you've pushed your latest changes?

Oh yeah, oops - pushed that now

src/containers/App/App.js
```diff
@@ -205,6 +206,16 @@ export /* istanbul ignore next */ class App extends Component {
...
                   <Route path={paths.secrets.all()} exact component={Secrets} />
+                  <Route
+                    path={paths.secrets.byName()}
+                    exact
+                    component={Secret}
+                  />
+                  <Route
+                    path={paths.secrets.byNamespace()}
+                    exact
+                    component={Secret}
```
AlanGreene:
```suggestion
                    component={Secrets}
```

Issue #1136, opened 2020-03-11
Secrets panelling is not clear as to how to use accesstokens


dibbles:
The dashboard UI does not clearly indicate that the secret of type accesstoken is to create an opaque secret creating an access token, rather than a basic-auth secret with the token in the password field.

This will potentially confuse new users.... the design around this needs work, as if users can specify the token in the user/password situation (in place of the password) - it is not clearly indicated.

![Screenshot 2020-03-11 at 15.46.18.png](https://images.zenhubusercontent.com/5809f18995e42b6c5baad615/e6621865-9a33-47d3-95b9-fe33069ff0db)

a-roberts:
Merged

PR #1141, opened 2020-03-12
Add tooltip for secret creation guidance

a-roberts:
For https://github.com/tektoncd/dashboard/issues/1136

# Changes

Adds a tooltip with a translatable string

PIc:
image

and

image

# Submitter Checklist

review them:

for more details._


a-roberts:
Adding a test now, layout/text could do with a look through, @AlanGreene what do you think?

a-roberts:
Wanted to add this test:

```

it('Secrets page displays a tooltip', async () => {
  jest.spyOn(selectors, 'isReadOnly').mockImplementation(() => false);

  const currentProps = {
    loading: false,
    error: null,
    intl
  };

  const { getByTestId, queryByTestId, queryByText } = renderWithRouter(
    <Provider store={store}>
      <Route
        path={urls.secrets.all()}
        render={props => <Secrets {...props} {...currentProps} />}
      />
    </Provider>,
    { route: urls.secrets.all() }
  );
  await waitForElement(() => queryByText(/github-repo-access-secret/i));
  const foundTooltip = await waitForElement(() => queryByTestId('theSecretHelperTooltip'));
  console.log(foundTooltip);;
  fireEvent.click(foundTooltip);
  expect(queryByText(/use password if you plan to/i)).toBeTruthy();
});
```

but it never finds the tooltip by test ID ever...data-testid='theSecretHelperTooltip' added too

src/containers/Secrets/Secrets.js
```diff
@@ -55,6 +57,9 @@ const initialState = {
   selectedType: 'password'
 };
 
+const secretCreationHelperString =
+  'Use Password if you plan to have a PipelineRun that clones from a Git repository or image registry that requires authentication. Use Access Token when interacting with webhooks or PullRequest resources. Check the official Tekton Pipelines documentation for more details.';
```
AlanGreene: This also applies to TaskRuns, and I think 'official' is implied. How about something like this?

> Use Password with git or image PipelineResources that require authentication. Use Access Token with webhooks or with pullRequest PipelineResources. Check the Tekton Pipelines documentation for more details on authentication.

src/containers/Secrets/Secrets.js
```diff
@@ -384,6 +389,20 @@ export /* istanbul ignore next */ class Secrets extends Component {
           <>
             <h1>Secrets</h1>
             <LabelFilter {...this.props} />
+            <Tooltip
+              className="secretHelperTooltip"
+              direction="bottom"
+              tabIndex={0}
+              tooltipBodyId="secret-type-helper"
+              triggerText="Secret type"
```
AlanGreene: also needs the react-intl treatment

src/containers/Secrets/Secrets.js
```diff
@@ -384,6 +389,20 @@ export /* istanbul ignore next */ class Secrets extends Component {
           <>
             <h1>Secrets</h1>
             <LabelFilter {...this.props} />
+            <Tooltip
```
AlanGreene: Would this be more helpful on the create secret form? Maybe we want it in both places?

If we're keeping the tooltip here above the table it would be great if we could put it inline with the radio buttons instead of adding another heading here taking up vertical space. Would the icon be enough on its own?

src/containers/Secrets/Secrets.scss
```diff
@@ -0,0 +1,4 @@
+.bx--tooltip__label {
```
AlanGreene: missing copyright header

src/containers/Secrets/Secrets.js
```diff
@@ -11,14 +11,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import './Secrets.scss';
```
AlanGreene: nit: can we move this to end of the imports? That would be consistent with other files where we roughly have:
1. external dependencies
1. internal
1. styles and other static resources

src/containers/Secrets/Secrets.scss
```diff
@@ -0,0 +1,4 @@
...
+  font-size: 120% !important;
```
AlanGreene: Maybe the text shouldn't be part of the tooltip? Also let's try to avoid `!important` as much as possible, instead we can make the selector slightly more specific (add a class to a parent element that we can reference) so we're only targeting this instance and it won't be overridden by more general styles.

src/containers/Secrets/Secrets.js
a-roberts: Yeah, I like it just on the creation page actually, that's where a user wants to create it and I like to think if you're trying to filter by a particular type you're wary of what each does, so I've moved it there

src/containers/Secrets/Secrets.scss
a-roberts: Hmm, what do you mean by not being part of the tooltip? 

Just some text above the type perhaps instead of using a tooltip? Or, having something like <Tooltip><our own div with ID that we style>...</div></Tooltip>?

a-roberts: Actually it does look a bit weird on the form if this has its own style so I'm gonna omit it

src/containers/Secrets/Secrets.js
a-roberts: @AlanGreene On the secret create page I've replaced the existing "legend" text with this tooltip with text, what do you think? Or I could keep the legend in and have the tooltip here but without the triggerText property? Otherwise, if there's both, we have "Type" and also "Secret type" - changes pushed up addressing the other comments too, thanks

image

a-roberts:
Ready for another look please @AlanGreene

src/components/CreateSecret/Form/UniversalFields.js
```diff
@@ -73,13 +74,24 @@ const UniversalFields = props => {
...
       />
-      <FormGroup
-        legendText={intl.formatMessage({
-          id: 'dashboard.universalFields.secretType',
-          defaultMessage: 'Type'
-        })}
-        disabled={loading}
-      >
+      <FormGroup disabled={loading}>
+        <Tooltip
```
AlanGreene: You can set the tooltip as the legendText. Having the legend is important for accessibility so that a screenreader can announce the correct group regardless of tab direction, i.e. if a user tabs backwards from the username input to the selected radio button it will announce the group + selection.

```js
      <FormGroup
        disabled={loading}
        legendText={
          <Tooltip
            direction="right"
            tabIndex={0}
            ...
```

a-roberts: Thanks, pushed

Issue #1144, opened 2020-03-13
Get 2 identical errors when creating a secret that already exists 


CarolynMabbott:
# Expected behavior
I only get one error 

# Actual behavior
I get two which are identical 

# Steps to reproduce the problem

1. Create two secrets (the second being identical to the first)

# Environment
- Kubernetes Platform:

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

master 

# Additional Info
It's also strange that if you click `x` on the second one then only the second error disapeears but if you click on the first `x` then both disappear 
<img width="1410" alt="Screenshot 2020-03-13 at 11 48 28" src="https://user-images.githubusercontent.com/32342288/76618586-dd09fe80-6520-11ea-82cb-c68f615bd820.png">

a-roberts:
![image](https://user-images.githubusercontent.com/12814972/76755792-679a6a00-677c-11ea-9197-afb7fd08818c.png)

on my code atm, git log being

```
0d8ecfbc (adam/secretcreatetooltip)
Author: a-roberts <aroberts@uk.ibm.com>
Date:   Thu Mar 12 15:38:26 2020 +0000

    Add tooltip for secret creation guidance

commit ce8b1c4ab85df23007f1058a139d8ea463f7d4b8 (origin/master, origin/HEAD)
Author: a-roberts <aroberts@uk.ibm.com>
Date:   Thu Mar 12 11:31:36 2020 +0000

    Don't always display secret notifications

commit b03c4b885ed8a6e54d2c0cf03c6afd6647585920
Author: Alan Greene <ALANGREE@ie.ibm.com>
Date:   Fri Mar 6 22:28:43 2020 +0000

    Extract Dropdown strings for transl
```

Have the same on master, going to close unless we get a reproduce on this one, @Megan-Wright and I really can't

AlanGreene:
Reproduced this a couple of times just now, not 100% sure of the steps but I'll try to narrow it down and add details

AlanGreene:
2 parts:

1. double error
  we have 2 separate props being passed to CreateSecret/Form with error messages: `errorMessageDuplicate` and `errorMessageCreated`. There should really just be one `errorMessage` that's set to the appropriate value, this way we simplify the Form code and avoid the possibility of displaying the same message for multiple reasons simultaneously.

1. not handling the duplicate with our cleaner custom error message in at least one case:
  i. on the Create Secret page create a secret in namespace A
[...]

PR #1153, opened 2020-03-17
Fix duplicate errors for secret notifications

a-roberts:
# Changes

For https://github.com/tektoncd/dashboard/issues/1144

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/Secrets/Secrets.js
+++ b/src/containers/Secrets/Secrets.js
@@ -41,7 +41,6 @@ import {
   getDeleteSecretsSuccessMessage,
   getPatchSecretsSuccessMessage,
   getSecrets,
-  getSecretsErrorMessage,
   getSelectedNamespace,
   getServiceAccounts,
   isFetchingSecrets,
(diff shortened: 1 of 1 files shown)


a-roberts:
Noticed getSecretsErrorMessage being called twice, this is an improvement but seeing an error for the on close button click handler now and the message I do get (once, not twice) is unformatted, even though everywhere we've got an InlineNotification it's using getErrorMessage which surely gives us the nicely formatted one

src/containers/Secrets/Secrets.js
```diff
@@ -502,8 +501,7 @@ function mapStateToProps(state, props) {
   const namespace = namespaceParam || getSelectedNamespace(state);
 
   return {
-    errorMessage:
-      getDeleteSecretsErrorMessage(state) || getSecretsErrorMessage(state),
+    errorMessage: getDeleteSecretsErrorMessage(state),
```
AlanGreene: do we end up missing some classes of error now?

ideally we'd just have one 'error' state, and set it with the correct details depending on whether it's create, delete, etc.

a-roberts: Yeah that's my concern too so just going through it again now (it seemed a little easy, I do think moving towards a (I think you said `global`) error state would thwart these kind of problems from repeatedly occurring)

a-roberts: Had a good try creating secrets, switching namespaces etc, even deleting ones (shouldn't be impacted) with errors (e.g. deleting first with a kubectl command), and patching to one we just deleted - that reports a success but I'm wondering if it did anyway on master so will check

a-roberts:
@AlanGreene can you take another look please? Didn't find any notifications no longer showing (see latest comment)

Issue #1164, opened 2020-03-18
Error creating secrets named the same as namespaces


CarolynMabbott:
# Expected behavior
I can create a secret called tekton-pipelines and not get an error 

# Actual behavior

# Steps to reproduce the problem

1. Create secret called tekton-pipelines or default 
2.
3.

# Environment
- Kubernetes Platform:

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

master 

# Additional Info
It errors with 
<img width="1434" alt="Screenshot 2020-03-18 at 13 11 59" src="https://user-images.githubusercontent.com/32342288/76970031-27c1b700-6923-11ea-9fcd-9a191851b596.png">

When you try and click on it you get the error 
<img width="1409" alt="Screenshot 2020-03-18 at 14 17 54" src="https://user-images.githubusercontent.com/32342288/76970134-5475ce80-6923-11ea-8dae-2118283a3907.png">

PR #1189, opened 2020-03-25
Secrets with names the same as a namespace can be created

Megan-Wright:
# Changes
For: https://github.com/tektoncd/dashboard/issues/1164
- The secrets reducer now has it's own `isSecretStale` function as secrets were not working with the `isStale` function in utils. 
- Should be a temporary fix while we work to introduce `byId()` to secrets, so that it can share the `isStale` function in utils.
- This means that secrets with the same name as their namespace do not now cause errors in the dashboard

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/reducers/secrets.js
+++ b/src/reducers/secrets.js
@@ -21,11 +21,12 @@ function byNamespace(state = {}, action) {
     case 'SECRETS_FETCH_SUCCESS':
       const namespaces = action.data.reduce((accumulator, secret) => {
         const { metadata } = secret;
-        if (isStale({ metadata }, state, 'name')) {
+        const { name, namespace } = metadata;
+        const secretsState = state[namespace] || {};
+        if (isStale({ metadata }, secretsState, 'name')) {
           return state;
         }
 
-        const { name, namespace } = metadata;
         return merge(accumulator, {
           [namespace]: {
             [name]: { ...secret }
@@ -35,8 +36,9 @@ function byNamespace(state = {}, action) {
       return merge({}, state, namespaces);
     case 'SecretCreated':
     case 'SecretUpdated':
+      const secretsState = state[action.payload.metadata.namespace] || {};
       if (
-        isStale(action.payload, state, 'name') ||
+        isStale(action.payload, secretsState, 'name') ||
         action.payload.type !== 'kubernetes.io/basic-auth'
       ) {
         return state;


src/reducers/secrets.js
```diff
@@ -14,14 +14,28 @@ limitations under the License.
 import { combineReducers } from 'redux';
 import merge from 'lodash.merge';
 import { ALL_NAMESPACES } from '@tektoncd/dashboard-utils';
-import { isStale } from '../utils';
+
+// TODO: replace isSecretStale() with utils.js isStale() when secrets store uses
+// byId structure.
```
AlanGreene: > when secrets store uses byId structure

what's stopping us from doing this?

src/reducers/secrets.js
```diff
@@ -14,14 +14,28 @@ limitations under the License.
...
+  const incomingVersion = parseInt(resource.metadata.resourceVersion, 10);
+  return existingVersion > incomingVersion;
+}
 
 function byNamespace(state = {}, action) {
   switch (action.type) {
     case 'SECRETS_FETCH_SUCCESS':
       const namespaces = action.data.reduce((accumulator, secret) => {
         const { metadata } = secret;
-        if (isStale({ metadata }, state, 'name')) {
```
AlanGreene: could we avoid creating a custom 'isStale' by adjusting the call to pass in the relevant slice of the state instead?

```js
const secretsState = state[namespace] || {};
if (isStale({ metadata }, secretsState, 'name')) {
```

src/reducers/secrets.js
Megan-Wright: Nothing, but I wouldn't know how to do it and I thought it was better to have a temporary solution in place to avoid the errors. I was going to make a follow up issue after this one was merged.

src/reducers/secrets.js
Megan-Wright: oh yeah this might be a better way  of doing it, I'll have a go at that now :)

Megan-Wright: @AlanGreene works perfectly, thanks :) I've updated this now

src/reducers/secrets.js
AlanGreene: no problem, can you link the issue here when you create it?

Megan-Wright:
@AlanGreene https://github.com/tektoncd/dashboard/issues/1191

Issue #1223, opened 2020-04-01
Don't properly handle K8s lifecycle


mattmoor:
When I run the dashboard as a Knative service this really pops out because we run with a higher default `terminationGracePeriodSeconds`.

Basically K8s sends SIGTERM to tell the dashboard to drain, and then SIGKILL after the grace period.  `knative.dev/pkg/signals` has a way of getting a channel ad a context to facilitate adhering to this.

mattmoor:
Looks like you are using the signal package from `k8s.io/sample-controller`, which is surely an interesting dependency :)

The informers seem to take the stopCh that sets up, but nothing ever drains the HTTP servers: https://github.com/tektoncd/dashboard/blob/4faecbc65ff0c0ee58269e100468f35b1018fa12/cmd/dashboard/main.go#L102-L103

mattmoor:
We invoke `ListenAndServe()` in a goroutine and then we see the `stopCh` file we call `server.Shutdown(context.Background())`.

mattmoor:
I'll take a crack at this, since at this point it should be plumbing.

a-roberts:
Merged so closing

PR #1225, opened 2020-04-01
This fixes the signal handling in the dashboard.

mattmoor:
See the linked issue for much more detail.

Fixes: https://github.com/tektoncd/dashboard/issues/1223

# Changes

Swap to knative signal package (based on the one used today anyways, but sheds a dep), and add logic to `Shutdown` the http server when the stop signal is received.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -703,14 +703,6 @@
   revision = "f442ecb314a3679150c272e2b9713d8deed5955d"
 
 [[projects]]
-  digest = "1:e1fc4a624e8497252148125a50cec64b416d69ef69afc2244498e0c9b73060d5"
-  name = "k8s.io/sample-controller"
-  packages = ["pkg/signals"]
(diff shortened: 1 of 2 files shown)


a-roberts:
Thanks for this @mattmoor 🎉 taking a look now

a-roberts:
Yeah this is good, so...

We already have a knative.dev/pkg requirement being used anyway (it's referenced in our Gopkg.toml and without it we get Convertable related errors popping up for Tekton Tasks at least)

https://github.com/tektoncd/pipeline/search?q=signals&unscoped_q=signals this particular dep is used throughout Pipelines here

Tested this out and noticed no problems - so built and deployed this through kustomize using my Docker Desktop cluster

Will take Matt's word for it that there are benefits when using this as a Knative service and certainly nothing breaks from what I've seen 🎉 

/lgtm
/approve
/meow hats

PR #1231, opened 2020-04-02
Update Table to add tooltips to all string values passed as cell content

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1232

We display tooltips on Table cell contents which is important
when the content may be truncated. This was not applied consistently
across all pages.

Update the Table component to automatically apply a tooltip to
any cell data of type `string` so this no longer has to be handled
by the consumer. This also makes it easier for us to update the tooltip
handling in future, for example adopting one of the Carbon tooltip components
as it's now controlled more centrally.

In the case of non-string cell data, the consumer is modifying / wrapping
the content in a custom manner, adding the tooltip remains the responsibility
of the consumer, e.g. we display full date in tooltip instead of
the relative time format displayed in the table.

This will likely change in a future PR related to sorting / filtering,
which would adopt a more flexible approach to specifying the contents,
allowing the consumer to provide tooltip values and sortable values
related to the display value.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineResources/PipelineResources.js
+++ b/packages/components/src/components/PipelineResources/PipelineResources.js
@@ -77,14 +77,10 @@ const PipelineResources = ({
           {pipelineResourceName}
         </Link>
       ) : (
-        <span title={pipelineResourceName}>{pipelineResourceName}</span>
+        pipelineResourceName
       ),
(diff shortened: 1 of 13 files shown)


sbwsg:
@googlebot rescan

Issue #1258, opened 2020-04-07
Wrong log format ?


eddycharly:
https://github.com/tektoncd/dashboard/blob/f6568713d1e81f64c74c715a47079e7614282099/pkg/endpoints/dashboard.go#L62

Shouldn't it be `logging.Log.Infof("Dashboard is in read-only mode: %t", asBool)` instead ?

eddycharly:
@AlanGreene what’s your opinion on this one https://github.com/tektoncd/dashboard/issues/1257 ?

I could open a PR if you’re ok with it.

AlanGreene:
I'll take a look now and respond over there

PR #1259, opened 2020-04-07
Fix log format

eddycharly:
Fixing #1258


AlanGreene:
Thanks @eddycharly, it's nice to clean up these things.

/ok-to-test

AlanGreene:
For reference, this updates the log line from:

`{"level":"info","msg":"Dashboard is in read-only mode: %!s(bool=false)"}`

to

`{"level":"info","msg":"Dashboard is in read-only mode: false"}`

Much nicer! 😄

eddycharly:
Indeed ;)

Issue #1261, opened 2020-04-08
Add support for configurable triggers and pipeline namespaces


eddycharly:
The namespace where Pipelines and/or Triggers is deployed is hard coded in a few places, see below for an example:

https://github.com/tektoncd/dashboard/blob/f6568713d1e81f64c74c715a47079e7614282099/pkg/endpoints/dashboard.go#L112

This makes it difficult to deploy the dashboard in a different namespace than `tekton-pipelines` (it works but a few things are failing).

a-roberts:
Merged so closing, thanks for this @eddycharly!

PR #1262, opened 2020-04-08
Add support for configurable triggers and pipeline namespaces

eddycharly:
Fixing #1261.

This PR adds support for configurable Triggers and Pipelines namespaces.

Triggers and Pipelines namespaces can be configured using `TRIGGERS_NAMESPACE` and `PIPELINE_NAMESPACE` environment variables (defaulting to `tekton-pipelines` if not set).

### DIFF
--- a/pkg/endpoints/dashboard.go
+++ b/pkg/endpoints/dashboard.go
@@ -83,8 +83,8 @@ func IsOpenShift(r Resource, installedNamespace string) bool {
 		if namespaceName == "openshift-pipelines" {
 			openshiftPipelineFound = SearchForDeployment(r, "pipelines", "openshift-pipelines")
 		}
-		if namespaceName == "tekton-pipelines" {
-			tektonPipelinesFound = SearchForDeployment(r, "pipelines", "tekton-pipelines")
+		if namespaceName == installedNamespace {
+			tektonPipelinesFound = SearchForDeployment(r, "pipelines", installedNamespace)
 		}
 	}
 
@@ -107,9 +107,9 @@ func IsOpenShift(r Resource, installedNamespace string) bool {
 }
 
 // Get pipelines version
-func GetPipelineVersion(r Resource, isOpenShift bool) string {
+func GetPipelineVersion(r Resource, pipelineNamespace string, isOpenShift bool) string {
 	version := ""
-	namespacesToCheck := []string{"tekton-pipelines"}
+	namespacesToCheck := []string{pipelineNamespace}
 
 	if isOpenShift {
 		namespacesToCheck = append(namespacesToCheck, "openshift-pipelines")
@@ -217,9 +217,9 @@ func getDeployments(r Resource, thingSearchingFor string, namespace string) stri
 }
 
 // Get triggers version
-func GetTriggersVersion(r Resource, isOpenShift bool) string {
+func GetTriggersVersion(r Resource, triggersNamespace string, isOpenShift bool) string {
 	version := ""
-	namespacesToCheck := []string{"tekton-pipelines"}
+	namespacesToCheck := []string{triggersNamespace}
 
 	if isOpenShift {
 		namespacesToCheck = append(namespacesToCheck, "openshift-pipelines")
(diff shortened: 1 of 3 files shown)


pkg/endpoints/dashboard.go
```diff
@@ -217,9 +217,9 @@ func getDeployments(r Resource, thingSearchingFor string, namespace string) stri
 }
 
 // Get triggers version
-func GetTriggersVersion(r Resource, isOpenShift bool) string {
+func GetTriggersVersion(r Resource, installedNamespace string, isOpenShift bool) string {
 	version := ""
-	namespacesToCheck := []string{"tekton-pipelines"}
+	namespacesToCheck := []string{installedNamespace}
```
CarolynMabbott, line 222: Tekton Triggers release yaml is currently hard coded to `tekton-pipelines` so is the most likely place to find it installed
If the dashboard is installed into a different namespace then this wouldn't pick up the triggers version on the About page as it would then look for Triggers in the installed namespace of the dashboard

pkg/endpoints/dashboard.go
```diff
@@ -241,9 +241,9 @@ func GetTriggersVersion(r Resource, isOpenShift bool) string {
 }
 
 // Check whether Tekton Triggers is installed
-func IsTriggersInstalled(r Resource, isOpenShift bool) bool {
+func IsTriggersInstalled(r Resource, installedNamespace string, isOpenShift bool) bool {
 	isTriggersInstalled := false
-	namespacesToCheck := []string{"tekton-pipelines"}
+	namespacesToCheck := []string{installedNamespace}
```
CarolynMabbott, line 246: Same comment as above for the func `GetTriggersVersion` about triggers install namespace

pkg/endpoints/dashboard.go
```diff
@@ -107,9 +107,9 @@ func IsOpenShift(r Resource, installedNamespace string) bool {
 }
 
 // Get pipelines version
-func GetPipelineVersion(r Resource, isOpenShift bool) string {
+func GetPipelineVersion(r Resource, installedNamespace string, isOpenShift bool) string {
 	version := ""
-	namespacesToCheck := []string{"tekton-pipelines"}
+	namespacesToCheck := []string{installedNamespace}
```
CarolynMabbott, line 112: Tekton Pipeline release yaml is always installed into `tekton-pipelines` except when using openshift operator where it is installed in `openshift-pipelines` so is the most likely place to find it installed
If the dashboard is installed into a different namespace then this wouldn't pick up the pipeline version on the About page as it would then look for Pipeline in the installed namespace of the dashboard and would surface errors

CarolynMabbott:
@eddycharly `INSTALLED_NAMESPACE` is for tekton-dashboard only 
Doesnt work for pipelines / triggers and they currently are defaulted via their `release.yaml` to only being installed in `tekton-pipelines` or `openshift-pipelines` if used through openshift operator

pkg/endpoints/dashboard.go
eddycharly: I know this is not perfect yet, that's why i added a note in the PR description, i thought about creating another PR to address this issue.

It could be improved by adding related environment variables i guess.

As i am working on adding Tekton helm charts, the target namespace can now be configured at release install time.

eddycharly:
@CarolynMabbott I am proposing this change to allow deploying Tekton components with helm charts.
Of course, it should not break other deployment solutions, just add more flexibility.

pkg/endpoints/dashboard.go
eddycharly: I haven't started working on Triggers helm chart yet, but once it's done the user will be able to choose the target namespace to deploy into.

What do you think about adding a `TRIGGERS_NAMESPACE` environment variable (defaulting to `tekton-pipelines` if not set) ?

eddycharly:
@CarolynMabbott I changed the PR to allow specifying Triggers and Pipelines namespaces separately, and defaulting to `tekton-pipelines` when not specified.

CarolynMabbott:
@eddycharly thanks for the updates, I'll test it out today to make sure basic functionality isnt broken

eddycharly:
@CarolynMabbott did you have time to try it ?

CarolynMabbott:
Works with tekton-dashboard being installed in tekton-pipelines namespace for both openshift and docker desktop 
Works on a non tekton-pipelines install on docker desktop

Doesnt appear to cause any errors that are noticeable 

@AlanGreene any thoughts on this pr?

AlanGreene:
Thanks @eddycharly 
/lgtm

Issue #1049, opened 2020-02-13
Dashboard should set the Content-Security-Policy header


mnuttall:
The `Content-Security-Policy` header is designed to modify the way browsers render pages, and thus to protect from various cross-site injections, including Cross-Site Scripting. It is important to set the header value correctly, in a way that will not prevent proper operation of the web site. Content-Security-Policy HTTP header allows you to create a whitelist of sources of trusted content and instructs the browser to only execute or render resources from those sources. 

IBM staff can find more information at https://github.ibm.com/IBMCloudPak4Apps/icpa-security/issues/10

PR #1280, opened 2020-04-10
Add Content-Security-Policy

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1049

I've added the Content-Security-Policy meta tag for both dev
and prod builds to keep us as close as possible to prod restrictions
while developing. There are 2 differences for dev, it allows
'unsafe-eval' scripts, and blob: URIs for styles. Both of these
are to accommodate the webpack dev server and hot reloading.

The policy is quite strict, it blocks everything by default, then
allows for scripts, styles, images, fonts, and connections to the
same origin only, with the exception of fonts which can also
come from Google Fonts. Taking this approach means that object embeds,
media, frames, etc. are all blocked giving us a more locked down
environment.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/index.template.html
+++ b/src/index.template.html
@@ -1,7 +1,8 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="utf-8" />
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <title><%= htmlWebpackPlugin.options.title %></title>
   </head>
   <body>
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -13,7 +13,14 @@ limitations under the License.
 const path = require('path');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 
-module.exports = {
+const contentSecurityPolicy = {
+  development:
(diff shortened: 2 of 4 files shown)


AlanGreene:
@a-roberts I've tested this out on CRC and not seeing any issues. Might be worth a quick check on a Kabanero environment to make sure there's no unexpected surprises lurking there given how strict these settings are.

a-roberts:
> @a-roberts I've tested this out on CRC and not seeing any issues. Might be worth a quick check on a Kabanero environment to make sure there's no unexpected surprises lurking there given how strict these settings are.

Good idea, alright I'll have a go today, could be something we bundle up into 0.6.1 as well - thanks for the PR 👍

a-roberts:
Not seeing any problems here, let's get it in
/lgtm
/approve

IamGabrielWu:
it seems to have trouble loading script from cloudflare when dashboard is set up with cloudflare proxy enabled. 

error
```
Refused to load the script 'https://ajax.cloudflare.com/cdn-cgi/scripts/xxx/cloudflare-static/rocket-loader.min.js' because it violates the following Content Security Policy directive: "script-src 'self'". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```
I think it needs to allow cloudflare as well.

Issue #1285, opened 2020-04-12
v0.6.1 non-openshift k8s release looks for openshift routes and fails to display some taskruns 


bbhuston:
# Expected behavior

The taskrun logs will properly display in the tekton dashboard.

# Actual behavior

Looks at logs of certain taskruns throws react error in tekton dashboard.  Dashboard appears to be looking for openshift networking despite being a standard k8s install.

# Steps to reproduce the problem

1. kubectl -n tekton-pipelines apply -f https://github.com/tektoncd/dashboard/releases/download/v0.6.1/tekton-dashboard-release.yaml
2. tkn task create -f https://raw.githubusercontent.com/tektoncd/pipeline/master/examples/v1beta1/taskruns/git-resource.yaml -n tekton-pipelines 
3. kubectl -n tekton-pipelines port-foward svc/tekton-dashboard 9097:9097 &
4. Go to "localhost:9097" and navigate to taskruns created upon.  Here you will see the following error message.

```
Error loading page

Minified React error #185; visit https://reactjs.org/docs/error-decoder.html?invariant=185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```

5. In Chrome developer console I see the following 500 status code for the `/` endpoint.  Same error appears in firefox as well.  There appears to be an extra `/` getting tacked onto the request URL.

```
Request URL: http://localhost:9097/proxy/apis/route.openshift.io/v1//
Request Method: GET
Status Code: 500 
Remote Address: 35.224.93.161:443
Referrer Policy: no-referrer-when-downgrade
```

# Environment
- Kubernetes Platform:

GKE cluster v15.9.   Istio v1.5
 
- Tekton Pipelines version:

0.11.1

- Tekton Triggers version:

0.4.0

- Tekton Dashboard version:

0.6.1

# Additional Info

a-roberts:
Thanks for the report @bbhuston! For the OpenShift route problem, yeah you're right and there's an issue for resolving it here https://github.com/tektoncd/dashboard/issues/778

Your TaskRun problem is more serious, I'll see if I can take a look today, @AlanGreene @steveodonovan  FYI 👀

a-roberts:
I see it,

```

vendors.187b855….js:93 TypeError: Object(...) is not a function
    at TaskRunContainer.loadTaskRun (main.9943ff2….js:1)
    at TaskRunContainer.value (main.9943ff2….js:1)
    at _o (vendors.187b855….js:93)
[...]

PR #1287, opened 2020-04-14
Fix step order issues

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1285

Dealing with unnamed steps has caused a number of problems in
the past. We've put in various fixes for specific cases but
every so often find another case where our assumptions don't
hold up.

This PR is an attempt to address that in a simpler and hopefully
more resilient way.

Unnamed steps are named based on the order they appear in the
Task spec. So given a Task with 2 unnamed steps these are
`unnamed-0` and `unnamed-1`. However, if the Task uses PipelineResources
there may be init steps included in the TaskRun status, and these
are also affect the order for naming unnamed steps.

e.g. given a Task using an input resource and having 1 unnamed
step, the steps listed in the status would be:
1. `some-resource-init-step`
1. `unnamed-1`

Instead of trying to match the status + spec by index across
the two lists:
1. extract the unnamed steps from the status (let's
call these `unnamedStepsStatus`)
1. sort the `unnamedStepsStatus` in natural order, i.e.
  `unnamed-2` < `unnamed-10`
1. iterate over the steps from the Task spec
1. if step name is falsy, shift the next item off the front of
  `unnamedStepsStatus` and use that as our step status
1. otherwise find a match by name as usual

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/utils/src/utils/index.js
+++ b/packages/utils/src/utils/index.js
@@ -132,16 +132,19 @@ export function reorderSteps(unorderedSteps, orderedSteps) {
   if (!unorderedSteps || !orderedSteps) {
     return [];
   }
-  return orderedSteps.map(({ name, ...rest }, idx) => {
-    let findName = name;
-    if (name === '') {
-      findName = `unnamed-${idx}`;
-    }
-    const orderedStep = unorderedSteps.find(step => step.name === findName);
+
+  const unnamedSteps = unorderedSteps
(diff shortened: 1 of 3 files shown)


a-roberts:
All good here, thanks for this one @AlanGreene, tested with my own TaskRuns too and not seeing any breakage, let's get this into 0.6.1.

/lgtm
/approve

Issue #1319, opened 2020-04-22
The ServiceAccount field in Import Tekton resources from repository is no longer optional


rnapoles-tktn:
**Problem:
The ServiceAccount field in Import Tekton resources from repository is identified as optional:
![image.png](https://images.zenhubusercontent.com/5db346d5b0111c0001df287a/68608b98-4134-4762-b33b-5cbb7b4e289f)

If you do NOT select a Service Account, then the corresponding pipelineRun will fail with:
`error: the path "/workspace/git-source/<APPLY-DIRECTORY>" does not exist`

**Versions:**
OCP 4.3, and 4.4-RC9
Dashboard/webhooks v0.5.3
Pipeline v0.10.1

**Expected:**
Should we be able to leave ServiceAccount empty?
If this is now mandatory then remove"(Optional)" from "ServiceAccount (Optional)" in the UI
Validate that a service account was selected
Should we be able to select multiple service accounts?

a-roberts:
Yeah this is an interesting one, my guess is when no SA is specified it'll try default which won't have permissions, or pipeline if installed through the operator

I propose removing the optional part anyway because why would you want to be importing resources without knowledge of the service account that'll be used. For the short term I've added a note in the 0.6.1 release about this.

a-roberts:
https://github.com/tektoncd/dashboard/pull/1320#issuecomment-619933247 latest discussion here, think we should hold off on this one until there's consensus, @eddycharly makes a good point. Added to the WG agenda for today.

rhd-gitops-functional:
Had a ponder over this

```
error: the path "/workspace/git-source/<APPLY-DIRECTORY>" does not exist
```

so that's likely due to permission problems on the default service account that gets used on OpenShift, but in theory a user could configure this in such a way for it to work I'm sure. So the optionalness really is a thing (and you don't need to provide one on Docker Desktop).

I'm wondering if we instead do nothing on this, or perhaps add a tooltip - something to make it more obvious what's going on.

@AlanGreene @eddycharly your thoughts? It's on the agenda today but happy to discuss here/update if you folks can't make it.

a-roberts:
Had a ponder over this

```
error: the path "/workspace/git-source/<APPLY-DIRECTORY>" does not exist
```

so that's likely due to permission problems on the default service account that gets used on OpenShift, but in theory a user could configure this in such a way for it to work I'm sure. So the optionalness really is a thing (and you don't need to provide one on Docker Desktop).

[...]

PR #1320, opened 2020-04-23
Make ServiceAccount mandatory for CreatePipelineRun & import

a-roberts:
As seen https://github.com/tektoncd/dashboard/issues/1319, it's mandatory for certain platforms and not for others. Instead of adding checks (e.g. if platform X, make it optional), let's be consistent. I don't think users should be creating PIpelineRuns or importing resources not knowing the ServiceAccounts they'll run as...

# Changes

Removes the optional text

# Submitter Checklist

review them:

for more details._


a-roberts:
Needs to be made to actually work so nothing to see here yet folks

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -481,6 +481,24 @@ class CreatePipelineRun extends React.Component {
                 disabled={this.isDisabled()}
                 onChange={this.handlePipelineChange}
               />
+              <ServiceAccountsDropdown
+                id="create-pipelinerun--sa-dropdown"
+                titleText={intl.formatMessage({
+                  id: 'dashboard.createPipelineRun.serviceAccountLabel',
+                  defaultMessage: 'ServiceAccount to run under'
```
AlanGreene, line 488: Was there a particular reason for changing the label here? It looks a little odd since the others all just name the field. We could add `helperText` if needed like we have on the Import Resources page.

a-roberts:
> We'll need to add validation since these fields are now required. I shouldn't be able to submit the forms without selecting a value (similar to Namespace).
> 
> On the Create PipelineRun form the ServiceAccount dropdown is hidden (along with Namespace + Pipeline) when clicking Create from a filtered view. We should move the ServiceAccount dropdown outside that FormGroup so it's not hidden.
> 
> Is the helper text on the Import Resources ServiceAccount dropdown still valid?
> 
> > "The ServiceAccount that the PipelineRun applying resources will run under (must be in the install namespace of the Tekton Dashboard)"
> 
> Should we also move the ServiceAccount dropdown on the Import Resources page up above the optional directory input? That way all of the required fields are presented first.

Mb, had totally slipped my mind that they're actually mandatory but only, effectively, once the user kicks off the flow (so the input is still not checking). I've had a look and noticed form validation for the create PipelineRun and import resources pages, so I'll add the checks there. Reverted the label change too, thanks for taking a look

a-roberts:
I've put this back into draft stage, added some of my "thoughts as code" so far, still have to make sure everything works and fix tests - looking to resume on Monday as something's came up today

a-roberts:
Working through 

```
  ● CreatePipelineRun › handles api error with text

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 1
    Received number of calls: 0
```

tests next

eddycharly:
Isn’t it clashing with the default-service-account feature in pipelines ?

Making the service account mandatory would break the feature.

a-roberts:
> Isn’t it clashing with the default-service-account feature in pipelines ?
> 
> Making the service account mandatory would break the feature.

Yeah, that's a good point actually, suppose it is a bit of a `feature`!

The rationale here was to remove the `optional` part simply because things are inconsistent across platforms, I'll keep this one on hold for now and welcome future discussion on how best to tackle https://github.com/tektoncd/dashboard/issues/1319.

PR #1321, opened 2020-04-23
Add panelling to create TaskRuns

### LINKED ISSUE #1247: Feature request - panelling for creating TaskRuns too
@mattmoor raised an interesting point here: https://twitter.com/mattomata/status/1246824264210112515?s=21

# Expected behavior
Users should be able to create TaskRuns too

# Actual behavior
One cannot

# Additional Info
Should be a low-hanging fruit with a clear benefit and we can reuse plenty of code for creating PipelineRuns, changing out the resource types. They'd be rerunnable too just like PIpelineRuns.

JustinKuli:
This allows users to create individual TaskRuns. From the TaskRuns
page, the modal will prompt the user for all of the required details.
If the user landed on the page by selecting a specific ClusterTask or
Task, then the modal will not prompt for the specification of the Task
itself (similar to the situation with PipelineRuns).

Most of this is based off of the existing CreatePipelineRun modal.
Major differences include the Kind (Task/ClusterTask), and that
PipelineResources are either inputs or outputs.

Closes #1247

# Changes

In addition to the new panel, routing from the ClusterTasks page needed to be adjusted slightly so that the new modal would correctly recognize the `type` of the Task/ClusterTask.

# Submitter Checklist

review them:

for more details._


JustinKuli:
This is a huge chunk of code... I couldn't find a good way to make it any smaller for feedback. Things are working as far as I can tell, but I wouldn't be surprised if there is something I missed.

packages/utils/src/utils/router.js
```diff
@@ -118,7 +118,7 @@ export const paths = {
       });
     },
     byClusterTask() {
-      return '/taskruns?labelSelector=tekton.dev%2Ftask%3D:taskName';
+      return '/taskruns?kind=clusterTask&labelSelector=tekton.dev%2Ftask%3D:taskName';
```
AlanGreene, line 121: This raises an interesting issue, the label on ClusterTasks should probably be `tekton.dev/clustertask` rather than `tekton.dev/task` so we can differentiate between runs of a Task and ClusterTask with the same name.

Edit: actually looks like this was changed recently in pipelines... https://github.com/tektoncd/pipeline/issues/2377
From now on labels are covered by the API deprecation policy.

@a-roberts fyi. I'll take another look at this tomorrow, not sure if it's included in 0.11.2 or if it'll be 0.12 before we get this

src/containers/CreateTaskRun/CreateTaskRun.js
```diff
@@ -0,0 +1,694 @@
...
+const labelKeyRegex = new RegExp(
+  '^(([a-z0-9A-Z]([a-z0-9A-Z-.]*[a-z0-9A-Z])?){0,253}/)?([a-z0-9A-Z]([a-z0-9A-Z-_.]*[a-z0-9A-Z])?){1,63}$'
+);
+const labelValueRegex = new RegExp(
+  '^([a-z0-9A-Z]([a-z0-9A-Z-_.]*[a-z0-9A-Z])?){0,63}$'
+);
+const isValidLabel = (type, value) => {
+  const regex = type === 'key' ? labelKeyRegex : labelValueRegex;
+  return regex.test(value);
+};
```
AlanGreene, line 108: we should extract this to src/utils so it can be reused across the 2 create run pages and ensure they're consistent

src/containers/CreateTaskRun/CreateTaskRun.scss
```diff
@@ -0,0 +1,30 @@
+/*
```
AlanGreene, line 1: we could move src/containers/CreatePipelineRun/CreatePipelineRun.scss to src/scss/CreateRun.scss and reuse that instead of duplicating the styles here

packages/utils/src/utils/router.js
JustinKuli: Good find! We could add `tekton.dev/clusterTask` to TaskRuns created from ClusterTasks in our API... but I think that TaskRuns created as part of a PipelineRun would be unaffected until we get the version of pipeline which does this. 

Also, TaskRuns created from ClusterTasks before this change wouldn't be listed on the TaskRuns page if we started using this label as a filter.

The point of this workaround is just to get the fact that it's a ClusterTask to the `CreateTaskRun` modal... I'm open to any suggestions on a better way to do that, since putting it in the URL doesn't feel ideal.

AlanGreene: I looked into this some more and it's actually been updated so the old label can be removed over multiple releases as per our API deprecation policy. This means that on pipelines master, TaskRuns referencing a ClusterTask now get both labels applied. The `tekton.dev/task` label on these runs will be marked as deprecated in the next (?) release, and removed in a subsequent release.

We still need to figure out how we're going to handle this in the dashboard.

I propose we do something like this:
1. update the URL to use the new label for filtering (replacing the current one) from the ClusterTasks page
  i. this includes updating the TaskRuns container's `mapStateToProps` to replace the old label with the new one
2. use the presence of the new label in the URL to determine which `kind` to use for the Create UI
3. in the container's `fetchTaskRuns` detect the presence of the new label and make 2 calls to `props.fetchTaskRuns`, 1 for each of the labels
4. in the container's `mapStateToProps` update `taskRuns` to return the union of 2 calls to the selector, 1 for each of the labels

Once we update to a pipelines release that removes the old label we can just remove the redundant call from 3 and 4.

Open to other suggestions if there's a cleaner way to handle this with minimal impact.

JustinKuli: Made this adjustment. It might be easier to check just those changes by looking at this commit: https://github.com/JustinKuli/dashboard/commit/bafee1e75c886ed2d33262ae138d23971bcd719b

Let me know what you think. I had to adjust some other logic to get the `kind` to appear correctly... there might still be something buggy with how I'm handling that in props vs state.

AlanGreene:
Hi @JustinKuli, thanks for making those changes. I'm testing it out locally now. Can you resolve the merge conflict on this PR? That should also take care of the missing CLA status.

JustinKuli:
@AlanGreene, thanks, updated now.

By the way, do you have any suggestions for updating PRs but keeping it as one commit? I have something that is working with fetching, rebasing, resetting, committing and then force-pushing, but it feels like I'm probably missing a better method. I think this also makes reviewing changes more difficult. Should I just be squashing to one commit when the PR is done?

AlanGreene:
Looks like there's another conflict (I'm done making changes for today, I promise 😂)

Yes squashing at the end and tidying up the commit message is probably easiest for reviews, especially for larger changes.

JustinKuli:
Don't be done making changes! (Merge mine first ;) )

JustinKuli:
@AlanGreene anything else required here?

PR #1325, opened 2020-04-26
ImportResources : Add dropdown for install namespace

eddycharly:
This PR adds a drop down for install namespace selection in the ImportResources container.

Now that the import resources pipeline run is not dependent of a pipeline/task installed in the same namespace as the dashboard, this is not a requirement anymore that the pipeline run executes in the namespace where the dashboard was installed.

This allows to select the namespace where the pipeline run executes and therefore a service account in the selected install namespace.

**NOTE**: install namespace is somewhat unclear as it is not related to dashboard install but i couldn't find a better name. Feel free to suggest a better name.

### DIFF
--- a/src/containers/ImportResources/ImportResources.js
+++ b/src/containers/ImportResources/ImportResources.js
@@ -14,11 +14,15 @@ limitations under the License.
 import React, { Component } from 'react';
 import { injectIntl } from 'react-intl';
 import {
+  Accordion,
+  AccordionItem,
   Button,
   Form,
+  FormGroup,
   InlineNotification,
   TextInput,
-  ToastNotification
+  ToastNotification,
+  Tooltip
 } from 'carbon-components-react';
 
 import { connect } from 'react-redux';
@@ -66,8 +70,9 @@ export class ImportResources extends Component {
       directory: '',
       invalidInput: false,
       invalidNamespace: false,
-      installNamespace: '',
-      installNamespaceError: false,
+      invalidImporterNamespace: false,
+      importerNamespace: '',
+      importerNamespaceError: false,
       logsURL: '',
       namespace: props.navNamespace !== ALL_NAMESPACES && props.navNamespace,
       repositoryURL: '',
@@ -79,14 +84,14 @@ export class ImportResources extends Component {
 
   componentDidMount() {
     getInstallNamespace()
-      .then(foundInstallNamespace => {
+      .then(foundImporterNamespace => {
         this.setState({
-          installNamespace: foundInstallNamespace
+          importerNamespace: foundImporterNamespace
         });
       })
       .catch(() => {
         this.setState({
-          installNamespaceError: true
+          importerNamespaceError: true
         });
       });
   }
@@ -104,6 +109,17 @@ export class ImportResources extends Component {
       this.setState({ invalidNamespace: false, namespace: selectedItem.id });
     } else {
       this.setState({ invalidNamespace: true, namespace: '' });
+    }
+  };
+
+  handleImporterNamespace = ({ selectedItem }) => {
+    if (selectedItem) {
+      this.setState({
+        invalidImporterNamespace: false,
+        importerNamespace: selectedItem.id
+      });
+    } else {
+      this.setState({ invalidImporterNamespace: true, importerNamespace: '' });
     }
   };
 
@@ -128,7 +144,7 @@ export class ImportResources extends Component {
       namespace,
       repositoryURL,
       serviceAccount,
-      installNamespace
+      importerNamespace
     } = this.state;
 
     // Without the if statement it will not display errors for both the namespace and the url at the same time
@@ -159,20 +175,20 @@ export class ImportResources extends Component {
       namespace,
       labels,
       serviceAccount,
-      installNamespace
+      importerNamespace
     })
       .then(headers => {
         const pipelineRunName = headers.metadata.name;
 
         const finalURL = urls.pipelineRuns.byName({
-          namespace: installNamespace,
+          namespace: importerNamespace,
           pipelineRunName
         });
 
         this.setState(prevState => {
           return {
             logsURL:
-              prevState.installNamespaceError === false
+              prevState.importerNamespaceError === false
                 ? finalURL
                 : urls.pipelineRuns.all(),
             submitSuccess: true,
@@ -194,11 +210,18 @@ export class ImportResources extends Component {
 
   render() {
     const { intl } = this.props;
-    const { namespace } = this.state;
+    const { namespace, importerNamespace } = this.state;
     const selectedNamespace = namespace
       ? {
           id: namespace,
           text: namespace
+        }
+      : undefined;
+
+    const selectedImporterNamespace = importerNamespace
+      ? {
+          id: importerNamespace,
+          text: importerNamespace
         }
       : undefined;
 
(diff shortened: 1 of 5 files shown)


a-roberts:
Nice catch @eddycharly 👍 definitely an improvement and reduces some complexity/confusion here

eddycharly:
Basically, there are two namespace parameters:
- `install namespace` is the one where the pipeline run importing the resources will execute
- `namespace` the target namespace where resources will be imported

None of these namespaces are related to the dashboard installation namespace though, and `install namespace` drives the selection of the service account used for pipeline run execution and consequently the github token used for checking out the repository.

The `namespace` and `install namespace` names are confusing. It's not naturally easy to understand their purpose based on the actual names.

Maybe `namespace` could be renamed to `target namespace`. For `install namespace` i don't have a better name at hand but i feel like it should reflect the fact that it's related to service account selection.

a-roberts:
> Basically, there are two namespace parameters:
> 
> * `install namespace` is the one where the pipeline run importing the resources will execute
> * `namespace` the target namespace where resources will be imported
> 
> None of these namespaces are related to the dashboard installation namespace though, and `install namespace` drives the selection of the service account used for pipeline run execution and consequently the github token used for checking out the repository.
> 
> The `namespace` and `install namespace` names are confusing. It's not naturally easy to understand their purpose based on the actual names.
> 
> Maybe `namespace` could be renamed to `target namespace`. For `install namespace` i don't have a better name at hand but i feel like it should reflect the fact that it's related to service account selection.

Yeah, clarified what I meant and added a suggestion too, looks like we're both in-sync regarding the target namespace bit too 👍

eddycharly:
An idea, maybe the form could be reworked so that the 2 set of parameters appear more separated making it clear that some parameters belong to the imported resources and other parameters belong to importing process execution ?

a-roberts:
> An idea, maybe the form could be reworked so that the 2 sets of parameters appear more separated making it clear that some parameters belong to the imported resources and other parameters belong to importing process execution ?
> 
> In this case, both parameters could be labelled `Namespace` without being confusing as they would refer to different concerns.

Yeah, I think that's a good idea. So they'd be separated (perhaps using FormGroup) so they're kinda on two different panels but as part of the same main panel - that would make it clearer.

eddycharly:
Looks better ?

image

a-roberts:
> Looks better ?
> 
> image

Yep, much better, now we just need to work out that heading so it's easy enough to follow, separating them is definitely an improvement I say.

For the header, I know what you mean by it but I'm wondering if a beginner user would. Perhaps we default to the Tekton Dashboard ServiceAccount and install namespace and make this an "Advanced configuration for the Import PipelineRun" 🤔

eddycharly:
> Perhaps we default to the Tekton Dashboard ServiceAccount
IMO tekton dashboard service account shouldn't be linked to any github token, it should not be in the scope of its responsibilities.

Ok for the header, i need to learn how to add a tooltip though 😅

a-roberts:
> > Perhaps we default to the Tekton Dashboard ServiceAccount
> 
> IMO tekton dashboard service account shouldn't be linked to any github token, it should not be in the scope of its responsibilities.
> 
> Ok for the header, i need to learn how to add a tooltip though 😅

Yep, agreed on the first point actually (I'm used to pulling from public repos not requiring a token but appreciate this isn't gonna be the same for everyone)

For the tooltip see if this is useful: https://github.com/tektoncd/dashboard/blob/master/src/containers/Secrets/Secrets.js#L398

eddycharly:
> For the tooltip see if this is useful: https://github.com/tektoncd/dashboard/blob/master/src/containers/Secrets/Secrets.js#L398

Thanks, it helped. That's the best i could achieve:
image

I am definitely not a frontend developer so it's probably not perfect, do not hesitate to comment if things can be improved.

a-roberts:
> > For the tooltip see if this is useful: https://github.com/tektoncd/dashboard/blob/master/src/containers/Secrets/Secrets.js#L398
> 
> Thanks, it helped. That's the best i could achieve:
> image
> 
> I am definitely not a frontend developer so it's probably not perfect, do not hesitate to comment if things can be improved.

Yep, is what I was thinking (although when I said it, it wasn't explicitly meant to be copied verbatim 😸) let's get @AlanGreene's input as well and then look at getting this merged, thanks for your work on this

eddycharly:
> although when I said it, it wasn't explicitly meant to be copied verbatim 😸

I guess I'm lacking creativity in this domain 😅

eddycharly:
ping @AlanGreene 
/meow

src/containers/ImportResources/ImportResources.js
```diff
@@ -107,6 +110,17 @@ export class ImportResources extends Component {
     }
   };
 
+  handleInstallNamespace = ({ selectedItem }) => {
```
AlanGreene, line 113: how about 'importerNamespace' instead of 'installNamespace' to avoid confusion with the Dashboard install namespace?

src/containers/ImportResources/ImportResources.js
```diff
@@ -225,47 +246,77 @@ export class ImportResources extends Component {
...
-          />
-          <ServiceAccountsDropdown
-            className="saDropdown"
-            helperText="The ServiceAccount that the PipelineRun applying resources will run under (must be in the install namespace of the Tekton Dashboard)"
-            id="import-service-accounts-dropdown"
-            namespace={this.state.installNamespace}
-            onChange={this.handleServiceAccount}
-            titleText="ServiceAccount (optional)"
-          />
+          <FormGroup legendText="Source repository and target namespace parameters">
```
AlanGreene, line 249:
```suggestion
          <FormGroup legendText="Source repository and target namespace">
```

avoids confusion with `params` / `parameters` used elsewhere to refer to Pipeline / Task input params.

src/containers/ImportResources/ImportResources.js
```diff
@@ -225,47 +246,77 @@ export class ImportResources extends Component {
...
+            <TextInput
+              data-testid="repository-url-field"
+              helperText="The location of the YAML definitions to be applied (Git URL's supported)"
```
AlanGreene, line 252:
```suggestion
              helperText="The location of the YAML definitions to be applied (Git URLs supported)"
```

src/containers/ImportResources/ImportResources.js
```diff
@@ -225,47 +246,77 @@ export class ImportResources extends Component {
...
+              <Tooltip
+                direction="right"
+                triggerText="Advanced configuration for the Import PipelineRun"
+              >
+                <div>
+                  Change these parameters if you want the PIpelineRun that will
+                  do the importing, to run in a different namespace from the
+                  Dashboard&apos;s.
+                  <br />
+                  You can optionally provide a different ServiceAccount too.
```
AlanGreene, line 296:
```suggestion
                  Change these parameters if you want the PipelineRun that will
                  do the importing to run in a different namespace from the
                  Dashboard&apos;s.
                  <br />
                  <br />
                  You can optionally provide a different ServiceAccount too.
```

src/containers/ImportResources/ImportResources.js
eddycharly: Sounds better

eddycharly:
All done.

Thinking about the service account again, in addition to hold the github token (in case of private a repository), the service account must also be allowed to create resources in the target namespace.

I fear that it won't be easy to understand the importance of this setting for newcomers. Especially as it's now in an advanced section, hidden by default.

Maybe we can merge this PR and create an issue to address this UX aspect ?

a-roberts:
```

@a-roberts did we mention on the WG call yesterday the possibility of hiding the advanced params by default? Something like this using an accordion perhaps:
```

yeah, exactly, looks good in your comment too. I'm going to pull in this PR now, give it a spin and give it the lgtm hopefully

src/containers/ImportResources/ImportResources.js
```diff
@@ -225,47 +248,80 @@ export class ImportResources extends Component {
...
+          </FormGroup>
+          <Accordion>
+            <AccordionItem
+              title={
+                <Tooltip
+                  direction="right"
+                  triggerText="Advanced configuration for the Import PipelineRun"
+                >
+                  <div>
+                    Change these parameters if you want the PIpelineRun that
```
a-roberts, line 295: Everything was perfect, until this capital I in PipelineRun 😆 

Everything else looks great, thanks for this, much better, I know it's a nit but it's gonna bug me forever

eddycharly: haha, i change that right now

a-roberts:
/lgtm
/meow hats

Nice work 🥳

PR #1327, opened 2020-04-28
Add recommended k8s labels

eddycharly:
This PR adds recommended k8s labels on resources.
Implementation for proposal https://github.com/tektoncd/pipeline/issues/2497

Related PR in pipelines: https://github.com/tektoncd/pipeline/pull/2501

Base: added labels and updated selectors for all resources in `/base`
Go: changed go code in `/pkg/endpoints/dashboard.go`

**NOTES**:
- I removed some old code that implemented various strategies to find dependent applications version. I don't think it's an issue because these versions are old and should not work with the current dashboard (pre v1beta1).
- I didn't touch the code in `/pkg/endpoints/cluster.go`. The code there is bound to ingresses which is not part of this repository. I will address this code in another PR.
- This will introduce a breaking change when trying to update deployments as label selectors have changed.

### DIFF
--- a/base/300-deployment.yaml
+++ b/base/300-deployment.yaml
@@ -5,18 +5,31 @@ metadata:
   name: tekton-dashboard
   namespace: tekton-pipelines
   labels:
+    app.kubernetes.io/name: dashboard
+    app.kubernetes.io/component: dashboard
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/version: devel
+    app.kubernetes.io/part-of: tekton-dashboard
+    dashboard.tekton.dev/release: "devel"
     app: tekton-dashboard
     version: "devel"
-    dashboard.tekton.dev/release: "devel"
 spec:
   replicas: 1
   selector:
     matchLabels:
-      app: tekton-dashboard
+      app.kubernetes.io/name: dashboard
+      app.kubernetes.io/component: dashboard
+      app.kubernetes.io/instance: default
+      app.kubernetes.io/part-of: tekton-dashboard
   template:
     metadata:
       name: tekton-dashboard
       labels:
+        app.kubernetes.io/name: dashboard
+        app.kubernetes.io/component: dashboard
+        app.kubernetes.io/instance: default
+        app.kubernetes.io/version: devel
+        app.kubernetes.io/part-of: tekton-dashboard
         app: tekton-dashboard
     spec:
       serviceAccountName: tekton-dashboard
--- a/base/200-clusterrole.yaml
+++ b/base/200-clusterrole.yaml
@@ -4,6 +4,10 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: tekton-dashboard-minimal
   namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/component: dashboard
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-dashboard
 rules:
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
--- a/base/201-clusterrolebinding.yaml
+++ b/base/201-clusterrolebinding.yaml
@@ -3,6 +3,10 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: tekton-dashboard-minimal
+  labels:
+    app.kubernetes.io/component: dashboard
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-dashboard
 subjects:
   - kind: ServiceAccount
     name: tekton-dashboard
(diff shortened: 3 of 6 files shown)


eddycharly:
Thanks for testing it !

It looks like tekton.dev/release label is missing in the list.

I would be happy if you can test this on openshift as I don’t know much about it.
Can testing be done with minishift ?

a-roberts:
> Thanks for testing it !
> 
> It looks like tekton.dev/release label is missing in the list.
> 
> I would be happy if you can test this on openshift as I don’t know much about it.
> Can testing be done with minishift ?

Possibly, it's worth giving it a try yeah, once you've got something we know works ok for the non-OpenShift environments I'm happy to go through the matrix 😄

eddycharly:
@a-roberts I added support for the `tekton.dev/release` label, it works fine with pipelines `v0.10.2`.

image

a-roberts:
> @a-roberts I added support for the `tekton.dev/release` label, it works fine with pipelines `v0.10.2`.
> 
> image

Thanks, won't have time to test this in-depth today but if you can verify what you can (e.g. on Minishift) that'd really help, rest of the PR's looking good, we'll want to make sure nothing breaks for Triggers (0.3.1 and later) either 👍

eddycharly:
I tested it successfully with:
- pipelines 0.10.0, 0.10.1, 0.10.2, 0.11.0, 0.11.1, 0.11.2, 0.11.3, 0.12.0
- triggers 0.3.0, 0.3.1, 0.4.0

I will try to test with minishift during the weekend if I have some free time.

a-roberts:
> I tested it successfully with:
> 
> * pipelines 0.10.0, 0.10.1, 0.10.2, 0.11.0, 0.11.1, 0.11.2, 0.11.3, 0.12.0
> * triggers 0.3.0, 0.3.1, 0.4.0
> 
> I will try to test with minishift during the weekend if I have some free time.

Thank you! No worries if you can't I'll give it a good go on OpenShift Monday anyway, that's great 😄

eddycharly:
It worked with minishift for:
- pipelines: 0.10.1 and 0.10.2
- triggers: 0.3.0 and 0.3.1

Strangely it didn't work with pipelines 0.10.0.

I couldn't test other versions as they require a k8s cluster 1.15.

a-roberts:
Had a go with the latest operator for installing Tekton on OpenShift (openshift-pipelines-operator) today, devel appears as the version for both the dev channel and canary releases, here's the pipeline controller yaml so you can see the label used

```
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: tekton-pipelines-controller
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      annotations:
        cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
        tekton.dev/release: devel
      creationTimestamp: null
      labels:
        app: tekton-pipelines-controller
        app.kubernetes.io/component: controller
        app.kubernetes.io/name: tekton-pipelines
        pipeline.tekton.dev/release: v0.11.3
        version: v0.11.3
```

The same thing happens with dev-channel or canary, so we could augment this PR to check for pipeline.tekton.dev/release as well, or just labels.version.

@eddycharly you're welcome to split this into PRs if you'd prefer (so labelling our Dashboard stuff accordingly, versus the lookup changes)

For Triggers, it's installing 0.2.1 currently which has no awesome labels for us to lookup, but will assume it'll be of the same pattern (@vdemeester ^^)

eddycharly:
Boouh, we need to clean this but keeping the backward compatibility is a mess 😢 

Alright, i will spit the PRs, it should help moving forward.

a-roberts:
> Boouh, we need to clean this but keeping the backward compatibility is a mess 😢
> 
> Alright, i will spit the PRs, it should help moving forward.

Agreed, thanks 😸

eddycharly:
@a-roberts i removed the lookup related changes, the PR should be all about labels now

base/300-deployment.yaml
```diff
@@ -5,18 +5,31 @@ metadata:
...
+      app.kubernetes.io/component: dashboard
+      app.kubernetes.io/instance: default
+      app.kubernetes.io/part-of: tekton-dashboard
   template:
     metadata:
       name: tekton-dashboard
       labels:
+        app.kubernetes.io/name: dashboard
+        app.kubernetes.io/component: dashboard
+        app.kubernetes.io/instance: default
```
a-roberts, line 30: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#applications-and-instances-of-applications

Why `default` for instance and not dashboard? All other changes look good

eddycharly: The `app.kubernetes.io/instance` label shouldn't express the application itself but the instance of the application, i feel like these manifests express the `default` deployment of the `dashboard` application.

Also it's more consistent, i did the same thing for `pipelines` and `triggers` (setting `app.kubernetes.io/instance` to `default`). I guess it will be easier to remember when trying to lookup a resource.

eddycharly: For example, when deploying with helm i set the value to the release name : https://github.com/tektoncd/experimental/pull/528/files#diff-718e9764773438f64515998a5c46e5a1R50

a-roberts: Got it, thanks for the clarifications

PR #1336, opened 2020-04-30
Add OpenShift support for release pipeline

a-roberts:
# Changes

This PR includes the changes I had to make to get the release pipeline working on OpenShift instead of Docker Desktop.

Noticably:
- `scc` commands for permissions then documenting a configured namespace and SA
- Omitting the lockdown step as this uses Docker (relying on a Docker socket)
- Docs

Outstanding:
- [x] Check no RBAC needs to be added
- [ ] See if we can reintroduce the lockdown step by perhaps using a buildah client if that exists, or some other way (like, bud pulling, getting that tag with sed, using that). Update: doesn't exist, podman looks to be using a Docker socket too, and you can't even docker pull on OCP 4.3 (mentions the socket too)
- [ ] See if we can reduce Tekton Task duplication/effort (testing now with removed use of /previous as that was where the pinned images ended up for us to stage)
- [x] Ensure resources.yaml doesn't reference a given bucket location 😄 
- [x] Check both oc permissions are required - I'm convinced they are, it'll run as root

We'll want to do the same for the webhooks extension then. I'd like to do this so we have another route to do releases on - I'd often like to use Docker Desktop for other things and it's handy to have the extra platform support incase someone wants to do a release and doesn't have Docker Desktop.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/README.md
+++ b/tekton/README.md
@@ -108,12 +108,57 @@ tkn pipeline start dashboard-release -p versionTag=$VERSION_TAG -r dashboard-sou
 
 Monitor the build logs to see the image coordinates that the image is pushed to. The release yaml files should appear under https://console.cloud.google.com/storage/browser/tekton-releases/dashboard.
 
+## Running the release pipeline on OpenShift
+
+Decide which namespace and ServiceAccount you'll use for this process, an example is provided below.
+
+`oc new-project dashboard-release-pipeline`
+
+```
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: release-pipeline-sa
+  namespace: dashboard-release-pipeline
+```
+
+Run the following commands to get around permission denied problems (as the build step uses `sudo`).:
+
+```
+oc adm policy add-scc-to-user privileged -z release-pipeline-sa -n dashboard-release-pipeline
+oc adm policy add-scc-to-user anyuid -z release-pipeline-sa -n dashboard-release-pipeline
+```
+
+Next, install the Tekton resources from the main `tekton` folder.
+
+- `oc apply -f tekton`
+
+Then, install the Tekton resources from the `openshift` folder:
+
+- `oc apply -f tekton/openshift`
+
+- Create your release secret and all Tekton resources in the namespace you've chosen earlier, e.g. `dashboard-release-pipeline`:
+
+```
+KEY_FILE=$HOME/googlekey/release.json
+GENERIC_SECRET=release-secret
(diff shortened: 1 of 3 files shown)


a-roberts:
@mnuttall i've removed the references to previous and the use of lockdown.py here, see in the PR comments above what I've found with the pinning of images, It'd be useful to get this in anyway I think simply because it does give us a second platform which we can do releases on.

There is some duplication of code here in the publish.yml, I'm wondering if we should instead have two different Pipelines and split up everything that's a step, into Tasks (we'd then have two very similar pipelines though, without steps, unless we introduce a parameter to do the lockdown.py business optionally). Lemme know your thoughts either way please, I have a working publish using my OCP 4.3 cluster to https://console.cloud.google.com/storage/browser/tekton-releases/dashboard/openshifttest2/mytestbucket/?project=tekton-releases

a-roberts:
Boooo

```

--------------------------------
---- Linting the yaml files ----
--------------------------------
found tekton/openshift/publish.yaml
found tekton/resources.yaml
tekton/openshift/publish.yaml
  136:35    error    no new line character at the end of file  (new-line-at-end-of-file)
```

I'll fix this in a sec, checking another PR on a different cluster...

tekton/README.md
```diff
@@ -108,12 +108,57 @@ tkn pipeline start dashboard-release -p versionTag=$VERSION_TAG -r dashboard-sou
...
+
 ## Manually complete the release work
 
 We have a number of tasks that are yet to be automated:
 
 - Write the release notes
 - Attach `.yaml` files from https://console.cloud.google.com/storage/browser/tekton-releases/dashboard - be sure you copy the locked down image ones (look under `previous`): any containers such as `kubectl` and `oauth-proxy` should reference an image sha and not a tag such as `latest`
-- Optionally repeat for the Webhooks Extension (automation in progress)
-- Fix up image coordinates in `/README.md` for the normal and Openshift installs
+- Note that the image pinning, if doing the release on OpenShift, has not yet been implemented - so you'll have to do this manually until then
```
mnuttall, line 161: Please can we link to a follow on issue to sort this out? There must be a way to port  https://github.com/tektoncd/dashboard/blob/master/tekton/scripts/lockdown.py#L49 to openshift, be it docker-in-docker, or `curl`-based magic, or something else.

a-roberts: Yeah, I'll link to https://github.com/tektoncd/dashboard/issues/1384

a-roberts: Done

PR #1342, opened 2020-05-01
Use ClusterRole aggregation for standalone extensions

eddycharly:
# Changes

This PR changes the way the `tekton-dashboard` `ClusterRole` is constructed by allowing [ClusterRole aggregation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles).

This way, extensions can create the `ClusterRole` they need to work, the additional permissions will automatically be aggregated on the `tekton-dashboard` `ClusterRole`.

This makes extensions more standalone.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/base/200-clusterrole.yaml
+++ b/base/200-clusterrole.yaml
@@ -3,11 +3,24 @@ kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: tekton-dashboard-minimal
-  namespace: tekton-pipelines
   labels:
     app.kubernetes.io/component: dashboard
     app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-dashboard
+aggregationRule:
+  clusterRoleSelectors:
+    - matchLabels:
+        rbac.dashboard.tekton.dev/aggregate-to-dashboard: "true"
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: tekton-dashboard-core
+  labels:
+    app.kubernetes.io/component: dashboard
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-dashboard
+    rbac.dashboard.tekton.dev/aggregate-to-dashboard: "true"
 rules:
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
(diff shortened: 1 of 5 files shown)


base/200-clusterrole.yaml
```diff
@@ -3,11 +3,24 @@ kind: ClusterRole
...
   labels:
     app.kubernetes.io/component: dashboard
     app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-dashboard
+aggregationRule:
+  clusterRoleSelectors:
+    - matchLabels:
+        rbac.authorization.k8s.io/aggregate-to-dashboard: "true"
+---
+kind: ClusterRole
```
AlanGreene, line 15: Can these rules not be kept as part of the `tekton-dashboard-minimal` ClusterRole above instead of splitting them out?

eddycharly: I don't think so, aggregating overwrites the rules.

AlanGreene: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles

> The aggregationRule defines a label selector that the controller uses to match other ClusterRole objects that should be combined into the rules field of this one.

I'll try this out locally but the wording above suggests it should merge them.

eddycharly: I tried it and the result was the rules were overwritten.

I think the wording is confusing, the matching ClusterRoles rules are merged but the receiver rules are not.

I guess it makes sense as the object stored in etcd is the result of the aggregation, the original rules of the receiver are not preserved.

AlanGreene: Thinking about this some more it does make sense, the wording is just a little ambiguous. 'combined into the rules field' in this case means the ClusterRole objects matching the label are combined and the result then used to populate a rules field for this ClusterRole.

AlanGreene:
`rbac.authorization.k8s.io/aggregate-to-*` labels seem to be used to aggregate to some of the default roles.

Can we change the label to `rbac.dashboard.tekton.dev/aggregate-to-dashboard` so it's clear we're not aggregating to any of the default roles?

base/200-clusterrole.yaml
AlanGreene: hrmm now also wondering if the full-fat kustomize overlay falls foul of this

eddycharly:
I renamed the label to `rbac.dashboard.tekton.dev/aggregate-to-dashboard`

a-roberts:
Looks all good 👍 let's get this fully tested out on OpenShift as well to check nothing strange happens then before getting it in, @Megan-Wright has kindly volunteered

eddycharly:
/meow

Is there anything left to validate ?

PR #1343, opened 2020-05-02
Use getInstallProperties for displaying logout

eddycharly:
# Changes

This PR changes the way `shouldDisplayLogout` is implemented.

Instead of calling an api that will succeed on openshift and fail when running on k8s, it reuses the `getInstallProperties` api and uses the `IsOpenShift` property from the results.

This prevents a call to an api that doesn't exist and moves the logic of determining the condition to the backend rather than doing this on the frontend side.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -431,11 +431,7 @@ export function getInstallProperties() {
 }
 
 export function shouldDisplayLogout() {
-  const routesUri = getResourcesAPI({
-    group: 'route.openshift.io',
-    version: 'v1'
-  });
-  return get(routesUri, { Accept: 'text/plain' });
+  return getInstallProperties().then(data => data.IsOpenShift);
 }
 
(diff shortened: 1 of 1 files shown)


eddycharly:
Maybe we should add a LogoutUrl in the properties so that it’s more driven by the backend and doesn’t couple logout with openshift on the frontend.

Right now it’s the same property for different semantics.

a-roberts:
> Maybe we should add a LogoutUrl in the properties so that it’s more driven by the backend and doesn’t couple logout with openshift on the frontend.
> 
> Right now it’s the same property for different semantics.

Yeah, that would be useful, right now we currently only support piggybacking off oauth-proxy as the login provider (so the link points to that) but decoupling it so you could log out from another provider would be useful too

eddycharly:
I'll do that in a follow up.

PR #1348, opened 2020-05-04
Decouple logout button and openshift

eddycharly:
# Changes

This PR changes the logic behind the logout button.
This a follow up for this PR https://github.com/tektoncd/dashboard/pull/1343

Instead of hardcoding the logout url and using the `IsOpenshift` property to show/hide the button, the backend returns a `LogoutUrl` property.

The `LogoutUrl` property is then used to show/hide the logout button (the button is not rendered when the value is not set).

/cc @a-roberts

### DIFF
--- a/packages/components/src/components/LogoutButton/LogoutButton.js
+++ b/packages/components/src/components/LogoutButton/LogoutButton.js
@@ -17,36 +17,37 @@ import { injectIntl } from 'react-intl';
 import React, { Component } from 'react';
 import './LogoutButton.scss';
 
-/* istanbul ignore next */
-function handleLogout() {
-  window.location.href = '/oauth/sign_out';
-}
-
 export class LogoutButton extends Component {
   state = {
-    showLogout: false
+    logoutURL: null
   };
 
   componentDidMount() {
-    this.determineOpenShift();
+    this.determineLogoutURL();
   }
 
-  async determineOpenShift() {
+  handleLogout = () => {
+    const { logoutURL } = this.state;
+    window.location.href = logoutURL;
+  };
+
+  async determineLogoutURL() {
     try {
-      const showLogout = await this.props.shouldDisplayLogout();
-      this.setState({ showLogout });
+      const logoutURL = await this.props.getLogoutURL();
+      this.setState({ logoutURL });
     } catch (error) {} // eslint-disable-line
   }
 
   render() {
-    if (!this.state.showLogout) {
+    const { logoutURL } = this.state;
+    if (!logoutURL) {
       return null;
     }
     return (
       <HeaderGlobalAction
         data-testid="logout-btn"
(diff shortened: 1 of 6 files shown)


packages/components/src/components/LogoutButton/LogoutButton.js
```diff
@@ -18,35 +18,36 @@ import React, { Component } from 'react';
...
+    const { logoutUrl } = this.state;
+    if (!logoutUrl) {
       return null;
     }
     return (
       <HeaderGlobalAction
         data-testid="logout-btn"
         className="tkn--logout-btn"
-        onClick={handleLogout}
+        onClick={() => handleLogout(logoutUrl)}
```
AlanGreene, line 50: we can define `handleLogout` on the LogoutButton class and avoid creating a new anonymous function on each render here:

`onClick={this.handleLogout}`

and

```js
handleLogout = () => {
  const { logoutURL } = this.state;
  window.location.href = logoutURL;
};

render() {
...
```

eddycharly:
> I wonder if we should go a step further with this change, and instead of still having the hardcoded logout URL, could we patch it onto the Dashboard deployment as an environment property (similar to the 'read only' and installed namespace properties) when laying down the oauth-proxy config?

Pobably yes, right now it is bound to the `isOpenshift` variable.

Would it be ok to do this in another PR ?

AlanGreene:
> Would it be ok to do this in another PR ?
Yes that would be fine, this change gets us a step closer to having it fully decoupled.

eddycharly:
@AlanGreene are you ok to merge this ?
It will allow me to go ahead with the About page changes.

AlanGreene:
If you can update the remaining instances of 'Url' to 'URL' this will be ready to merge.

eddycharly:
My bad, i forgot some instances.
It should be good now 🤞

AlanGreene:
hrmm just gave this one last check locally and I'm not seeing TriggersVersion being returned from the `/v1/properties` API so it's not showing up on the About page.

pkg/endpoints/cluster.go
```diff
@@ -195,13 +195,24 @@ func (r Resource) GetProperties(request *restful.Request, response *restful.Resp
...
+		IsOpenShift:      isOpenShift,
+		ReadOnly:         isReadOnly,
+	}
+
+	// If running on OpenShift, set the logout url
+	if isOpenShift {
+		properties.LogoutURL = "/oauth/sign_out"
+	}
+
 	if isTriggersInstalled == false {
```
eddycharly: Wrong test 😢

eddycharly: If you can try again... sorry for the inconvenience

AlanGreene: no problem at all 😸

Issue #1255, opened 2020-04-07
Cross Site Request Forgery vulnerability


mnuttall:
Cross-Site Request Forgery (CSRF) is an attack that forces an end user to execute unwanted actions on a web application in which they're currently authenticated.
CSRF attacks specifically target state-changing requests, not theft of data, since the attacker has no way to see the response to the forged request. 

A specific CSRF attack is detailed in https://github.ibm.com/IBMCloudPak4Apps/icpa-security/issues/24

a-roberts:
I'll take a look, one suggested remedy is to:
 
> Generate a unique nonce for each form, place the nonce into the form, and verify the nonce upon receipt of the form. Be sure that the nonce is not predictable.

AlanGreene:
We need a nonce on API requests that modify the system, i.e. any POST / PUT / PATCH / DELETE requests. Not sure if go-restful has any CSRF support built in or if we'd need to look into gorilla/csrf or other packages. We should avoid trying to implement something ourselves if possible.

The nonce directive on the CSP header is to allow for exceptions to rules in specific cases, for example enabling specific known inline scripts to be executed when unsafe-inline is blocked. To do this, the script tag would have the nonce attribute added and the browser must very it matches the value in the header before executing its content.

a-roberts:
> We need a nonce on API requests that modify the system, i.e. any POST / PUT / PATCH / DELETE requests. Not sure if go-restful has any CSRF support built in or if we'd need to look into gorilla/csrf or other packages. We should avoid trying to implement something ourselves if possible.
> 
> The nonce directive on the CSP header is to allow for exceptions to rules in specific cases, for example enabling specific known inline scripts to be executed when unsafe-inline is blocked. To do this, the script tag would have the nonce attribute added and the browser must very it matches the value in the header before executing its content.

Not seeing any mentions of nonces or csrf protection for go-restful, we've put a lot of work into the current router set up so shifting away in favour of gorilla-mux (which is easy to use with gorilla/csrf) just for this wouldn't be ideal. 

gorilla/csrf seems reasonable enough to use, I've yet to figure out how to attach it our current set up though (snooping around in pkg/router/router.go looking at our restful/container objects)

a-roberts:
@AlanGreene's been running with this one using a simplified approach from what I had 🥳

PR #1351, opened 2020-05-05
Add CSRF protection

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1255

Wrap the routerHandler with gorilla/csrf to enable CSRF protection
on our APIs. This uses the 'double submit' method requiring both
a cookie and custom header on mutating requests.

A CSRF_SECURE_COOKIE environment variable is added to allow for
dev access over insecure connections. Default is "true" which sets
the Secure attribute on the anti-CSRF token cookie, meaning browsers
will not send this cookie over an insecure connection. We override
this in the dev overlays to set it to "false".

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/cmd/dashboard/main.go
+++ b/cmd/dashboard/main.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Tekton Authors
+Copyright 2019-2020 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -15,11 +15,14 @@ package main
 
 import (
 	"context"
+	"crypto/rand"
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
+	"github.com/gorilla/csrf"
 	routeclient "github.com/openshift/client-go/route/clientset/versioned"
 	"github.com/tektoncd/dashboard/pkg/controllers"
 	"github.com/tektoncd/dashboard/pkg/endpoints"
@@ -33,12 +36,36 @@ import (
 	"knative.dev/pkg/signals"
 )
 
+const csrfTokenLength = 32
+
 // Stores config env
 type config struct {
 	kubeConfigPath string
 	// Should conform with http.Server.Addr field
 	port             string
 	installNamespace string
+}
+
+func isCSRFSecureCookieEnabled() bool {
+	asBool, err := strconv.ParseBool(os.Getenv("CSRF_SECURE_COOKIE"))
+	if err == nil {
+		logging.Log.Infof("Is CSRF_SECURE_COOKIE enabled: %t", asBool)
+		return asBool
+	}
+	logging.Log.Warnf("Couldn't get CSRF_SECURE_COOKIE, defaulting to true: %s", err.Error())
+	return true
+}
+
+func getCSRFAuthKey() []byte {
+	logging.Log.Info("Generating CSRF auth key")
+	authKey := make([]byte, csrfTokenLength)
+	_, err := rand.Read(authKey)
+	if err == nil {
+		return authKey
+	}
+
+	logging.Log.Errorf("Couldn't generate random value for CSRF auth key: %s", err.Error())
+	return nil
 }
 
 func main() {
@@ -102,7 +129,14 @@ func main() {
 	controllers.StartKubeControllers(resource.K8sClient, resyncDur, dashboardConfig.installNamespace, routerHandler, ctx.Done())
 
 	logging.Log.Infof("Creating server and entering wait loop")
-	server := &http.Server{Addr: dashboardConfig.port, Handler: routerHandler}
+	CSRF := csrf.Protect(
+		getCSRFAuthKey(),
+		csrf.CookieName("token"),
+		csrf.Path("/"),
+		csrf.SameSite(csrf.SameSiteLaxMode),
+		csrf.Secure(isCSRFSecureCookieEnabled()),
+	)
+	server := &http.Server{Addr: dashboardConfig.port, Handler: CSRF(routerHandler)}
 
 	errCh := make(chan error, 1)
 	defer close(errCh)
--- a/overlays/dev/csrf-secure-cookie-patch.yaml
+++ b/overlays/dev/csrf-secure-cookie-patch.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: tekton-dashboard
+  namespace: tekton-pipelines
+spec:
+  template:
+    spec:
+      containers:
+        - name: tekton-dashboard
+          env:
+            - name: CSRF_SECURE_COOKIE
+              value: "false"
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Tekton Authors
+Copyright 2019-2020 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -79,6 +79,7 @@ func Register(resource endpoints.Resource) *Handler {
 	registerHealthProbe(resource, h.Container)
 	registerReadinessProbe(resource, h.Container)
 	registerKubeAPIProxy(resource, h.Container)
+	registerCSRFTokenEndpoint(resource, h.Container)
 	h.registerExtensions()
 	return h
 }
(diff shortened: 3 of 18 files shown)


cmd/dashboard/main.go
```diff
@@ -102,7 +103,8 @@ func main() {
 	controllers.StartKubeControllers(resource.K8sClient, resyncDur, dashboardConfig.installNamespace, routerHandler, ctx.Done())
 
 	logging.Log.Infof("Creating server and entering wait loop")
-	server := &http.Server{Addr: dashboardConfig.port, Handler: routerHandler}
+	CSRF := csrf.Protect([]byte("46adc6e0242a4fc1919bba0c744db625"), csrf.Secure(false), csrf.Path("/"), csrf.CookieName("token"))
```
AlanGreene, line 106: TODO:
- move auth key to config
- set 'Secure' false for dev only

AlanGreene:
TODO:
- move auth key to config
  - or generate random key on startup and live with the occasional need to refresh the browser if it's still use a stale cookie from a previous deployment?

AlanGreene:
@a-roberts this is ready for a proper look now.

The only thing remaining is deciding how we want to handle the `CSRF_SECURE_COOKIE` flag for dev. I was thinking about patching it in the dev overlays for kustomize so devs don't have to manually edit the deployment when working locally.

a-roberts:
> @a-roberts this is ready for a proper look now.
> 
> The only thing remaining is deciding how we want to handle the `CSRF_SECURE_COOKIE` flag for dev. I was thinking about patching it in the dev overlays for kustomize so devs don't have to manually edit the deployment when working locally.

Yep, this approach sounds good, checking the rest of the PR now

cmd/dashboard/main.go
```diff
@@ -102,7 +129,14 @@ func main() {
 	controllers.StartKubeControllers(resource.K8sClient, resyncDur, dashboardConfig.installNamespace, routerHandler, ctx.Done())
 
 	logging.Log.Infof("Creating server and entering wait loop")
-	server := &http.Server{Addr: dashboardConfig.port, Handler: routerHandler}
+	CSRF := csrf.Protect(
+		getCSRFAuthKey(),
```
a-roberts, line 133: This part's interesting, so while the return value doesn't get checked here, IDK why getCSRFAuthKey would ever error so this *should* never be nil 🤔

cmd/dashboard/main.go
```diff
@@ -102,7 +129,14 @@ func main() {
...
+		csrf.CookieName("token"),
+		csrf.Path("/"),
+		csrf.SameSite(csrf.SameSiteLaxMode),
+		csrf.Secure(isCSRFSecureCookieEnabled()),
+	)
+	server := &http.Server{Addr: dashboardConfig.port, Handler: CSRF(routerHandler)}
```
a-roberts, line 139: Ah, so CSRF's constructor accepts a routerHandler I see, this is the key I didn't figure out, good approach

pkg/router/router.go
```diff
@@ -302,6 +303,14 @@ func registerPropertiesEndpoint(r endpoints.Resource, container *restful.Contain
 	container.Add(wsDefaults)
 }
 
+func registerCSRFTokenEndpoint(r endpoints.Resource, container *restful.Container) {
+	ws := new(restful.WebService)
+	ws.Filter(restful.NoBrowserCacheFilter)
+	ws.Path("/v1/token").Produces("text/plain")
+	ws.Route(ws.GET("/").To(r.GetToken))
```
a-roberts, line 310: Question here, and this is more so I can understand it fully - on anything but OpenShift where we've got the secure proxy on top (let's say they're on GKE or AWS), what's to stop the malicious party from visiting this endpoint first to get a token? It'd be pointless though wouldn't it because it's a new token on top of every single request we do that can alter state?

src/api/comms.js
```diff
@@ -40,7 +43,8 @@ export function checkStatus(response = {}) {
       case 204:
         return {};
       default:
-        if (response.headers.get('content-type') === 'text/plain') {
+        const contentType = response.headers.get('content-type');
+        if (contentType && contentType.indexOf('text/plain') !== -1) {
```
a-roberts, line 47: Nit, any advantage to an indexOf and not a contains here? Easier to interpret with contains imo

test/e2e-tests.sh
```diff
@@ -62,10 +62,15 @@ export REPO_URL="https://github.com/ncskier/go-hello-world"
 export EXPECTED_RETURN_VALUE="Hello World!"
 export REGISTRY="gcr.io/${E2E_PROJECT_ID}/${E2E_BASE_NAME}-e2e-img"
 export TEKTON_PROXY_URL="http://localhost:9097/proxy/apis/tekton.dev/v1alpha1/namespaces/tekton-pipelines"
+export CSRF_HEADERS_STORE="csrf_headers.txt"
 
 # Kubectl static resources
 kubectl apply -f ${tekton_repo_dir}/test/resources/static
 
+curl -D $CSRF_HEADERS_STORE http://localhost:9097/v1/token
+export CSRF_TOKEN=`grep 'X-Csrf-Token' $CSRF_HEADERS_STORE | sed -e 's/^X-Csrf-Token: //;s/\r//'`
```
a-roberts, line 71: Is the casing here correct/required? It's X-CSRF-Token elsewhere

AlanGreene: yeah header case gets mangled all sorts of ways, to avoid confusion I think I'll change this to `grep -i 'X-CSRF-Token'` for a case-insensitive match

src/api/comms.js
AlanGreene: contains was removed from the spec so it would be `includes`, I'll change it 👍 Just too used to having to support IE which doesn't have `includes`

cmd/dashboard/main.go
AlanGreene: yeah, found a load of conflicting docs scattered around the place that had me confused and sent me down the wrong path with filters etc. Eventually threw all that out and reviewed the godocs for both projects directly to figure this out. The match was not obvious at first.

cmd/dashboard/main.go
AlanGreene: if for some reason the system's random number generator is unable to provide sufficient output it'll fail and we return nil, should be very unlikely to happen though and gorilla/csrf has some code to attempt to recover from a missing key anyway.

pkg/router/router.go
AlanGreene: There are 2 main pieces to this, a cookie and a custom header, both must be provided to pass the check.

CSRF relies on the default browser behaviour of forwarding cookies on requests to a given domain. So an attacker would craft a link / form on a page they control, that navigates / submits to the target domain. If a user visits this malicious page and follows the link (or the form is auto-submitted using a script on the page) the browser will happily forward any cookies it has for the target. If the user is already authenticated to that system, the attacker now has the ability to perform actions in the system as that user.

To mitigate against this:
- do not use GET for mutations, this prevents malicious links from causing harm
- set SameSite=Lax on session / auth cookies (this is the default in most modern browsers when SameSite is omitted), this means that in a cross-domain scenario they'll only be submitted on GET requests which should be safe (point above)
- since we may not be able to set the SameSite attribute on all affected cookies (may not control them ourselves) we can take additional steps:
  - require a custom header on requests containing some random token that can be matched against some server-side / session value
  - the token value cannot be retrieved by scripts on another origin (same origin protection on reading response body / headers)

gorilla/csrf uses the 'double submit' method where masked tokens are generated for each request that must be submitted in a custom header as well as a cookie that will be automatically provided by the browser on subsequent requests. For any 'unsafe' methods (POST, PUT, DELETE, etc.) these values are unmasked on the server and compared against the current key to verify them.

Summary:
- need to provide both cookie + header
- cannot read cookies cross-origin
- cannot read response body / headers cross-origin
- by default modern browsers will not submit cookies on 'unsafe' requests made cross-origin

a-roberts: Thanks Alan, excellent explanation 👍

a-roberts:
Looking good @AlanGreene, lemme know when you're ready to get this in (with no more WIP and draft status 😎 )

AlanGreene:
Just need to update the dev overlays to patch the CSRF_SECURE_COOKIE flag and we should be good to go.

AlanGreene:
@a-roberts should be good now

overlays/dev/csrf-secure-cookie-patch.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: tekton-dashboard
+  namespace: tekton-pipelines
```
a-roberts, line 6: Would this patch still be applied in the OpenShift dev incase given the namespace is now openshift-pipelines?

AlanGreene: oh that's a good question... should be able to leave out the namespace here right?

a-roberts:
/lgtm
/approve
/meow hats

Thanks for this one, nice approach

Issue #1374, opened 2020-05-13
0.6.1.2 - SA and Secrets page that should be hidden in read-only mode can be accessed through URL still


a-roberts:
# Expected behavior
If in read-only mode, we're omitting them from the sidenav atm but they can still be accessed through the URL

# Steps to reproduce the problem

1. Access /secrets or /serviceaccounts for example - you can still, as we're not conditionally creating the URL route/path I assume

# Environment
- Kubernetes Platform:
Using Docker Desktop here, likely any

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:
0.6.1.2 release candidate

# Additional Info
As with https://github.com/tektoncd/dashboard/issues/1373, note, there is no available release candidate to the public yet, so this is for those working on a security issue to pick up (as that's needed for the 0.6.1.2 release)

Whatever fix we have should be tested on master too

AlanGreene:
Ah so the routes are still registered. There's no content returned from the APIs but we should still clean this up since it could cause confusion if someone hits an old link and gets an incomplete page.

PR #1378, opened 2020-05-13
Disable Secrets and ServiceAccounts routes in read-only mode

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1374

The SideNav items for the Secrets and ServiceAccounts pages
were removed in a previous change, but the client routes were
still active. This means that a user could construct a URL (or
follow an old link) to one of these pages and the dashboard
would still attempt to render them.

Of course, since the Dashboard does not have permission to read
Secrets or ServiceAccounts in read-only mode (not included in RBAC)
the pages would have no content, and would display their empty states.

This is a minor cleanup to ensure we don't register the routes
when we know there'll be nothing to display.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -198,43 +198,50 @@ export /* istanbul ignore next */ class App extends Component {
                   />
 
                   {!this.props.isReadOnly && (
-                    <Route
-                      path={paths.importResources()}
-                      component={ImportResources}
-                    />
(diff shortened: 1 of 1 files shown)


Megan-Wright:
Tested out locally, cant access either of the routes so all looking good :)
/lgtm

Issue #1367, opened 2020-05-11
Pipelineruns in Tekton dashboard show incorrect status for each step


tam512:
# Expected behavior
When running a pipelinerun in Tekton dashboard,  the status for each step run in the pipelinerun should report correctly 

# Actual behavior
Tekton dashboard does not report status for the steps in a pipelinerun correctly

For example, in Kabanero 0.9.0-rc.2, run pipeline nodejs-express-build-deploy-pl in Tekton dashboard. It failed at step `image-digest-exporter-pndfj` and other steps showed as `Not run` as following

![image](https://user-images.githubusercontent.com/22480038/81605926-cf91b900-9397-11ea-9bde-c1df5792cccc.png)

But I looked into each step, I see that it successfully ran steps `enforce-stack-policy-pre-build`, `build`, `enforce-stack-policy-post-build`, and failed on` push` step.
So the steps that ran successfully should show as `Completed` and the push step should show as `Failed` and the rest of the steps should be `Not run`

The followings are the screenshots for the steps that ran successfully, but they showed as `Not run`

![image](https://user-images.githubusercontent.com/22480038/81606478-cc4afd00-9398-11ea-94f3-2e217c42dadf.png)

![image](https://user-images.githubusercontent.com/22480038/81606506-d4a33800-9398-11ea-820f-cb55cf1ca005.png)

![image](https://user-images.githubusercontent.com/22480038/81606524-dd940980-9398-11ea-881f-f6c5e4955aaf.png)

- The `push` step is actually failed with error,  but it showed as `Not run` instead of Failed

![image](https://user-images.githubusercontent.com/22480038/81606806-4da28f80-9399-11ea-8ddd-7ff97cff53ef.png)

# Steps to reproduce the problem

1. Create a webhook for nodejs express app with pipeline nodejs-express-build-deploy-pl from the Tekton dashboard.  Use OpenShift internal registry (e.g. _image-registry.openshift-image-registry.svc:5000/svt-project2_)
2. Make a small change in the app git repo to trigger the pipelinerun.

# Environment
- Kubernetes Platform: OPC 4.4.3

- Tekton Pipelines version:   tekton.version: v0.11.3

- Tekton Triggers version:

- Tekton Dashboard version:

Image ID:       gcr.io/tekton-releases/github.com/tektoncd/dashboard/cmd/dashboard@sha256:3b4dc45de5ae0e741099018afe3e1108cb93f70aa24138456be03d9fde677269

# Additional Info

[...]

PR #1385, opened 2020-05-15
[WIP] Use step logs to check if step was skipped

JustinKuli:
# Changes

Currently, we determine if steps were skipped by assuming that first failing step in the `steps` array is the actual failure, and then steps after that in the array were skipped. It is true that all steps "after" the failing step are skipped, but the steps array is not necessarily in the correct order, so our current method fails in some cases. See #1367 for examples.

This change will adjust the function to use logs from the steps to determine if they were skipped, because they do reliably show the correct message (which is how users know the status in the UI is incorrect).

Overall, I expect this might be a temporary fix until we can fully rely on pipeline correctly ordering the steps (possibly in https://github.com/tektoncd/pipeline/pull/2455), and then the current implementation would be fine.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -198,43 +198,50 @@ export /* istanbul ignore next */ class App extends Component {
                   />
 
                   {!this.props.isReadOnly && (
-                    <Route
-                      path={paths.importResources()}
-                      component={ImportResources}
-                    />
-                  )}
+                    <>
+                      <Route
+                        path={paths.importResources()}
+                        component={ImportResources}
+                      />
 
-                  <Route path={paths.secrets.all()} exact component={Secrets} />
-                  <Route
-                    path={paths.secrets.byName()}
-                    exact
-                    component={Secret}
-                  />
-                  <Route
-                    path={paths.secrets.byNamespace()}
-                    exact
-                    component={Secrets}
-                  />
-                  <Route
-                    path={paths.secrets.create()}
-                    exact
-                    component={CreateSecret}
-                  />
-                  <Route
-                    path={paths.serviceAccounts.byName()}
-                    exact
-                    component={ServiceAccount}
-                  />
-                  <Route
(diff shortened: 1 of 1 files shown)


JustinKuli:
Putting this out there in case anyone wants to see how the logs are getting into the `updateUnexecutedSteps` function, since I expect that might need to be changed.

It is still a work in progress because there are other instances where the `updateUnexecutedSteps` function is called where I have not updated it to use the logs yet.

One potential downside to this implementation is that more items on the page have to wait for the logs to be retrieved before they fully render, as indicated by the additional async code in the tests...

AlanGreene:
The log content isn't part of the API so I wouldn't rely on it for functionality.

JustinKuli:
Yeah... right on both points @AlanGreene.

Looking more at https://github.com/tektoncd/pipeline/pull/2455, that's expected to update the TaskRun's `status.conditions.*.message`to correctly display the failing step. It won't change the order of the steps in the TaskRun's status.

The order of the steps in the status is covered by pipeline's deprecation policy, so we will be stuck with all the "unspecified" / "init" steps being added at the end (the current behavior based on https://github.com/tektoncd/pipeline/blob/1fbac2aa003733a4a01bcb8de23e35478ccc1d4a/pkg/pod/status_test.go#L831).

JustinKuli:
New idea: for the `updateUnexecutedSteps`function, sort the steps based on finishedAt and startedAt. It will still be incorrect for now, but it might be better. And when pipeline increases the precision on startedAt, it should improve again. It might never be perfect: regardless of precision, I think there could still be ties.

JustinKuli:
Something with the way I pushed a branch to my fork messed this up... I think because I pushed to get even with master before pushing my new version.

I can make a new PR, or we can try to fix this one? I was going to have to re-title it anyway. Sorting the steps by the timestamps seems to work.

PR #1391, opened 2020-05-18
Ensure strings in ImportResources are extracted for translation

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/453

Ensure all strings in the ImportResources container are externalized
so they can be translated.

Also add simple pseudolocalization to help identify hardcoded
strings in the UI. This wrap all externalized strings in
`%%_<string>_%%` to help identify improper concatenation, or
any missed strings.

In future we can enhance this to provide a more 'traditional'
pseudolocalization experience, replacing characters in the string,
e.g. `Account Settings` becomes `%%_Àççôûñţ Šéţţîñĝš_%%`, to
validate encoding and other handling is also correct.

# Submitter Checklist

review them:

for more details._


src/containers/App/App.js
```diff
@@ -75,10 +75,23 @@ import {
...
-import messages from '../../nls/messages_en.json';
+import messageBundle from '../../nls/messages_en.json';
 
 import '../../components/App/App.scss';
 
+const messages = messageBundle;
+
+/* istanbul ignore next */
+if (process.env.I18N_PSEUDO) {
+  // Make it easier to identify untranslated strings in the UI
```
a-roberts, line 86: Yeah, wondering if it's worth documenting this too (can be a follow-up PR) in DEVELOPMENT.md to help us spot such occurrences easier - atm we're surely just populating messagesToDisplay but they never actually appear anywhere? We could, say, conditionally do a console log when running in dev mode to see what this variable contains

AlanGreene: Yup definitely planning to document this in DEVELOPMENT.md once it's ready, not quite there yet. The intent is to wrap the strings displayed in the UI in some clear boundary markers so we can easily identify:
- which strings are not yet externalised (i.e. the ones missing the markers)
- improper concatenation of strings (we'll see multiple sets of boundary markers in a single string in the UI)
- encoding issues (shouldn't happen...)
- areas where we're not handling overflow / variable width strings properly
- etc.

Pseudotranslation / pseudolocalization is a common practice to help with these and avoid issues for users. It can be as simple as what I've included in this initial PR, although that doesn't quite achieve all of the goals just yet.

What I'm aiming for is something like this:

PR #1394, opened 2020-05-19
Sort steps by timestamps

### LINKED ISSUE #1367: Pipelineruns in Tekton dashboard show incorrect status for each step

JustinKuli:
# Changes

Sort steps by timestamps in a TaskRun

Currently, all injected steps appear before the regularly defined steps in some views. Now they are sorted by the finishedAt timestamp in each step's status, with ties broken by the startedAt timestamp.

The current order is not always correct, since some injected steps actually run after regular steps. This ordering also causes some steps to be incorrectly labelled as either skipped or failed.

The new order should be better, but it will still depend on the precision of the timestamps: seconds might not be sufficient in all cases. There are plans in 'pipeline' to increase the precision.

Fixes #1367

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/utils/src/utils/index.js
+++ b/packages/utils/src/utils/index.js
@@ -182,13 +182,34 @@ export function formatLabels(labelsRaw) {
   return formattedLabelsToRender;
 }
 
+// Sorts the steps by finishedAt and startedAt timestamps
+export function sortStepsByTimestamp(steps) {
+  return steps.sort((i, j) => {
+    const iFinishAt = new Date(i.stepStatus?.terminated?.finishedAt).getTime();
+    const jFinishAt = new Date(j.stepStatus?.terminated?.finishedAt).getTime();
+    const iStartedAt = new Date(i.stepStatus?.terminated?.startedAt).getTime();
+    const jStartedAt = new Date(j.stepStatus?.terminated?.startedAt).getTime();
+
+    if (!iFinishAt || !jFinishAt) {
+      return 0;
+    }
+    if (iFinishAt !== jFinishAt) {
+      return iFinishAt - jFinishAt;
+    }
+    if (!iStartedAt || !jStartedAt) {
+      return 0;
+    }
+    return iStartedAt - jStartedAt;
+  });
+}
+
 // Update the status of steps that follow a step with an error
 export function updateUnexecutedSteps(steps) {
   if (!steps) {
     return steps;
   }
   let errorIndex = steps.length - 1;
-  return steps.map((step, index) => {
+  return sortStepsByTimestamp(steps).map((step, index) => {
     // Update errorIndex
     if (step.reason !== 'Completed') {
       errorIndex = Math.min(index, errorIndex);
--- a/packages/components/src/components/TaskTree/TaskTree.js
+++ b/packages/components/src/components/TaskTree/TaskTree.js
@@ -12,6 +12,7 @@ limitations under the License.
 */
 
 import React, { Component } from 'react';
+import { sortStepsByTimestamp } from '@tektoncd/dashboard-utils';
 import Task from '../Task';
 
 import './TaskTree.scss';
@@ -47,7 +48,7 @@ class TaskTree extends Component {
               expanded={expanded}
               onSelect={this.handleSelect}
               reason={reason}
-              steps={steps}
+              steps={sortStepsByTimestamp(steps)}
               succeeded={succeeded}
               pipelineTaskName={pipelineTaskName}
             />
--- a/packages/utils/src/utils/index.test.js
+++ b/packages/utils/src/utils/index.test.js
@@ -26,6 +26,7 @@ import {
   reorderSteps,
   selectedTask,
   selectedTaskRun,
+  sortStepsByTimestamp,
   stepsStatus,
   taskRunStep,
   updateUnexecutedSteps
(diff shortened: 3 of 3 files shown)


JustinKuli:
Original problem: the "image-digest-exporter" step ( which is injected by a git PipelineResource) is incorrectly identified as the failed step.

Actually, the "push" step is the failing one, as the "image-digest-exporter" was skipped. We can see the correct output after we sort by the timestamps.

I just want to note that sorting them by the timestamps isn't a perfect solution. The "image-digest-exporter" step should happen after the "push" step, like we see above. But when multiple steps are skipped, the start and finish times might be equal, and it leads to slightly incorrect results.

## Current state (example where push fails):
1367-current-1

## After this change (example where push fails):
1367-adjusted-1

## Imperfect ordering even after this change:
1367-adjusted-2

AlanGreene:
@JustinKuli can you share some of the Pipelines / Tasks you used to test this? I don't think I'm seeing any difference testing this out locally. The `image-digest-exporter` step is still showing before the failed step, and the failed step is showing as not run as a result. This could be down to the lack of precision in the times as you mentioned.

JustinKuli:
@AlanGreene I was just using the pipeline / tasks from https://developer.ibm.com/tutorials/tekton-triggers-101/, since that was something I already had in my cluster. The Tasks and Pipeline are here: https://github.com/ncskier/catapp/blob/master/tekton/build-and-deploy-openshift-pipeline.yaml

In your Task, do you have another "normal" step before the failing one (which would complete well before the `image-digest-exporter` step)? Because in the current implementation, the image-digest-exporter step will always come first... regardless of if there was another step that finished even earlier (the "build" step in my example runs).

Let me put together a more repeatable Task that can demonstrate this. I might have just gotten lucky in my run.

JustinKuli:
@AlanGreene here's a simple example which seems to give repeatable results for me (I ran it five times, all with the same result):
```
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  generateName: fail-bbb-skip-image-
spec:
  resources:
    outputs:
    - name: image
      resourceSpec:
        type: image
        params:
        - name: url
          value: fake-registry.io/fake-image
  taskSpec:
    resources:
      outputs:
      - name: image
        type: image
    steps:
    - name: aaa
      image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3
        echo "Step aaa will complete successfully..."
        sleep 3
    - name: bbb
      image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3
        echo "Step bbb will fail..."
        sleep 3
        exit 1
```

## Viewed using the current dashboard:
tekton-1367-new-1

## Viewed using my build:
tekton-1367-new-2

packages/utils/src/utils/index.js
```diff
@@ -182,13 +182,38 @@ export function formatLabels(labelsRaw) {
...
+      return 0;
+    }
+    if (iFinishedAt !== jFinishedAt) {
+      return iFinishedAt - jFinishedAt;
+    }
+    // if finishedAt timestamps are equal, compare the startedAt timestamps
+    if (!iStartedAt || !jStartedAt) {
+      return 0;
+    }
+    if (iStartedAt !== jStartedAt) {
```
AlanGreene, line 203: should be able to `return iStartedAt - jStartedAt;` here and avoid the additional `if` and `return 0`

JustinKuli:
@AlanGreene Review comments implemented, commits squashed, and rebased to master.

PR #1400, opened 2020-05-19
Add more variations in e2e tests

eddycharly:
# Changes

This PR adds more variations in e2e tests.
It now tests `dev` and `dev-locked-down` overlays.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -18,6 +18,8 @@
 
 source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/e2e-tests.sh
 
+pipeline_release=https://github.com/tektoncd/pipeline/releases/download/v0.11.0/release.yaml
+
 function print_diagnostic_info() {
   echo "Diagnostics:"
   resources=("pv" "pvc" "pods")
@@ -35,7 +37,7 @@ function install_kustomize() {
 
 function install_pipeline_crd() {
   echo ">> Deploying Tekton Pipelines"
-  kubectl apply --filename https://github.com/tektoncd/pipeline/releases/download/v0.11.0/release.yaml || fail_test "Tekton pipeline installation failed"
+  kubectl apply --filename $pipeline_release  || fail_test "Tekton pipeline installation failed"
 
   # Make sure thateveything is cleaned up in the current namespace.
   for res in pipelineresources tasks pipelines taskruns pipelineruns; do
@@ -44,6 +46,11 @@ function install_pipeline_crd() {
 
   # Wait for pods to be running in the namespaces we are deploying to
   wait_until_pods_running tekton-pipelines || fail_test "Tekton Pipeline did not come up"
+}
+
+function delete_pipeline_crd() {
+  echo ">> Deleting Tekton Pipelines"
+  kubectl delete --filename $pipeline_release || fail_test "Tekton pipeline deletion failed"
 }
 
 # Called by `fail_test` (provided by `e2e-tests.sh`) to dump info on test failure
@@ -69,8 +76,9 @@ function dump_extra_cluster_state() {
 }
 
 function install_dashboard_backend() {
-  echo ">> Deploying the Dashboard backend"
-  kustomize build overlays/dev | ko apply -f - || fail_test "Dashboard backend installation failed"	
+  overlay=$1
+  echo ">> Deploying the Dashboard backend ($overlay)"
+  kustomize build --load_restrictor none overlays/$overlay | ko apply -f - || fail_test "Dashboard backend installation failed"	
   # Wait until deployment is running before checking pods, stops timing error
   for i in {1..30}
   do
@@ -86,6 +94,12 @@ function install_dashboard_backend() {
   done
   # Wait for pods to be running in the namespaces we are deploying to
   wait_until_pods_running tekton-pipelines || fail_test "Dashboard backend did not come up"
+}
+
+function uninstall_dashboard_backend() {
+  overlay=$1
+  echo ">> Removing the Dashboard backend ($overlay)"
+  kustomize build --load_restrictor none overlays/$overlay | ko delete -f - || fail_test "Dashboard backend removal failed"	
 }
 
 function dump_cluster_state() {
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -23,136 +23,147 @@ source $(dirname $0)/e2e-common.sh
 # Script entry point.
 
 initialize $@
-
-header "Setting up environment"
-
-install_pipeline_crd
 install_kustomize
-install_dashboard_backend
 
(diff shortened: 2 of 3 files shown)


eddycharly:
Testing read-only mode fails. I suspect the controller watching secrets is started though RBAC doesn’t allow secrets access.

test/e2e-tests.sh
```diff
@@ -23,136 +23,156 @@ source $(dirname $0)/e2e-common.sh
...
+  # Kubectl envsubst resources
+  pipelineResourceFiles=($(find ${tekton_repo_dir}/test/resources/envsubst -iname "pipelineresource*.y?ml"))
+  for file in ${pipelineResourceFiles[@]};do
+    cat "${file}" | envsubst | kubectl apply -f - || fail_test "Failed to create pipelineresource: ${file}"
+  done
+
+  pipelineRunFiles=($(find ${tekton_repo_dir}/test/resources/envsubst -iname "pipelinerun*.y?ml"))
+  for file in ${pipelineRunFiles[@]};do
+    cat "${file}" | envsubst | kubectl apply -f - || fail_test "Failed to create pipelinerun: ${file}"
+  done
```
eddycharly, line 98: One concern i have is that it's not using the dashboard proxy to create the resources.

a-roberts: Hmm, this is a big drawback though surely - we're just doing a normal kube apply instead of hitting our PIpelineRuns/PipelineResources APIs which we were previously doing - have I missed something?

eddycharly: When testing the read only overlay, calling the proxy won't work (the dashboard SA cannot create resources).
Kubectl will work because it's using the test runner kube context.

Should i use the proxy in read write mode and kubectl in read only mode ?

a-roberts: Yep, sounds good to me

eddycharly: done

eddycharly:
Can we get this in ? Would like to test single namespace overlay too

test/e2e-common.sh
```diff
@@ -46,6 +46,11 @@ function install_pipeline_crd() {
   wait_until_pods_running tekton-pipelines || fail_test "Tekton Pipeline did not come up"
 }
 
+function delete_pipeline_crd() {
+  echo ">> Deleting Tekton Pipelines"
+  kubectl delete --filename https://github.com/tektoncd/pipeline/releases/download/v0.11.0/release.yaml || fail_test "Tekton pipeline deletion failed"
```
AlanGreene, line 51: Can we extract the release.yaml location to a variable and use it in `install_pipeline_crd` above too? Hopefully this will help avoid issues in future where we upgrade and forget to update one of these.

test/e2e-tests.sh
```diff
@@ -23,136 +23,163 @@ source $(dirname $0)/e2e-common.sh
...
-done
 
-if [ "$podCurled" = "false" ]; then
-  fail_test "Test Failure, Not able to curl the pod"
-fi
+  kill -9 $dashboardForwardPID
+  kill -9 $podForwardPID
+
+  uninstall_dashboard_backend $overlay
+  delete_pipeline_crd
```
AlanGreene, line 179: Do we need to install / uninstall pipelines for each dashboard overlay being tested or can we just do this once up front?

test/e2e-tests.sh
```diff
@@ -23,136 +23,163 @@ source $(dirname $0)/e2e-common.sh
...
+    echo "Creating resources using kubectl..."
+    pipelineResourceFiles=($(find ${tekton_repo_dir}/test/resources/envsubst -iname "pipelineresource*.y?ml"))
+    for file in ${pipelineResourceFiles[@]};do
+      cat "${file}" | envsubst | kubectl apply -f - || fail_test "Failed to create pipelineresource: ${file}"
+    done
+
+    pipelineRunFiles=($(find ${tekton_repo_dir}/test/resources/envsubst -iname "pipelinerun*.y?ml"))
+    for file in ${pipelineRunFiles[@]};do
+      cat "${file}" | envsubst | kubectl apply -f - || fail_test "Failed to create pipelinerun: ${file}"
+    done
```
AlanGreene, line 90: It doesn't look like this is actually testing anything dashboard related, it's just using kubectl and checking the exit code. Is there something missing here?

eddycharly: You're right, `kubectl` is used when testing read-only mode as creating resources with the proxy is not allowed.
So it goes like this:
- read-write mode -> create and test using proxy
- read-only mode -> create using kubectl, test using proxy

test/e2e-tests.sh
eddycharly: As we put everything in the same namespace it helps making sure is cleaned up before moving to the next tests.

It should not be long anyway.

test/e2e-tests.sh
a-roberts: Interesting one, see https://github.com/tektoncd/dashboard/pull/1400#pullrequestreview-415257523 (I mentioned this and it's been resolved so easy to miss)

test/e2e-common.sh
eddycharly: Good catch

eddycharly: done

test/e2e-tests.sh
```diff
@@ -23,136 +23,163 @@ source $(dirname $0)/e2e-common.sh
...
-    break
+
+  # API/resource configuration
+  export APP_SERVICE_ACCOUNT="e2e-tests"
+  export PIPELINE_NAME="simple-pipeline-insecure"
+  export IMAGE_RESOURCE_NAME="docker-image"
+  export GIT_RESOURCE_NAME="git-source"
+  export GIT_COMMIT="master"
+  export REPO_NAME="go-hello-world"
+  export REPO_URL="https://github.com/ncskier/go-hello-world"
```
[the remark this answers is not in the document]
eddycharly: changed it

test/e2e-tests.sh
```diff
@@ -23,136 +23,163 @@ source $(dirname $0)/e2e-common.sh
...
+    wait=$(kubectl wait --namespace tekton-pipelines --for=condition=available deployments/go-hello-world --timeout=30s)
+    if [ "$wait" = "deployment.apps/go-hello-world condition met" ]; then
+      deploymentExist=true
+      break
+    elif [ "$wait" = "deployment.extensions/go-hello-world condition met" ]; then
+      deploymentExist=true
+      break
+    else
+      echo "About to sleep, getting and describing pod info"
+      kubectl get pod --namespace tekton-pipelines -l app=go-hello-world -o name --namespace tekton-pipelines -o yaml
```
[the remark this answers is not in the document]
eddycharly: done

PR #1371, opened 2020-05-12
[WIP] Limit visibility to single namespace

eddycharly:
# Changes

This PR implements single namespace visibility in the tekton dashboard (see https://github.com/tektoncd/dashboard/issues/1018)

/cc @a-roberts @AlanGreene 

Changes are detailed below:

## Backend

RBAC is split into multiple `Role`/`ClusterRole` according to the concern they address.

| Concern | Tenancy mode | Resource type | Binding type | Scope |
| --- | --- | --- | --- | --- |
| Lookup pipelines infos | NA | `ClusterRole` | `RoleBinding` | namespace `tekton-pipelines` |
| Lookup triggers infos | NA | `ClusterRole` | `RoleBinding` | namespace `tekton-pipelines` |
| Backend | NA | `ClusterRole` | `ClusterRoleBinding` | cluster |
| Aggregate | NA | `ClusterRole` | `ClusterRoleBinding` | cluster |
| Tenant | single namespace | `ClusterRole` | `RoleBinding` | namespace `tekton-tenant` |
| Tenant | whole cluster | `ClusterRole` | `ClusterRoleBinding` | cluster |

A few more details:
- keep in mind that `ClusterRole` doesn't mean permissions at the cluster level, it depends on the type of binding (`RoleBinding` means namespace level permissions, `ClusterRoleBinding` means cluster level permissions)

- backend `ClusterRole` is the necessary permissions for the backend to work, it doesn't cover permissions related to tekton pipelines or triggers resources, only those related to the backend internals (dashboard extensions essentially). It also covers cluster wide tekton resources (`clustertasks` and `clustertriggerbindings`) as they are cluster wide by nature.

- aggregate `ClusterRole` is cluster wide. This is used by extensions and as extensions are not very mature yet this could change as things go

The properties endpoint now has a `TenantNamespace` property set to `tekton-tenant` when single namespace tenancy is set.

Additionally, controllers are run against the tenant namespace and only see changes occuring in this namespace so that changes in another namespace are not broadcasted.

## Frontend

On the frontend side, when retrieving the properties from the backend, if `TenantNamespace` is set
- we use this namespace to populate namespace selection drop downs.
- the `All namespaces` item in the drop down is not shown
- we call the `selectNamespace` action to make the `tekton-tenant` namespace the active selection
- all api calls performed to the backend are consequently related to the `tekton-tenant` namespace
- api calls to the backend requesting cluster level permissions will most likely fail and should not be performed

## Kustomize

[...]

PR #1406, opened 2020-05-20
Separate RBAC used to lookup triggers and pipelines versions

eddycharly:
# Changes

This PR separates `ClusterRole`s used to look up pipelines and triggers version from the main `ClusterRole`.

These two new `ClusterRole`s are added to the dashboard service account using `RoleBinding`s and therefore target a single namespace (not the whole cluster).

/cc @AlanGreene @a-roberts 

Extracted from the giant PR https://github.com/tektoncd/dashboard/pull/1371

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/dashboard/blob/master/CONTRIBUTING.md)
for more details._

PR #1411, opened 2020-05-20
Add backend support for single namespace visibility

eddycharly:
# Changes

This PR adds backend support for single namespace visibility.

It adds a new flag `--namespace` to specify the namespace that should the backend should be limited to.

It also splits the `ClusterRole` into a backend one and a tenant one, the tenant `ClusterRole` can be used with a `ClusterRoleBinding` for full cluster visibility or with a `RoleBinding` for single namespace visibility.

/cc @AlanGreene @a-roberts 

Extracted from the giant PR https://github.com/tektoncd/dashboard/pull/1371

# Submitter Checklist

review them:

for more details._


eddycharly:
/hold

Based on https://github.com/tektoncd/dashboard/pull/1406, should be merged after.

eddycharly:
@AlanGreene @a-roberts This PR raises two questions:
- what scope do we want for extensions ? cluster ? tenant namespace ? controller namespace ? in case of multiple dashboard instances, should each instance have its own extensions ?
- how do we want to treat cluster level resources ? clustertasks and clustertriggerbindings ?

eddycharly:
Regarding extensions, i changed the rules so that they apply to the tenant scope (cluster or single namespace).

The extensions themselves can still be created in any namespace though.
We should think about this a little bit more, but maybe in another PR.

a-roberts:
> @AlanGreene @a-roberts This PR raises two questions:
> 
> * what scope do we want for extensions ? cluster ? tenant namespace ? controller namespace ? in case of multiple dashboard instances, should each instance have its own extensions ?
> * how do we want to treat cluster level resources ? clustertasks and clustertriggerbindings ?

I think having each instance being able to have its own extensions makes sense. If you install multiple dashboards in an enterprise environments, you may only want the webhooks extension for one dashboard instance and not another. We currently rely on being in the same namespace anyway (as I've found when the webhooks extension is in tekton-pipelines but the dashboard is in openshift-pipelines).

RE:

```
how do we want to treat cluster level resources ? clustertasks and clustertriggerbindings ?
```

hmm. I believe if a user's made them at the cluster scope, they should be possible to everyone on the cluster (barring any additional custom RBAC a user may have set up), otherwise they should just be using normal Tasks etc. So I'd be cool with displaying those anyway personally.

eddycharly:
> I think having each instance being able to have its own extensions makes sense. If you install multiple dashboards in an enterprise environments, you may only want the webhooks extension for one dashboard instance and not another. We currently rely on being in the same namespace anyway (as I've found when the webhooks extension is in tekton-pipelines but the dashboard is in openshift-pipelines).

I agree, the service based extensions should be dashboard instance specific.
There's also the extension crd thing, right now they are queried using the proxy at the cluster level, then resources are queried at the cluster or namespace level depending on the namespace dropdown selection.
If we want them to be dashboard instance specific we will have to change the current implementation.

> hmm. I believe if a user's made them at the cluster scope, they should be possible to everyone on the cluster (barring any additional custom RBAC a user may have set up), otherwise they should just be using normal Tasks etc. So I'd be cool with displaying those anyway personally.

Sounds good and should not require changes 👍

AlanGreene:
I have no objections to changing the resource extensions to be consistent with behaviour of the service based extensions. Consistency is easier to explain, and in this case I think it makes sense for the same reasons Adam mentioned for the service based ones.

base/kustomization.yaml
```diff
@@ -17,10 +17,20 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
   - 200-clusterrole-backend.yaml
+  - 200-clusterrole-extensions.yaml
   - 200-clusterrole-pipelines.yaml
+  - 200-clusterrole-tenant.yaml
   - 200-clusterrole-triggers.yaml
-  - 201-clusterrolebinding.yaml
+  - 201-clusterrolebinding-backend.yaml
+  - 201-clusterrolebinding-extensions.yaml
+  - 201-clusterrolebinding-tenant.yaml
+  # this one should be used instead of 201-clusterrolebinding-extensions.yaml
```
a-roberts, line 27: So what's the purpose in these comments still being here, are users expected to comment this part out accordingly (going against the way we use kustomize in general with base/overlays here, or maybe left behind as a mistake and can now be removed?)

cmd/dashboard/main.go
```diff
@@ -47,6 +47,7 @@ var (
 	readOnly           = flag.Bool("read-only", false, "Enable or disable read only mode")
 	webDir             = flag.String("web-dir", "", "Dashboard web resources dir")
 	csrfSecureCookie   = flag.Bool("csrf-secure-cookie", true, "Enable or disable Secure attribute on the CSRF cookie")
+	tenantNamespace    = flag.String("namespace", "", "Tenant namespace. If set, limits the scope of resources watched to this namespace")
```
a-roberts, line 50: It's not just `watched` right, it's anything? E.g. I can create/rerun resources etc in the tenant ns only?

pkg/controllers/controller.go
```diff
@@ -29,35 +29,41 @@ import (
...
+func StartKubeControllers(clientset k8sclientset.Interface, resyncDur time.Duration, installNamespace, tenantNamespace string, readOnly bool, handler *router.Handler, stopCh <-chan struct{}) {
 	logging.Log.Info("Creating Kube controllers")
-	kubeInformerFactory := k8sinformers.NewSharedInformerFactory(clientset, resyncDur)
+	clusterInformerFactory := k8sinformers.NewSharedInformerFactory(clientset, resyncDur)
+	tenantInformerFactory := k8sinformers.NewSharedInformerFactoryWithOptions(clientset, resyncDur, k8sinformers.WithNamespace(tenantNamespace))
 	// Add all kube controllers
-	kubecontroller.NewExtensionController(kubeInformerFactory, installNamespace, handler)
-	kubecontroller.NewNamespaceController(kubeInformerFactory)
+	kubecontroller.NewExtensionController(clusterInformerFactory, installNamespace, handler)
+	if tenantNamespace == "" {
```
a-roberts, line 58: Right because we don't need to watch for new namespaces in the event we know the tenant namespace and want to only create/edit/view resources in this namespace, so this is a good optimisation

eddycharly:
> hmm. I believe if a user's made them at the cluster scope, they should be possible to everyone on the cluster (barring any additional custom RBAC a user may have set up), otherwise they should just be using normal Tasks etc. So I'd be cool with displaying those anyway personally.

Rethinking about this, one potential issue with this is that team 1 could modify a ClusterTask used by team 2. This opens doors for bad user experience.
On one hand it’s a great feature for sharing tasks between teams and on the other hand it could be a source of conflicts.

How about a feature flag to make cluster level resources read only even in a read write install ?

base/kustomization.yaml
eddycharly: It’s here as a comment to help review. I didn’t write overlays for single namespace yet.
I am not completely sure how to design base/overlays as the resources you pointed out are mutually exclusive.
Two bases ?
I think I will make a dedicated PR for that, I am not sure it’s a good idea to do it here.

cmd/dashboard/main.go
eddycharly: Good catch, will update the description.

pkg/controllers/controller.go
eddycharly: Yes, the backend should not be allowed to list/watch namespaces so better not to try.
This an issue right now with secrets in read only, the backend is not allowed to list/watch secrets but the controller is started, logs are full of access denied messages.  I got a PR merged for that yesterday.

AlanGreene:
I think cluster-scoped resources should be read-only in a namespace-scoped dashboard install.

If I choose to restrict the dashboard to a specific namespace I think I'd be unlikely to want users modifying content outside that namespace.

Imagine the following use case:
- multiple project teams each with their own namespace + dashboard
- set of common tasks for compliance checks for example, administered by a central team
- project teams consume the common tasks but cannot modify them

If there are use cases for allowing creation / modification of cluster-scoped resources by a namespace-scoped dashboard I'd be interested in discussing them, but would prefer not to add additional config options unless necessary.

cmd/dashboard/main.go
eddycharly: fixed

eddycharly:
Thanks for the comments/suggestions around extensions. I will take care of this in a next PR based on the discussion here.

The PR is quite large already and should be ready for review.

Everything has been implemented but is not exposed yet, the dashboard should behave exactly as before. I will start addressing the frontend side in the same way.

Once frontend and backend are ready, i will work on putting everything together to get the single namespace manifests generated.

cmd/dashboard/main.go
eddycharly: @a-roberts in fact no, it's really for watching (and driving the frontend).

> I can create/rerun resources etc in the tenant ns only?

This is not taken care of by this flag but by RBAC rules.

It should be noted somewhere (at least in the docs). Not sure about the correct description to set here.

eddycharly:
I took the following approach for resource extensions in the frontend PR:
- cluster scope deployment -> all extensions in the cluster are considered
- namespace scope deployment -> only extension in the tenant namespace are considered

AlanGreene:
So far this looks good, I haven't had a chance to test it out in an OpenShift environment yet but that's next on my list.

We'll need to document the kustomize overlays once we're done with all of this refactoring, as I don't think it'll be clear to newcomers how all of the different pieces are used.

We should probably also spend some time revising the content / structure of the README to make it more approachable.

eddycharly:
>So far this looks good, I haven't had a chance to test it out in an OpenShift environment yet but that's next on my list.

> We'll need to document the kustomize overlays once we're done with all of this refactoring, as I don't think it'll be clear to newcomers how all of the different pieces are used.

I added a `dev-single-namespace` overlay in the frontend support PR for testing if you want (look at the PR description here #1425 ).

I also started thinking about how to reorganize kustomize code, the approach i have in mind is built around three aspects:
- plaform: we have currently 3 platfoms: k8s, openshift, and openshift-image-stream. the platform is not enough to produce a complete base, we need to consider rbac mode also
- rbac: single namespace or full cluster (platform + rbac should make a complete base)
- features: dev and/or readonly (only mutate existing manifests, don't bring new resources, patches should be reusable across base variations but they should also be consistent with base variations).

In the end it would make 24 possibilities, that's a lot but it's manageable if organized correctly.

I feel like we will hit the kustomize limits at some points, maybe a solution based on templating would allow easier maintenance and more productive work.

From the end user point of view, i would really love to be able to provide the configuration only, this is not the case with kustomize, if something changes in the base release, i also have to modify my kustomization.
Having a config with a boolean saying `read-only: true` should be enough IMO, i shouldn't need to know how this boolean config is applied in the various manifests of the release.

> We should probably also spend some time revising the content / structure of the README to make it more approachable.

Completely agree.

README.md
```diff
@@ -120,9 +120,13 @@ These options are documented below:
 | `--read-only` | Enable or disable read only mode | `bool` | `false` |
 | `--web-dir` | Dashboard web resources directory | `string` | `""` |
 | `--logout-url` | If set, enables logout on the frontend and binds the logout button to this url | `string` | `""` |
+| `--namespace` | If set, limits the scope of resources watched to this namespace only | `string` | `""` |
```
AlanGreene, line 123: Let's hold off on adding the readme update until the frontend piece is in place. We know some of the experience will be broken util that's complete.

eddycharly: References to `--namespace` have been removed.

eddycharly:
@a-roberts lgty ? 😅

a-roberts:
/lgtm

A big step in the right direction, thanks for this one

Issue #1409, opened 2020-05-20
No option to change what pipeline /taskrun to create when filter is applied


CarolynMabbott:
# Expected behavior
I can change the taskrun / pipelinerun without having to close the modal and remove the filter

# Actual behavior
I have to close the modal, remove the filter than go back to the create modal

# Steps to reproduce the problem

1. have a filter on pipeline / taskrun
2. click on create 
3. not be able to change pipelinerun / taskrun 

# Environment
- Kubernetes Platform:

docker

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

master 

# Additional Info

AlanGreene:
This is consistent with the behaviour of the Create PipelineRun page, so if we want to change this we should change both.

AlanGreene:
This is working as designed, but I agree it would be better to default the selection to the filtered Pipeline / Task but still allow the user to change this in the dialog without having to close, remove the filter, and open the dialog again.

PR #1415, opened 2020-05-20
Update Create PipelineRun and Create TaskRun dialog behaviour

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1409

Update the Create PipelineRun and Create TaskRun dialogs so that
when a user opens the dialog from a filtered list, they still
have the ability to choose a different Pipeline / Task without
having to close the dialog, clear the filter, and open the dialog
again.

As part of this change I've also simplified some of the logic
around populating the dialog, by consolidating all of the required
information into the component state instead of splitting some
of it out to props for the filtered case.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/CreatePipelineRun/CreatePipelineRun.test.js
+++ b/src/containers/CreatePipelineRun/CreatePipelineRun.test.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Tekton Authors
+Copyright 2019-2020 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -151,7 +151,7 @@ const testStore = {
 };
 
 const props = {
-  open: true,
+  open: false,
   namespace: 'namespace-1'
 };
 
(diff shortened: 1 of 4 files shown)


src/containers/CreatePipelineRun/CreatePipelineRun.test.js
```diff
@@ -488,32 +528,6 @@ describe('CreatePipelineRun', () => {
     expect(getByPlaceholderText(/select serviceaccount/i)).toBeTruthy();
   });
 
-  it('resets Pipeline and ServiceAccount when namespace changes controlled', async () => {
```
JustinKuli, line 491: Who don't we want it reset in this situation anymore?

AlanGreene: We do, but there's no difference in behaviour anymore between controlled / uncontrolled for this scenario so this test was adding no additional value. It's already covered by the test above it.

src/containers/CreateTaskRun/CreateTaskRun.js
```diff
@@ -450,61 +406,57 @@ class CreateTaskRun extends React.Component {
...
+              id="create-taskrun--namespaces-dropdown"
+              invalid={validationError && !namespace}
+              invalidText={intl.formatMessage({
+                id: 'dashboard.createTaskRun.invalidNamespace',
+                defaultMessage: 'Namespace cannot be empty'
+              })}
+              selectedItem={namespace ? { id: namespace, text: namespace } : ''}
+              onChange={this.handleNamespaceChange}
+            />
+            {this.state.kind === 'Task' && (
```
steveodonovan, line 432: pull out this status.kind and error fields on line 345 ?

AlanGreene: yep, updated 👍

Issue #1388, opened 2020-05-18
Run dashboard as a client application


AlanGreene:
# Expected behavior
I would like to have a dashboard experience running as a client app without requiring a new service running in my cluster. This should also ensure the application is only running with the permissions that I have in the cluster.

# Actual behavior
The Dashboard currently requires a service running in the cluster, and provides the same level of access to all users, i.e. all users can act with the same permissions granted to the dashboard service account.

# Additional Info
There appear to be 3 main categories of kubernetes UI:
1. services running in the cluster (e.g. kubernetes dashboard / Tekton Dashboard)
2. kubectl plugins, e.g. [kui](https://github.com/IBM/kui/)
3. standalone client app, e.g. [Octant](https://github.com/vmware-tanzu/octant)

Since the Tekton Dashboard is mostly just using the kubernetes APIs directly with very few of its own custom APIs, it should be possible to provide something similar to 2.

In fact, the current Dashboard front end code can be used unmodified as a frontend for `kubectl proxy` with much of the basic read-only functionality working.

Demoed on the working group call 2020/05/18

Steps:
1. checkout Dashboard repo
2. follow dev instructions to set up environment
3. `npm run build` (puts static frontend resources in `<path_to_dashboard_project>/dist/`)
4. `kubectl proxy --port 9876 --www="<path_to_dashboard_project>/dist" --www-prefix=/ --api-prefix=/proxy`
5. launch browser at `http://localhost:9876`
6. enjoy 😸 

Limitations with unmodified code:
- read-only
  - we currently use the install properties API provided by the dashboard backend to detect whether we should display Import Resources, Create PipelineRun, etc.
- no real-time updates
  - we currently use a custom websocket implementation provided by the Dashboard backend
  - it's possible to use websockets through `kubectl proxy` directly with some changes to the Dashboard frontend code, it may be possible to detect this and automatically switch approaches but would need further investigation

With some minor modifications to the code we get much closer to the full dashboard experience:
- switch the websocket to use the kubernetes watch APIs on the resource type of interest for any given page
- switch default value for 'read only' attribute in redux store from true to false to enable Import Resources, Create PipelineRun, etc.

[...]

PR #1416, opened 2020-05-20
pkg/endpoints: forward Impersonate headers in ProxyRequest

gpaul:
# Changes

Forward the `Impersonate-User`, `Impersonate-Group` and `Authorization` headers when proxying requests to the apiserver. See https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation

This partially addresses the privilege escalation where users with limited privileges, who can view the Tekton dashboard, essentially have the same privileges as the dashboard itself via the `/proxy` endpoint. 

There remain endpoints that still amount to privilege escalation: a user with read-only permissions to a given namespace can still re-trigger pipelines, as just one example.

This PR assumes that the Tekton dashboard will be fronted by an authenticating reverse proxy that forwards end-users' OIDC `email` and `group` claims in the `Impersonate-User` and `Impersonate-Group` headers, respectively.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/endpoints/cluster.go
+++ b/pkg/endpoints/cluster.go
@@ -61,11 +61,35 @@ func (r Resource) ProxyRequest(request *restful.Request, response *restful.Respo
 	uri := request.PathParameter("subpath") + "?" + parsedURL.RawQuery
 	forwardRequest := r.K8sClient.CoreV1().RESTClient().Verb(request.Request.Method).RequestURI(uri).Body(request.Request.Body)
 	forwardRequest.SetHeader("Content-Type", request.HeaderParameter("Content-Type"))
-	// Copy authorization (for serviceaccount clients) and impersonation headers (for user clients) from the client request to the forwarded request.
-	for k, vs := range request.Request.Header {
-		if k == "Impersonate-User" || k == "Impersonate-Group" || k == "Authorization" {
-			forwardRequest.SetHeader(k, vs...)
+	if r.Options.Impersonate {
+		// If the dashboard is running with Impersonate enabled, then proxy requests using the
+		// incoming requests Impersonate-* and Authorization headers. The Authorization header must
+		// correspond to a serviceaccount that has `impersonate` privileges, otherwise the apiserver
+		// will reject the request.
+		//
+		// In this mode, the intention is for the dashboard to be behind an authenticating reverse proxy
+		// that authenticates the user using an OIDC flow. The reverse proxy then attaches an
+		// Authorization header to the request containing its own authentication proof.
+		// It also adds Impersonate-* headers that list the identity claim (e.g., `sub`, `email`, etc.)
+		// and/or user group (i.e., `groups` claim) of the end-user.
+		//
+		// The apiserver receives the forwarded request, determines that the serviceaccount corresponding
+		// to the Authorization token has `impersonate` privileges (the reverse proxy must have
+		// `impersonate` privileges) and responds to the request as though it was performed by the
+		// account given by the Impersonate-User header and/or Impersonate-Group header.
+		//
+		// In this mode, the Authorization header must be set, it is important that the
+		// dashboard not perform this request using its own serviceaccount (the default behaviour), as
+		// that leads to a privilege escalation whereby a less-privileged user can access resources
+		// that the dashboard can access, but the user may not.
+		authz := request.Request.Header.Get("Authorization")
+		if authz == "" {
+			utils.RespondError(response, err, http.StatusForbidden)
+			return
 		}
+		forwardRequest.SetHeader("Authorization", authz)
+		forwardRequest.SetHeader("Impersonate-User", request.Request.Header.Get("Impersonate-User"))
+		forwardRequest.SetHeader("Impersonate-Group", request.Request.Header.Get("Impersonate-Group"))
 	}
 	forwardResponse := forwardRequest.Do()
 
--- a/pkg/endpoints/cluster.go
+++ b/pkg/endpoints/cluster.go
@@ -61,6 +61,12 @@ func (r Resource) ProxyRequest(request *restful.Request, response *restful.Respo
 	uri := request.PathParameter("subpath") + "?" + parsedURL.RawQuery
 	forwardRequest := r.K8sClient.CoreV1().RESTClient().Verb(request.Request.Method).RequestURI(uri).Body(request.Request.Body)
 	forwardRequest.SetHeader("Content-Type", request.HeaderParameter("Content-Type"))
+	// Copy authorization (for serviceaccount clients) and impersonation headers (for user clients) from the client request to the forwarded request.
+	for k, vs := range request.Request.Header {
+		if k == "Impersonate-User" || k == "Impersonate-Group" || k == "Authorization" {
+			forwardRequest.SetHeader(k, vs...)
+		}
+	}
 	forwardResponse := forwardRequest.Do()
 
 	if secretsURIPattern.Match([]byte(uri)) {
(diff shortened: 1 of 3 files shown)


pkg/endpoints/cluster.go
```diff
@@ -57,6 +57,12 @@ func (r Resource) ProxyRequest(request *restful.Request, response *restful.Respo
 	uri := request.PathParameter("subpath") + "?" + parsedURL.RawQuery
 	forwardRequest := r.K8sClient.CoreV1().RESTClient().Verb(request.Request.Method).RequestURI(uri).Body(request.Request.Body)
 	forwardRequest.SetHeader("Content-Type", request.HeaderParameter("Content-Type"))
+	// Copy authorization (for serviceaccount clients) and impersonation headers (for user clients) from the client request to the forwarded request.
+	for k, vs := range request.Request.Header {
```
AlanGreene, line 61: @steveodonovan @a-roberts thoughts on this? Personally I think I'd rather just copy all headers rather than cherry-picking, but looks like we used to do that and removed it for some reason.

Steve, can you remember why you removed this? Was it just the issue with the logs and the Accept header or was there more to it? https://github.com/tektoncd/dashboard/commit/911a81666ee736c6c559c7df26c88b7ea2f26c86#diff-13330ca6ce530de93565ddc93380b446L41

eddycharly:
Isn’t there a risk that it facilitates privilege escalation too ?

Does it require that all users exist in the cluster to be able to use the dashboard ?

pkg/endpoints/cluster.go
AlanGreene: Spoke with Steve about this and tested out proxying all headers. The log issue is still present when sending the `Accept` header:
`{"level":"error","msg":"Error: the server was unable to respond with a content type that the client supports"}`

Omitting this seems fine.

We also need to verify the default behaviour without an OIDC authenticator configured on the cluster and ensure any impersonation requests are rejected.

gpaul:
@AlanGreene 

> We also need to verify the default behaviour without an OIDC authenticator configured on the cluster and ensure any impersonation requests are rejected.

Indeed - I believe a simple CLI flag `--proxy-impersonate-headers`, or a more generic but less user-friendly `--proxy-header="Impersonate-Group" --proxy-header="Impersonate-User" --proxy-header="Authorization"` slice CLI option would do.

@eddycharly 

> Isn’t there a risk that it facilitates privilege escalation too ?

Definitely, if the dashboard is not deployed behind a reverse proxy that manages the `Impersonate-*` headers, then those headers should not be trusted and therefore no forwarded.

> Does it require that all users exist in the cluster to be able to use the dashboard ?

With the current design, if the headers aren't specified, then no impersonation is performed and the request is served using the `tekton-dashboard` serviceaccount's privileges.

If, however, the dashboard is hosted behind an authenticating reverse proxy that sets the `Impersonate-*` headers on the request to the dashboard, then the user's identity (i.e., `sub` or `email` claim), or a group that it belongs to (i.e., `groups` claim) would need to be granted explicit RBAC permissions in the cluster.

eddycharly:
A regular user could push a task run that could call the dashboard service directly (not going through the reverse proxy because the pod is running inside the cluster) impersonating an admin ?

Can we ensure all incoming traffic is actually going through the reverse proxy ?

AlanGreene:
From the docs:
>- A user makes an API call with their credentials and impersonation headers.
>- API server authenticates the user.
>- API server ensures the authenticated users have impersonation privileges.

So the user must already have permissions to impersonate for the desired user/group/resource, and must be authenticated to the API server. How could this be abused to achieve privilege escalation?

I would expect the API server to reject any impersonation attempt if the user cannot be authenticated (whether missing credentials, or if no suitable authenticator is configured), or if they have not already been granted the required privileges for impersonation.

gpaul:
> A regular user could push a task run that could call the dashboard service directly (not going through the reverse proxy because the pod is running inside the cluster) impersonating an admin ?

Correct, a user who is allowed to launch tasks could do that.

> Can we ensure all incoming traffic is actually going through the reverse proxy ?

One might use a NetworkPolicy to restrict incoming connections to the tekton-dashboard to those that originated at the reverse proxy. https://kubernetes.io/docs/concepts/services-networking/network-policies/

gpaul:
@AlanGreene the attack vector is as follows:

By default, a pod in the cluster can reach any service in the cluster via cluster DNS, e.g., https://tekton-dashboard.default/

That would not go through the ingress controller / reverse proxy, as the request is originating inside the cluster.

If a user is allowed to launch tasks, e.g., pods, then that user can send HTTP requests to http://tekton-dashboard.default/...proxy/...secrets/ and set the `Imporsonate-User: admin` HTTP header on their request.

The tekton dashboard will receive the request, notice the Impersonate-User header, and simply forward it to the kube-apiserver, which will honour it and respond with the privileged data.

AlanGreene:
I think I'm still missing some piece of the puzzle here.

The authenticated user would be the dashboard service account in that case, and it does not have impersonation privileges so I would expect the API server to reject the request.

If the API server is blindly trusting any `Impersonate-*` header it receives then it's acting contrary to the documentation and there are much bigger problems to worry about even without the dashboard in the picture at all.

gpaul:
Ah yes - the dashboard would need impersonation privileges.

gpaul:
If the dashboard is the one making the request to the apiserver, it doesn't help that the end user has impersonation privileges. The dashboard needs to have them.

gpaul:
Actually, the `Authorization` header of the reverse proxy should also be forwarded. Then the impersonating user is not the dashboard, but the reverse proxy. In the case of a malicious user active inside the cluster, the malicioius pod would have to run as a serviceaccount that has impersonation privileges.

gpaul:
I've updated the PR:
- I've threaded an "Impersonate" option through from `main()`. It is false unless `IMPERSONATE=true` is specified in the environment.
- I've made the `Authorization` header required when Impersonation is enabled, as per https://github.com/tektoncd/dashboard/pull/1416#issuecomment-632422577

gpaul:
There's the case of a malicious pod in the cluster not providing any of the headers, in which case the request is served using the dashboard serviceaccount which results in escalated privilege.

AlanGreene:
I wonder if a network policy would be more suitable here to prevent untrusted access to the dashboard, so only the auth proxy would be allowed.

One of my concerns is that the more configuration options we add to the dashboard the more complex the test story becomes, and also the harder it becomes to ensure we're not inadvertently creating new gaps. There are already well defined and understood kubernetes approaches to limiting access within the cluster and we should consider if some of those would be more appropriate and perhaps more flexible here rather than trying to implement something custom in the dashboard itself.

I'm curious what others think about this, especially in light of some of the other work in progress around limiting dashboard access / functionality.

eddycharly:
I have a few concerns too.

I feel like it will be hard to make a good user experience from this, mainly because we don't know what a user can or cannot do, so we cannot hide actions and menu items that are not allowed etc...

It probably won't work well with the backend controllers, there will be no consistency between the impersonated role and the role used by the controllers, the web socket will not be inline with the current user permissions.

Another concern is that it adds quite a bit of complexity, if we want this to make it, it should be well documented and possible misconfigurations should be explained in details so that end users don't make mistakes when enabling impersonation.

One more concern i have is that the proxy exposes the api server on an external network, it's not directly related to this PR but exposing the control plane is something i'm not comfortable with from the beginning. Adding the possibility to use the exposed proxy combined with impersonation gives me cold sweats.

Read-only and single namespace modes are not a perfect solution but at least it should help to limit the blast radius in a controlled and consistent way.

For sure i would love to see authn/authz supported in the dashboard, i really find the attempt here interesting but i feel like it's a big feature that will go beyond just enabling impersonation headers and setting up a reverse proxy.

It's just my personal opinions, very curious about what others think and quite happy that the topic gets more discussions.

AlanGreene:
Closing this PR, but happy to review / discuss further in a design proposal that addresses the concerns above. I do think the user impersonation approach is worth exploring, and a solution would likely overlap with some of the front end work required for https://github.com/tektoncd/dashboard/issues/1388 as well as some enhancements to our recent changes for single namespace visibility.

PR #1425, opened 2020-05-21
Add frontend support for single namespace visibility

eddycharly:
# Changes

This PR adds frontend support for single namespace visibility.

Single namespace or full cluster visibility is driven by the backend returning the `TenantNamespace` property.

When the `TenantNamespace` property is not set, the visibility is considered to be for the whole cluster. The list of namespaces is fetched and the namespace drop down is shown with the `All namespaces` option available.

When the `TenantNamespace` property is not set, the visibility is considered scoped to the tenant namespace itself. The list of namespaces is not fetched and the namespace drop down is not shown at all. The  selected namespace in the store is forced to the tenant namespace. Therefore all API calls are run against the tenant namespace.

This introduces the following change in extensions management (both resource based and service based):
- cluster scope deployment -> all extensions in the cluster are considered
- namespace scope deployment -> only extension in the tenant namespace are considered

I added a simple overlay to test:
```
kubectl create ns tekton-tenant
kustomize build --load_restrictor none overlays/dev-single-namespace | ko apply -f -
```

/cc @AlanGreene @a-roberts 

Extracted from the giant PR https://github.com/tektoncd/dashboard/pull/1371

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -21,15 +21,16 @@ import {
   Switch
 } from 'react-router-dom';
 
-import { IntlProvider } from 'react-intl';
-import { Content } from 'carbon-components-react';
+import { injectIntl, IntlProvider } from 'react-intl';
+import { Content, InlineNotification } from 'carbon-components-react';
 
 import {
   Header,
+  LoadingShell,
   LogoutButton,
   PageErrorBoundary
 } from '@tektoncd/dashboard-components';
-import { paths, urls } from '@tektoncd/dashboard-utils';
+import { getErrorMessage, paths, urls } from '@tektoncd/dashboard-utils';
 
 import {
   About,
@@ -75,6 +76,7 @@ import {
   getLocale,
   getLogoutURL,
   getSelectedNamespace,
+  getTenantNamespace,
   isReadOnly,
   isWebSocketConnected
 } from '../../reducers';
@@ -104,16 +106,61 @@ if (process.env.I18N_PSEUDO) {
   });
 }
 
-export /* istanbul ignore next */ class App extends Component {
-  componentDidMount() {
-    this.fetchData();
+const ConfigErrorComponent = ({ intl, loadingConfigError }) => {
+  if (!loadingConfigError) {
+    return null;
   }
 
-  componentDidUpdate(prevProps) {
+  return (
+    <InlineNotification
+      kind="error"
+      title={intl.formatMessage({
+        id: 'dashboard.app.loadingConfigError',
+        defaultMessage: 'Error loading configuration'
+      })}
+      subtitle={getErrorMessage(loadingConfigError)}
+      lowContrast
+    />
+  );
+};
+
+const ConfigError = injectIntl(ConfigErrorComponent);
+
+const initialState = {
+  loadingConfigError: null,
+  loadingConfig: true
+};
+
+export /* istanbul ignore next */ class App extends Component {
+  state = initialState;
+
+  componentDidMount() {
+    this.fetchConfig();
+  }
+
+  componentDidUpdate(prevProps, prevState) {
     const { webSocketConnected } = this.props;
     const { webSocketConnected: prevWebSocketConnected } = prevProps;
+
     if (webSocketConnected && prevWebSocketConnected === false) {
-      this.fetchData();
+      this.fetchConfig();
+    }
+
+    const { loadingConfig } = this.state;
+    const { loadingConfig: prevLoadingConfig } = prevState;
+
+    if (prevLoadingConfig && loadingConfig === false) {
+      if (this.props.tenantNamespace) {
+        this.props.selectNamespace(this.props.tenantNamespace);
+      }
+
+      if (!this.props.tenantNamespace) {
+        this.props.fetchNamespaces();
+      }
+
+      this.props.fetchExtensions({
+        namespace: this.props.tenantNamespace
+      });
     }
   }
 
@@ -121,14 +168,20 @@ export /* istanbul ignore next */ class App extends Component {
     this.props.onUnload();
   }
 
-  fetchData() {
-    this.props.fetchExtensions();
-    this.props.fetchInstallProperties();
-    this.props.fetchNamespaces();
+  async fetchConfig() {
+    this.setState({ loadingConfig: true });
+    try {
+      await this.props.fetchInstallProperties();
+      this.setState({ loadingConfig: false });
+    } catch (error) {
+      console.error(error); // eslint-disable-line no-console
+      this.setState({ loadingConfigError: error, loadingConfig: false });
+    }
   }
 
   render() {
     const { extensions } = this.props;
+    const { loadingConfigError, loadingConfig } = this.state;
 
     const lang = messages[this.props.lang] ? this.props.lang : 'en';
     const logoutButton = (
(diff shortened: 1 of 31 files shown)


eddycharly:
/hold

Based on https://github.com/tektoncd/dashboard/pull/1411, should be merged after.

eddycharly:
I changed service based extensions behavior:
- cluster scope deployment -> only extension in the install namespace are considered
- namespace scope deployment -> only extension in the tenant namespace are considered

I'm not particularly happy with this as it's not consistent with resource extensions.
How about this instead ?
- cluster scope deployment -> all extensions in the cluster are considered

src/containers/App/App.js
```diff
@@ -104,26 +115,65 @@ export /* istanbul ignore next */ class App extends Component {
...
     this.props.onUnload();
   }
 
-  fetchData() {
-    this.props.fetchExtensions();
-    this.props.fetchInstallProperties();
-    this.props.fetchNamespaces();
+  async fetchData() {
+    try {
+      const properties = await this.props.fetchInstallProperties();
```
eddycharly, line 129: I'm not happy with this, directly using the result of the action looks bad isn't it ?

eddycharly: I guess the action should dispatch to the store and the container should in turn react to store change...

eddycharly: Changed it, should be better hopefully.

eddycharly:
I added unit tests, it should be ready for review.

src/containers/App/App.js
```diff
@@ -93,39 +94,91 @@ if (process.env.I18N_PSEUDO) {
...
   render() {
     const { extensions } = this.props;
+    const { loadingConfigError, loadingConfig } = this.state;
 
     const lang = messages[this.props.lang] ? this.props.lang : 'en';
     const logoutButton = (
       <LogoutButton getLogoutURL={() => this.props.logoutURL} />
     );
 
+    if (loadingConfig) {
```
AlanGreene, line 159: General comment / question before I review the rest in detail.

Do we need this global loading state? Is there something preventing us from beginning to render the application shell at least, and handling the loading state locally to the components that depend on the install properties?

eddycharly: Not sure, maybe it could be structured differently but we need to load the config first before deciding if we have to fetch namespaces and adapt fetching extensions.

AlanGreene: It should be fine to get rid of this branch in the render, allow the app shell to start rendering, and we already respond to extension updates from the store.

`fetchConfig` also needs a small change to wait for the install properties to be loaded:

```diff
 async fetchConfig() {
     this.setState({ loadingConfig: true });
     try {
-      this.props.fetchInstallProperties();
+      await this.props.fetchInstallProperties();
       this.setState({ loadingConfig: false });
     } catch (error) {
       console.error(error); // eslint-disable-line no-console
       this.setState({ loadingConfigError: error, loadingConfig:
 false });
     }
   }
```

Since there's only a single namespace available to users when tenantNamespace is set we should also replace the dropdown in SideNav with text:
image

eddycharly: Good catch with the missing `await`.
I will remove the `loadingConfig` branch in render to see how it goes.

I'm working on reorganizing `kustomize` bases/overlays right now, I will take care of this asap.

eddycharly: If the app renders before the config is fetched, the user is redirected to the pipeline runs page and a call to load pipeline runs at the cluster level is initiated.

Of course this call fails in single namespace mode.

How should we handle this case ?

eddycharly: I added the missing `await` and improved the sidenav to show the namespace when tenant namespace is set.

eddycharly:
/hold cancel

Backend support was merged

src/containers/App/App.js
eddycharly: @AlanGreene any advice on how to prevent blocking rendering until config is loaded ?

AlanGreene: hrmm interesting problem alright...

I'll take a look and see what options we have. I'd really like to avoid a full-page loading state if possible, or at least restrict it to only the routes affected.

AlanGreene: It looks like we already have this problem intermittently on the Secrets / ServiceAccounts pages (users gets redirected to PipelineRuns), and I'd expect to encounter it with extensions too but haven't managed to reproduce it there yet.

I've been trying out some changes to the Route components to see if we can handle the access checks more locally, building on what I already did with the ReadWriteRoute (which is working around some annoying react-router behaviour when dynamically adding routes) but so far it's not making things any simpler or really improving the experience.

We're going to need to handle more granular access control anyway, with each Route responsible for checking whether/how it should render its content. I have some thoughts that might be worth discussing / expanding upon at the WG but don't want to hold up the single namespace change (and other changes behind it...) until this is all resolved.

Maybe we can go with the full-page loading experience for now, just style it so it's not quite so bare, and do our best to improve it and minimise the instances where it's shown until we have a more complete / flexible solution in place.

eddycharly: I will try to improve style a little bit. Let's see what i can do with css :P

AlanGreene: We could use the Carbon Loading component with the overlay, and see if we can add a string below the spinner 'Loading configuration...' or something.

The loading component should not cover the top navbar.

https://www.carbondesignsystem.com/patterns/loading-pattern#full-screen-loading
https://react.carbondesignsystem.com/?path=/story/loading--default

eddycharly: I added the `Loading` component while fetching config.
There was no option to add text below the spinner though.

AlanGreene: Played around with this a bit and came up with something I think works well enough for now: https://github.com/tektoncd/dashboard/pull/1464

image

```diff
diff --git a/src/containers/App/App.js b/src/containers/App/App.js
index b9a4349..627f123 100644
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -22,10 +22,12 @@ import {
 } from 'react-router-dom';
 
 import { IntlProvider } from 'react-intl';
-import { Content, InlineNotification, Loading } from 'carbon-components-react';
+
+import { Content, InlineNotification } from 'carbon-components-react';
 
 import {
   Header,
+  LoadingShell,
   LogoutButton,
   PageErrorBoundary
 } from '@tektoncd/dashboard-components';
@@ -165,18 +167,6 @@ export /* istanbul ignore next */ class App extends Component {
       <LogoutButton getLogoutURL={() => this.props.logoutURL} />
     );
 
-    if (loadingConfig) {
-      return (
-        <IntlProvider
-          locale={lang}
-          defaultLocale="en"
-          messages={messages[lang]}
-        >
-          <Loading description="Loading configuration..." withOverlay />
-        </IntlProvider>
-      );
-    }
-
     return (
       <IntlProvider locale={lang} defaultLocale="en" messages={messages[lang]}>
         {loadingConfigError && (
@@ -188,6 +178,8 @@ export /* istanbul ignore next */ class App extends Component {
           />
         )}
 
+        {loadingConfig && <LoadingShell />}
+        {!loadingConfig && (
           <Router>
             <>
               <Header logoutButton={logoutButton} />
@@ -433,6 +425,7 @@ export /* istanbul ignore next */ class App extends Component {
               </Content>
             </>
           </Router>
+        )}
       </IntlProvider>
     );
   }
```

Let me know what you think.

eddycharly: Wow that’s gold !!
Will try it this morning, thank you so much !

eddycharly: Changed to use `LoadingShell`, works perfectly ! Thanks

src/containers/App/App.js
```diff
@@ -135,251 +168,263 @@ export /* istanbul ignore next */ class App extends Component {
...
-                  <Redirect to={urls.pipelineRuns.all()} />
-                </Switch>
-              </PageErrorBoundary>
-            </Content>
-          </>
-        </Router>
+        {loadingConfigError && (
+          <InlineNotification
+            kind="error"
+            title="Error getting data"
```
AlanGreene, line 174: will need to use `intl` for this message, also maybe 'Error loading configuration' for consistency with the loading state and to make it clearer what failed

src/containers/NamespacesDropdown/NamespacesDropdown.js
```diff
@@ -17,7 +17,11 @@ import { connect } from 'react-redux';
 import { ALL_NAMESPACES } from '@tektoncd/dashboard-utils';
```
AlanGreene, line 17: Do we need to make changes to the NamespacesDropdown? If we're only giving access to a single namespace a Dropdown probably doesn't make sense in most cases, and would either be hidden or replaced with text showing the tenant namespace.

If we do need to display a list of namespaces (only one in this case), would it make more sense to handle this in the `getNamespaces` selector so it's transparent to most consumers? There should be very few places in the code that need to be aware of the tenant namespace specifically.

src/containers/SideNav/SideNav.js
```diff
@@ -238,15 +239,21 @@ class SideNav extends Component {
...
-            element={NamespacesDropdown}
-            id="sidenav-namespace-dropdown"
-            selectedItem={{ id: namespace, text: namespace }}
-            showAllNamespaces
-            onChange={this.selectNamespace}
-          >
-            &nbsp;
-          </SideNavMenuItem>
+          {this.props.tenantNamespace ? (
+            <SideNavMenu defaultExpanded title="Namepsace">
```
AlanGreene, line 243: using a menu here seems a little odd but let's leave it for now since we'll be moving the dropdown out of the side nav soon anyway

typo:
```suggestion
            <SideNavMenu defaultExpanded title="Namespace">
```

src/containers/SideNav/SideNav.test.js
```diff
@@ -618,3 +630,39 @@ it('SideNav does not render import in read-only mode', async () => {
   await waitForElement(() => queryByText(/about/i));
   expect(queryByText(/import/i)).toBeFalsy();
 });
+
+it('Namespace dropdown does not render in signle namespace visibility mode', async () => {
```
AlanGreene, line 634:
```suggestion
it('Namespace dropdown does not render in single namespace visibility mode', async () => {
```

src/containers/App/App.js
eddycharly: done

src/containers/SideNav/SideNav.js
eddycharly: fixed

src/containers/SideNav/SideNav.test.js
eddycharly: fixed

src/containers/NamespacesDropdown/NamespacesDropdown.js
eddycharly: Good point, i will look at it tomorrow.

eddycharly: @AlanGreene Is it ok if i improve that in a follow up PR ?

eddycharly:
Any chance to get this in soon ?
I rebased it so many times... 🤪

I will improve namespace dropdown in another pr if it’s ok, maybe introduce a namespace picker component that renders differently in single namespace or full cluster mode.

AlanGreene:
> I will improve namespace dropdown in another pr if it’s ok

Yes that's fine. Just giving it one last test and I'll approve it 👍

src/containers/App/App.js
```diff
@@ -424,5 +476,5 @@ export default hot(
   connect(
     mapStateToProps,
     mapDispatchToProps
-  )(App)
+  )(injectIntl(App))
```
AlanGreene, line 479: We can't use `injectIntl` here, it has to be on a descendant of the `IntlProvider`.

image

The easiest thing to do for now would be to extract the InlineNotification to a function that we can use as a component and wrap that with `injectIntl`.

At the top of App.js add:

```js
const ConfigErrorComponent = ({ intl, loadingConfigError }) => {
  if (!loadingConfigError) {
    return null;
  }

  return (
    <InlineNotification
      kind="error"
      title={intl.formatMessage({
        id: 'dashboard.app.loadingConfigError',
        defaultMessage: 'Error loading configuration'
      })}
      subtitle={getErrorMessage(loadingConfigError)}
      lowContrast
    />
  );
};

const ConfigError = injectIntl(ConfigErrorComponent);
```

Then replace the code that displays the notification with:

```diff
+        <ConfigError loadingConfigError={loadingConfigError} />
-        {loadingConfigError && (
-          <InlineNotification
-            kind="error"
-            title={intl.formatMessage({
-              id: 'dashboard.app.loadingConfigError',
-              defaultMessage: 'Error loading configuration'
-            })}
-            subtitle={getErrorMessage(loadingConfigError)}
-            lowContrast
-          />
-        )}
```

and remove the `injectIntl` on `App`

eddycharly: Done, thank you for the fix !

eddycharly:
Thank you @AlanGreene and @a-roberts for your great patience and support !

PR #1430, opened 2020-05-25
Use the redux store to get about informations

eddycharly:
# Changes

This PR fetches informations to display on the About page from the redux store instead of calling the API.

**Note:** I removed the error notification when some required properties were missing, i can put it back if you want, just let me know.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/About/About.js
+++ b/src/containers/About/About.js
@@ -12,43 +12,26 @@ limitations under the License.
 */
 
 import React, { Component } from 'react';
+import { connect } from 'react-redux';
 import { injectIntl } from 'react-intl';
 import { InlineNotification } from 'carbon-components-react';
 import { Table } from '@tektoncd/dashboard-components';
 import { getErrorMessage, getTitle } from '@tektoncd/dashboard-utils';
 
-import { getInstallProperties } from '../../api';
-
-const initialState = {
-  dashboardInfo: null,
-  error: null,
-  loading: true
-};
-
-const dashboardPropertiesToCheck = [
-  { property: 'DashboardNamespace', required: true, display: 'Namespace' },
-  { property: 'DashboardVersion', required: true, display: 'Version' },
-  { property: 'IsOpenShift' },
-  { property: 'ReadOnly' }
-];
-
-const pipelinesPropertiesToCheck = [
-  { property: 'PipelineNamespace', required: true, display: 'Namespace' },
-  { property: 'PipelineVersion', required: true, display: 'Version' }
-];
-
-const triggersPropertiesToCheck = [
-  { property: 'TriggersNamespace', display: 'Namespace' },
-  { property: 'TriggersVersion', display: 'Version' }
-];
-
-const propertiesToCheck = dashboardPropertiesToCheck
-  .concat(pipelinesPropertiesToCheck)
-  .concat(triggersPropertiesToCheck);
+import {
+  getDashboardNamespace,
+  getDashboardVersion,
+  getLogoutURL,
+  getPipelineNamespace,
+  getPipelineVersion,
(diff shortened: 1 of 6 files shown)


a-roberts:
Yeah, makes sense to do so and I'm gonna assume this is way more efficient, and reminiscent of how we do things for checking IsReadOnly - testing this now

eddycharly:
Alright, let's say the notifications make sense, i will add them back later after lunch.
Thanks !

src/containers/About/About.js
```diff
@@ -174,33 +98,58 @@ export /* istanbul ignore next */ class About extends Component {
...
             headers={headers}
-            rows={filteredRows(dashboardPropertiesToCheck)}
-            loading={loading}
+            rows={[
+              getRow('Namespace', this.props.dashboardNamespace),
+              getRow('Version', this.props.dashboardVersion),
+              getRow('IsOpenShift', this.props.isOpenShift),
+              getRow('ReadOnly', this.props.isReadOnly),
+              getRow('LogoutURL', this.props.logoutURL)
+            ].filter(item => item)}
```
AlanGreene, line 107: `.filter(Boolean)` is a shorthand for this that we've used elsewhere including in this file

a-roberts:
> My only issue with this is that we're no longer guaranteed to display the latest information on the About page, only whatever was loaded by the initial request to get the install properties from App.js when the application was first loaded. There are no real time updates for this information.

That's what I thought too, but refreshing the page after deleting/installing the Triggers deployment presented no problems - all the information remained accurate 🤔 

I went through the usual process
```
git fetch origin refs/pull/1430/head
git checkout -b eddycharly/about FETCH_HEAD
npm run build_ko && npm run bootstrap
```

with kubectl proxy access on Docker Desktop, are you seeing something different?

AlanGreene:
No that's exactly what I mean, it requires a full page refresh to get the latest info whereas before we called the API whenever the About page was loaded so it was guaranteed fresh.

It's not just the About page that's affected though, so maybe something we need to handle outside this PR?

eddycharly:
Definitely, informations are not updated in real time, it will need a refresh to reload the App and populate the store with properties fetched from the api to get the new infos displayed here.

If we want real time update we should watch changes in the backend and notify the frontend to update the store.

Plug everything on the store and improve changes tracking seems the best approach to me.
Maybe i can look into it in another PR or do you prefer to close this one and keep the current implementation ?

AlanGreene:
We already have the problem in other parts of the app so this change isn't really making it worse, maybe just making it more obvious. We'll need to make this change to use the store anyway so I think I'm OK with getting this in and addressing the other issues separately.

src/containers/About/About.js
eddycharly: fixed

eddycharly:
Added back notifications.

src/reducers/properties.js
```diff
@@ -37,4 +41,24 @@ export function getDashboardNamespace(state) {
...
+}
+
+export function getTriggersNamespace(state) {
+  return state.TriggersNamespace;
+}
+
+export function getTriggersVersion(state) {
+  return state.TriggersVersion;
+}
+
```
AlanGreene, line 63: Since most of these are only used on the About page we could just add one new selector that returns the entire properties object.

The existing selectors (isReadOnly etc.) are mostly a convenience since those values are used in multiple places.

I'm OK with either approach, but it would be nice to see some tests for the new selectors (we've had a couple of typos in selectors that went unnoticed before due to missing tests).

eddycharly: Added the missing tests

PR #1433, opened 2020-05-26
Stop creation of non-existent Pipelines being used in PipelineRuns

### LINKED ISSUE #1432: Can create a pipelinerun in a ns where that pipeline doesnt exist
# Expected behavior
Shouldnt happen 

# Actual behavior
I can create the pipelinerun 

# Steps to reproduce the problem

1. Click on label in pipelines
2. Change ns to a ns which doesnt have that pipeline
3. Click on create and press create 

# Environment
- Kubernetes Platform:

doc ker 

- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

master 

# Additional Info

CarolynMabbott:
# Changes
Fixes #1432 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/CreatePipelineRun/CreatePipelineRun.js
+++ b/src/containers/CreatePipelineRun/CreatePipelineRun.js
@@ -318,15 +318,20 @@ class CreatePipelineRun extends React.Component {
   };
 
   initialState = () => {
-    const { namespace, pipelineRef } = this.props;
+    const { namespace } = this.props;
+    let { pipelineRef } = this.props;
     const pipelineInfo = parsePipelineInfo(pipelineRef, namespace);
+    if (pipelineInfo.pipelineError) {
+      pipelineRef = '';
+    }
     return {
       ...initialState,
       ...pipelineInfo,
       namespace: namespace !== ALL_NAMESPACES ? namespace : '',
       pipelineRef: pipelineRef || '',
       params: initialParamsState(pipelineInfo.paramSpecs),
-      resources: initialResourcesState(pipelineInfo.resourceSpecs)
+      resources: initialResourcesState(pipelineInfo.resourceSpecs),
+      pipelineError: ''
     };
   };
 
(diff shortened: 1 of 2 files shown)


src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -318,15 +318,24 @@ class CreatePipelineRun extends React.Component {
   };
 
   initialState = () => {
-    const { namespace, pipelineRef } = this.props;
+    const { namespace } = this.props;
+    let { pipelineRef } = this.props;
     const pipelineInfo = parsePipelineInfo(pipelineRef, namespace);
+    if (
```
a-roberts, line 324: This whole block here doesn't look right... are you sure there's not a neater way of doing things?

Surely pipelineRef is nothing in the case of the Pipeline no longer exists, so why do we need to do the undefined checks as well

CarolynMabbott: I wasnt sure if getting a pipelineError is not definitive of the pipeline not existing, I tihnk you can get these for other reasons 
I wanted to add an extra check(s) to make sure it wasnt just an error, I could change it to just check if there is an error then set it to empty?

a-roberts: But if there's an error it doesn't matter what paramSpecs and resourceSpecs are surely, so you could just omit those - did you see any value in adding them?

CarolynMabbott: I have taken them out and left it as just the error, I dont think they would add any value

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -318,15 +318,20 @@ class CreatePipelineRun extends React.Component {
   };
 
   initialState = () => {
-    const { namespace, pipelineRef } = this.props;
+    const { namespace } = this.props;
+    let { pipelineRef } = this.props;
     const pipelineInfo = parsePipelineInfo(pipelineRef, namespace);
+    if (pipelineInfo.pipelineError === true) {
```
a-roberts, line 324: It's a nit but you need to fix the commit message typo as well please (go with the description I've got for this PR), can this just be ` if (pipelineInfo.pipelineError === true) {`?

Otherwise all good

PR #1436, opened 2020-05-26
Namespace should not be translated

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/453

The word 'Namespace' should not be translated as it's a
Kubernetes kind, and should be treated similar to the others,
e.g. PipelineRun, Task, ServiceAccount

This also resolves a react-intl warning about missing messages
for consumers who are not exposing the namespace column in our
tables, e.g.:

```
[React Intl] Missing message: "dashboard.tableHeader.namespace"
for locale: "fr", using default message as fallback.
```

They shouldn't have to provide a value for a feature they're
not using so this addresses both issues in one. It also gives us
consistency as we were already not translating it in some places.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/PipelineResources/PipelineResources.js
+++ b/packages/components/src/components/PipelineResources/PipelineResources.js
@@ -39,10 +39,7 @@ const PipelineResources = ({
     },
     {
       key: 'namespace',
-      header: intl.formatMessage({
-        id: 'dashboard.tableHeader.namespace',
-        defaultMessage: 'Namespace'
-      })
+      header: 'Namespace'
     },
     {
       key: 'type',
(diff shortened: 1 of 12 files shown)


AlanGreene:
Where 'Namespace' appears as part of another string we'll still need to include it in messages.json, that includes the 'Namespace:' label for the Trigger interceptors, as the surrounding string may change, e.g. in some languages `:` may be substituted for another character, moved to the start of the string, or removed altogether.

We would provide a list of terms that should not be translated when used in certain contexts.

AlanGreene:
@a-roberts this should be good to go

Issue #1276, opened 2020-04-09
Sidenav doesnt auto update for triggers install / uninstall


CarolynMabbott:
# Expected behavior
Sidenav updates 

# Actual behavior
Doesnt update 

# Steps to reproduce the problem

1. Install triggers 
2.
3.

# Environment
- Kubernetes Platform:

docker 
- Tekton Pipelines version:

- Tekton Triggers version:

- Tekton Dashboard version:

0.6.1
# Additional Info

AlanGreene:
If we want to do this, one option would be to add a websocket handler for CRDs and detect when any of the triggers CRDs are added / deleted.

eddycharly:
I’ll look at it, most of the UI has been plugged on the redux store, it’s just a matter of notifying the frontend when something changes. 

/assign

PR #1446, opened 2020-05-27
Periodically refresh install properties

eddycharly:
# Changes

This PR periodically refreshes install properties (every 60s).
This allows updating the frontend when something changes on the backend side.

It's a quick and easy way of detecting changes. Something more advanced like using the websocket would require more work and would probably involve watching deployments in potentially various namespaces.

What do you think ? It seems to me that something simple like this requires almost no change and is acceptable in this case.

Closes https://github.com/tektoncd/dashboard/issues/1276

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -96,6 +96,10 @@ if (process.env.I18N_PSEUDO) {
 export /* istanbul ignore next */ class App extends Component {
   componentDidMount() {
     this.fetchData();
+    this.interval = setInterval(
+      () => this.props.fetchInstallProperties(),
+      60000
+    );
   }
 
   componentDidUpdate(prevProps) {
@@ -107,6 +111,7 @@ export /* istanbul ignore next */ class App extends Component {
   }
 
   componentWillUnmount() {
+    clearInterval(this.interval);
     this.props.onUnload();
   }
 


AlanGreene:
We already have websocket events for extensions, and a change to the dashboard config (e.g. switching to readonly / tenant mode) would require a redeploy so we should already be handling this through the websocket disconnect/reconnect which triggers us to re-fetch the data.

So the only case that's not handled by this is detecting Triggers / Pipelines updates, which would require watching deployments.

Is my understanding correct?

eddycharly:
That’s correct. Watching pipelines and triggers deployment is not easy though. They can be deployed in separate namespaces or not deployed at all, the namespace could not even exist.

And when you look at the code to find the version it’s definitely a mess. I feel like watching for this on the backend is not as easy as it looks like.

AlanGreene:
I agree, definitely not as easy as it should be.

Since installing / uninstalling Triggers isn't going to happen very often, and updates of Triggers or Pipelines should be similarly infrequent, I wonder if we need this at all?

While it would be nice to always display the latest data, I'm not a huge fan of frequent polling for something that's going to change so rarely. Maybe we could reduce the frequency, but then at what point does it become easier for the user to just refresh the page rather than wait for the UI to update itself?

eddycharly:
I agree, and i don't mind closing this PR if we don't want to support this use case (that's just two lines of code ;-)

I just think it's not worth the effort to do it backend side and notifying though the web socket, at least not with the current level of complexity implemented in the backend.

eddycharly:
@AlanGreene any more thoughts ? Do you prefer that i close this PR ?

AlanGreene:
Yes let's close this for now, thanks.

PR #1454, opened 2020-06-01
Switch to go modules

eddycharly:
# Changes

This PR switches to go modules for the backend.
This should lower the requirements to build the dashboard backend (no more dep).
This also allows to checkout outside `GOPATH`.

# Submitter Checklist

review them:

for more details._


eddycharly:
/cc @a-roberts @AlanGreene 

Should be ready for review.

a-roberts:
Yeah I think this is a good one to do particularly for the GOPATH reasons, it's been done for various other repositories under the Tekton space so I'll be eager to try things out

test/presubmit-tests.sh
```diff
@@ -26,7 +26,7 @@
 export DISABLE_MD_LINTING=1
 
 # FIXME(vdemeester) we need to come with something better (like baking common scripts in our image, when we got one)
-dep ensure || exit 1
+go mod vendor || exit 1
```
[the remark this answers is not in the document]
eddycharly: Yes, else the source command in e2e-*.sh will fail.
We need this to get tektoncd/plumbing into vendor.

test/e2e-common.sh
```diff
@@ -32,7 +32,11 @@ function print_diagnostic_info() {
 
 function install_kustomize() {
   echo ">> Installing kustomize"
-  GO111MODULE=on go get sigs.k8s.io/kustomize/kustomize/v3@v3.5.4
+  tar=kustomize_v3.6.1_linux_amd64.tar.gz
```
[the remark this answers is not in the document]
eddycharly: go get failed, it couldn’t compile kustomize locally, probably a side effect of go.mod file :(

eddycharly:
Dockerfiles should be fixed.

eddycharly:
Maybe we should add a CI job to build Dockerfile, Dockerfile_test and Dockerfile_race_test ?

AlanGreene:
I'm actually not sure why we have Dockerfile_test and Dockerfile_race_test. Running `go test` and `go test -race` directly seem to work fine for me, and `presubmit-tests.sh` is doing this already in CI. Maybe this wasn't always the case? @a-roberts can you shed some light on this?

.dockerignore
```diff
@@ -1,3 +1,18 @@
...
+config_frontend
+coverage
 dist
+docs
+ingress
+node_modules
+overlays
+packages/components/dist
+packages/components/node_modules
+packages/utils/dist
```
AlanGreene, line 14: can we just ignore `packages` entirely so we don't need to worry about this if we add more frontend packages in future?

eddycharly: I don't think we can ignore `packages/src` as we need that to build the frontend in the Dockerfile.

AlanGreene: oops, yeah that could be tough to do without the source 😆

a-roberts:
> I'm actually not sure why we have Dockerfile_test and Dockerfile_race_test. Running `go test` and `go test -race` directly seem to work fine for me, and `presubmit-tests.sh` is doing this already in CI. Maybe this wasn't always the case? @a-roberts can you shed some light on this?

Can't think of a good justification tbh, good to go I say

.dockerignore
eddycharly:
```
**/.git
**/dist
**/node_modules
```
maybe ?

AlanGreene: let's leave it as-is for now, looks like it doesn't follow the same format as .gitignore 😞

eddycharly: https://docs.docker.com/engine/reference/builder/#dockerignore-file

> Beyond Go’s filepath.Match rules, Docker also supports a special wildcard string ** that matches any number of directories (including zero). For example, **/*.go will exclude all files that end with .go that are found in all directories, including the root of the build context.

PR #1459, opened 2020-06-03
Fix character encoding of displayed Secrets username column

AlanGreene:
# Changes

`window.atob` only supports ASCII, which means that usernames
with non-ASCII characters get garbled when displayed in the table.

Add a `base64Decode` implementation that handles an expanded
character set to ensure these are displayed correctly.

Before:
username with value '测试' displayed as:
image

After:
image

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/Secrets/Secrets.js
+++ b/src/containers/Secrets/Secrets.js
@@ -59,6 +59,25 @@ const initialState = {
   toBeDeleted: []
 };
 
+function base64Decode(input) {
+  return decodeURIComponent(
+    atob(input)
+      .split('')
+      .map(
+        character =>
+          '%' + // eslint-disable-line prefer-template
+          (
+            '00' + // eslint-disable-line prefer-template
+            character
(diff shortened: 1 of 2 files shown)


src/containers/Secrets/Secrets.js
```diff
@@ -59,6 +59,18 @@ const initialState = {
   toBeDeleted: []
 };
 
+function base64Decode(input) {
+  return decodeURIComponent(
+    atob(input)
+      .split('')
+      .map(
+        character =>
+          '%' + ('00' + character.charCodeAt(0).toString(16)).slice(-2) // eslint-disable-line prefer-template
```
a-roberts, line 68: I think a comment here explaining this goodness might be useful for maintaining later - looks like awesome magic! Kudos for getting things working, definitely a great improvement

AlanGreene: yeah string encoding is always fun 😂 

added some comments

PR #1468, opened 2020-06-05
Remove Sanitizing Secret debug message

CarolynMabbott:
# Changes
Remove the debug message  `logging.Log.Debug("Sanitizing Secret")` as it clogs up my dashboard logs and cant see anything else 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -130,7 +130,6 @@ func SanitizeSecret(obj interface{}, skipDeletedCheck bool) interface{} {
 		Annotations:       annotations,
 	}
 
-	logging.Log.Debug("Sanitizing Secret")
 	data := make(map[string][]byte)
 	if secret.Data["username"] != nil {
 		data["username"] = secret.Data["username"]


eddycharly:
Why not make output level configurable instead ?
Debug logs can be useful, we could not output them by default though.

AlanGreene:
> we could not output them by default though.

I agree, debug logs should be opt-in, we shouldn't be outputting them by default.

eddycharly:
We use `Logger := zap.NewExample()`, from the zap docs:
> NewExample builds a Logger that's designed for use in zap's testable examples. It writes DebugLevel and above logs to standard out ...

CarolynMabbott:
This message is of no use to users though, its left in from when there was a security fix. A user doesnt need to know that we've filtered the resources
Plus it takes up about 2 whole pages of my terminal screen when I first opened an empty secrets page which is extremely annoying when trying to find out why something wasnt working

AlanGreene:
The log level should be configurable, and set to a reasonable level by default (e.g. info). If a developer wants to view the debug logs they would make a config change to set a more verbose log level which includes debug.

eddycharly:
I will open a PR to improve the logger config, i wanted to do it from some time now.
I should have some time today to work on this.

Maybe just switch to `NewProduction` for now if you want json output or `NewDevelopment` if console output is easier to read.

a-roberts:
^^ agreed on the configurable logger, for now I'm cool with getting this in because I'd really rather it not go into 0.7 (I think users with many secrets now, trying to debug something, are going to be in for a bunch of pain - piping output to file and grepping for example).

Thanks @eddycharly and @AlanGreene for the suggestions going forward - could be a good one for the working group calls (hopefully I'll be able to make one soon 🏳️ )

PR #1470, opened 2020-06-05
Add support for log level and log format flags

eddycharly:
# Changes

This PR adds support for zap logger presets.

Presets are easy to use, a more flexible solution would be to use a config map with a json zap config.
It would allow more configuration options but would be more complex.

What do you prefer ?

This is related to https://github.com/tektoncd/dashboard/pull/1468

/cc @AlanGreene @a-roberts @CarolynMabbott 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/cmd/dashboard/main.go
+++ b/cmd/dashboard/main.go
@@ -49,6 +49,8 @@ var (
 	logoutUrl          = flag.String("logout-url", "", "If set, enables logout on the frontend and binds the logout button to this url")
 	csrfSecureCookie   = flag.Bool("csrf-secure-cookie", true, "Enable or disable Secure attribute on the CSRF cookie")
 	tenantNamespace    = flag.String("namespace", "", "If set, limits the scope of resources watched to this namespace only")
+	logLevel           = flag.String("log-level", "info", "Minimum log level output by the logger")
+	logFormat          = flag.String("log-format", "json", "Format for log output (json or console)")
 )
 
 func getCSRFAuthKey() []byte {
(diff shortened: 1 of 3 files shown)


cmd/dashboard/main.go
```diff
@@ -49,6 +49,7 @@ var (
 	logoutUrl          = flag.String("logout-url", "", "If set, enables logout on the frontend and binds the logout button to this url")
 	csrfSecureCookie   = flag.Bool("csrf-secure-cookie", true, "Enable or disable Secure attribute on the CSRF cookie")
 	tenantNamespace    = flag.String("namespace", "", "If set, limits the scope of resources watched to this namespace only")
+	loggerPreset       = flag.String("logger-preset", "production", "Logger preset (production, example or development)")
```
AlanGreene, line 52: can we hide the implementation details here and go with something like: "--loglevel: debug" instead?
if we switch our logger package in future I'd rather not be tied to zap's preset names

eddycharly: Sure, i liked the preset idea as it's simple but ok for log level, maybe combined with a log format (json or console), structured logging is nice when collecting logs but a bit painful when reading in the console.

AlanGreene: yep that sounds reasonable

eddycharly:
Arrf.. the zap config is so... awful :(

cmd/dashboard/main.go
eddycharly: i choose either production or develpment preset according to the format:
- json -> production
- console -> development

eddycharly: done

PR #1443, opened 2020-05-27
Rework kustomize bases/overlays

eddycharly:
# Changes

This PR reorganizes kustomize bases and overlays for easier maintenance.
It adds support for single namespace or full cluster visibility.

The way bases are organized is as follows:
- platform level: `k8s`, `openshift` or `openshift-image-stream`. they contain the resources related to the deployment itself
- rbac level: `namespace` or `cluster`. they contain the resources needed to binding platform resources clusterroles at the namespace or cluster scope

For each platform + rbac combination, the following overlays are available:
- read-only
- read-only-dev
- read-write
- read-write-dev

Overlays are organized in a nested folder structure, each folder level brings kustomizations that are reused by children folder, eg:
`ovelays/k8s/single-namespace/read-only` ->
- `ovelays/k8s` -> takes k8s base and overrides dashboard image 
- `ovelays/k8s/single-namespace` -> adds rbac-namespace base and patches deployment to add --namespace flag
- `ovelays/k8s/single-namespace/readonly` -> patches deployment to add --read-only flag

And so on...

Feedback and ideas welcome, i found it was a logical way to split and organize things, and it allows for reusing as kustomizations gets more and more refined but i would definitely like to hear about other ideas too.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/dashboard/blob/master/CONTRIBUTING.md)
for more details._

PR #1494, opened 2020-06-09
Add installer script for local dev

eddycharly:
# Changes

This PR adds an installer script for local dev only.

It supports the following arguments:
```
Global command syntax:
        installer COMMAND [OPTIONS]

Accepted commands:
        help|h                                  Prints this help
        install|i                               Installs the dashboard
        uninstall|u                             Uninstalls the dashboard
        build|b                                 Builds the manifests and docker image and outputs manfiests in the console

Accepted options:
        [--debug]                               Prints additional messages in the console
        [--csrf-secure-cookie]                  Enable secure CSRF cookie
        [--openshift]                           Will build manifests for openshift
        [--read-only]                           Will build manifests for a readonly deployment
        [--namespace <namespace>]               Will override install namespace
        [--pipelines-namespace <namespace>]     Override the namespace where Tekton Pipelines is installed (defaults to tekton-pipelines)
        [--triggers-namespace <namespace>]      Override the namespace where Tekton Triggers is installed (defaults to tekton-pipelines)
        [--ingress-url <url>]                   Will create an additional ingress with the specified url
        [--ingress-secret <secret>]             Will add ssl support to the ingress
```

The script will use `kustomize` with new overlays dedicated to the installer to build the manifests, then it will invoke `ko` to build the docker image.
Some resources will be added to get a complete set of manifests (role bindings and ingress).
At the end, it will execute `sed` commands to patch the manifest with the arguments provided and will apply the results to the cluster using `kubectl`.

Installing for kubernetes should be as simple as:
- read-write: `./scripts/installer install`
- read-only: `./scripts/installer install --read-only`

Installing for openshift (pipelines and triggers installed with operator pipelines operator):
- read-write: `./scripts/installer install --openshift`
- read-only: `./scripts/installer install --openshift --read-only`

Installing for openshift (pipelines and triggers installed without operator pipelines operator):
- read-write: `./scripts/installer install --openshift --pipelines-namespace tekton-pipelines --triggers-namespace tekton-pipelines`
- read-only: `./scripts/installer install --openshift --read-only --pipelines-namespace tekton-pipelines --triggers-namespace tekton-pipelines`

Next step will be to get the installer script to work with our releases.

/cc @a-roberts @AlanGreene 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/scripts/installer
+++ b/scripts/installer
@@ -1,0 +1,378 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Tekton Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# dashboard flavour
+OPENSHIFT="false"
+READONLY="false"
+
+# configuration default values
+DEBUG="false"
+INSTALL_NAMESPACE="tekton-pipelines"
+PIPELINES_NAMESPACE="tekton-pipelines"
+TRIGGERS_NAMESPACE="tekton-pipelines"
+LOGOUT_URL=""
+CSRF_SECURE_COOKIE="false"
+LOG_LEVEL="info"
+LOG_FORMAT="json"
+
+initOS() {
+  OS=$(echo `uname`|tr '[:upper:]' '[:lower:]')
+
+  case "$OS" in
+    # Minimalist GNU for Windows
+    mingw*) OS='windows';;
+  esac
+
+  debug "Detected OS: $OS"
+}
+
+verifySupported() {
+  if ! type "kubectl" > /dev/null; then
+    echo "kubectl is required"
+    exit 1
+  fi
+
+  if ! type "kustomize" > /dev/null; then
+    echo "kustomize is required"
+    exit 1
+  fi
+
+  if ! type "ko" > /dev/null; then
+    echo "ko is required"
+    exit 1
+  fi
+
+  if ! type "sed" > /dev/null; then
+    echo "sed is required"
+    exit 1
+  fi
+}
+
+debug() {
+  local messsage=$1
+
+  if [ "$DEBUG" == "true" ]; then
+    echo "$message"
+  fi
+}
+
+build() {
+  local overlay="overlays/installer"
+
+  if [ "$OPENSHIFT" == "true" ]; then
+    overlay="$overlay/openshift"
+  else
+    overlay="$overlay/k8s"
+  fi
+
+  if [ "$READONLY" == "true" ]; then
+    overlay="$overlay/read-only"
+  else
+    overlay="$overlay/read-write"
+  fi
+
+  debug "Building overlay $overlay ..."
+
+  kustomize build --load_restrictor none "$overlay" | ko resolve -f - > "$TMP_FILE"
+}
+
+setup() {
+  if [ ! -z "$OVERRIDE_NAMESPACE" ]; then
+    INSTALL_NAMESPACE="$OVERRIDE_NAMESPACE"
+  elif [ "$OPENSHIFT" == "true" ]; then
+    INSTALL_NAMESPACE="openshift-pipelines"
+  fi
+
+  if [ ! -z "$OVERRIDE_PIPELINES_NAMESPACE" ]; then
+    PIPELINES_NAMESPACE="$OVERRIDE_PIPELINES_NAMESPACE"
+  elif [ "$OPENSHIFT" == "true" ]; then
+    PIPELINES_NAMESPACE="openshift-pipelines"
+  fi
+
+  if [ ! -z "$OVERRIDE_TRIGGERS_NAMESPACE" ]; then
+    TRIGGERS_NAMESPACE="$OVERRIDE_TRIGGERS_NAMESPACE"
+  elif [ "$OPENSHIFT" == "true" ]; then
+    TRIGGERS_NAMESPACE="openshift-pipelines"
+  fi
+
+  if [ -z "$LOGOUT_URL" ] && [ "$OPENSHIFT" == "true" ]; then
+    LOGOUT_URL="/oauth/sign_out"
+  fi
+}
+
+replace() {
+  local src=$1
+  local dest=$2
+  
+  debug "REPLACE $src -> $dest"
+
+  if [ "$OS" == "darwin" ]; then
+    sed -i "" "s~$src~$dest~g" $TMP_FILE
+  else
+    sed -i "s~$src~$dest~g" $TMP_FILE
+  fi
+}
+
+patch() {
+  replace "--pipelines-namespace=--pipelines-namespace" "--pipelines-namespace=$PIPELINES_NAMESPACE"
+  replace "--triggers-namespace=--triggers-namespace" "--triggers-namespace=$TRIGGERS_NAMESPACE"
+  replace "--csrf-secure-cookie=--csrf-secure-cookie" "--csrf-secure-cookie=$CSRF_SECURE_COOKIE"
+  replace "--log-level=--log-level" "--log-level=$LOG_LEVEL"
+  replace "--log-format=--log-format" "--log-format=$LOG_FORMAT"
+  replace "--logout-url=--logout-url" "--logout-url=$LOGOUT_URL"
+  replace "--read-only=--read-only" "--read-only=$READONLY"
+  replace "namespace: tekton-dashboard" "namespace: $INSTALL_NAMESPACE"
+}
+
+ingress() {
+if [ ! -z "$INGRESS_URL" ] && [ ! -z "$INGRESS_SECRET" ]; then
+cat <<EOF >> $TMP_FILE
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: tekton-dashboard
+  namespace: $INSTALL_NAMESPACE
+spec:
+  tls:
+  - hosts:
+    - $INGRESS_URL
+    secretName: $INGRESS_SECRET
+  rules:
+  - host: $INGRESS_URL
+    http:
+      paths:
+      - backend:
+          serviceName: tekton-dashboard
+          servicePort: 9097
+EOF
+elif [ ! -z "$INGRESS_URL" ]; then
+cat <<EOF >> $TMP_FILE
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: tekton-dashboard
+  namespace: $INSTALL_NAMESPACE
+spec:
+  rules:
+  - host: $INGRESS_URL
+    http:
+      paths:
+      - backend:
+          serviceName: tekton-dashboard
+          servicePort: 9097
+EOF
+fi
+}
+
+rbac() {
+cat <<EOF >> $TMP_FILE
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: tekton-dashboard-pipelines
+  namespace: $PIPELINES_NAMESPACE
+  labels:
+    app.kubernetes.io/component: dashboard
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-dashboard
+subjects:
+  - kind: ServiceAccount
+    name: tekton-dashboard
+    namespace: $INSTALL_NAMESPACE
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: tekton-dashboard-pipelines
+EOF
+
+cat <<EOF >> $TMP_FILE
+---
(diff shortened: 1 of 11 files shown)


AlanGreene:
This is an interesting idea. I'd be a lot more comfortable if we could use the GItHub APIs instead of screen scraping since there's no guarantee the page structure will remain the same.

https://developer.github.com/v3/repos/releases

Something like:

```
RELEASE_URL=https://api.github.com/repos/tektoncd/dashboard/releases/latest
# or for a specific tag
# RELEASE_URL=https://api.github.com/repos/tektoncd/dashboard/releases/tags/v0.6.1.5

ASSET_URL=`curl $RELEASE_URL | jq -r '.assets[] | select(.name == "openshift-tekton-dashboard-release-readonly.yaml").url'`

# Download the release YAML
curl -L -H "Accept: application/octet-stream" $ASSET_URL -o $TMP_FILE
```

Of course this introduces a dependency on `jq`, so there's a trade-off...

eddycharly:
Added support for ingress.

Regarding the screen scraping, it's only to find the latest version when not specified by the user.
We could also make the version mandatory or use the bucket and use `latest`.

eddycharly:
Added support for overriding namespace

eddycharly:
Added support for overriding pipelines and triggers namespace

AlanGreene:
> it's only to find the latest version

https://github.com/tektoncd/dashboard/releases/latest
returns a 302 redirect to the latest tag, so currently https://github.com/tektoncd/dashboard/releases/tag/v0.6.1.5

Relying on the URL structure + redirect should be safe.

eddycharly:
Didn't know that, thanks !

Still not sure if using the bucket instead of GitHub could allow more freedom, like having a nightly folder in the bucket.

In the end it would be nice to allow installing latest, nightly, a specific version, or even master.

What do you think ?

a-roberts:
> Didn't know that, thanks !
> 
> Still not sure if using the bucket instead of GitHub could allow more freedom, like having a nightly folder in the bucket.
> This could be interesting for testing the release pipeline itself, to avoid bugs like we discovered with @a-roberts this afternoon (#1516).
> 
> In the end it would be nice to allow installing latest, nightly, a specific version, or even master.
> 
> What do you think ?

Yeah, there is more freedom (since there isn't a bunch of arbitrary categories for GitHub releases, e.g. nightly and such), but it's more maintenance/plumbing work for us to do (and that's traditionally been me for the Dashboard). I think doing it right would benefit end users though so it would be useful I think to try and develop this idea further and in parallel, potentially adding more users to the list of those that can write files to the bucket.

eddycharly:
I think we should automate it, pretty easy with prow.
I’ll make a proposal, maybe with discuss it on next wg meeting.

AlanGreene:
Yes the release process should be as automated as possible, I'd probably go as far as suggesting all we should really need to do is create a tag on GitHub.

a-roberts:
> Yes the release process should be as automated as possible, I'd probably go as far as suggesting all we should really need to do is create a tag on GitHub.

Big +1 on this, so normally I'd be looking at Travis or GitHub actions but if there's a Prow hook that'd be perfect. Wondering if @afrittoli or @vdemeester are looking into this kinda thing already...

eddycharly:
IMO we should exercise the release process on every commit on master.

eddycharly:
How about something like this:
- every push on master triggers the release pipeline and get stored in the bucket (version would be the commit hash, maybe master/<commit hash>)
- every night at fixed time execute the release pipeline and store the result in the bucket (maybe nightly/<date>)
- publish a release by tagging the repo, this triggers the release pipeline and store the result in the bucket (maybe release/<tag>)

On the top of that we could maintain master/latest, nightly/latest and release/latest.

This would let us validate the release pipeline often, we could have automated tests to deploy and validate the produced release every day too.

a-roberts:
> How about something like this:
> 
> * every push on master triggers the release pipeline and get stored in the bucket (version would be the commit hash, maybe `master/<commit hash>`)
> * every night at fixed time execute the release pipeline and store the result in the bucket (maybe `nightly/<date>`)
> * publish a release by tagging the repo, this triggers the release pipeline and store the result in the bucket (maybe `release/<tag>`)
> 
> On the top of that we could maintain `master/latest`, `nightly/latest` and `release/latest`.
> 
> This would let us validate the release pipeline often, we could have automated tests to deploy and validate the produced release every day too.

Yep, sounds like the right kinda timeline we'd want to be doing things on, perhaps the next push (after the installation/security enhancements) should be in this space. 

Regarding specific points

```
every push on master triggers the release pipeline and get stored in the bucket (version would be the commit hash, maybe `master/<commit hash>`)
```

Hmm - we'd end up with yaml for every single commit from the point of doing this; that would actually make for really easily testing and the files aren't too big...

```
> * every night at fixed time execute the release pipeline and store the result in the bucket (maybe `nightly/<date>`)
```

instead of actually doing the release we could perhaps do something smarter - just find the last "commit release" (see previous point) within a 24h interval and provide a marker to that somehow.

eddycharly:
> instead of actually doing the release we could perhaps do something smarter - just find the last "commit release" (see previous point) within a 24h interval and provide a marker to that somehow.

Yeah, it could be something like this too.

The idea makes me think about something like `release channels`, we could `dev`, `nightly` and `stable` channels, each one based on the same pipeline but executed at different rates.

Another point is source of truth for the pipeline itself, if i'm correct, right now we have the pipeline definition in the repo but it doesn't sync automatically with the build cluster.
Would it make sense to use the pipeline in the repo rather than a centrally managed one ?

At work we configured prow to use in repo config and we use self contained pipelines, this allows use to evolve the pipelines independently per branch (different branches use different nodejs versions for example).

scripts/installer
```diff
@@ -0,0 +1,287 @@
...
+downloadFile() {
+  MANIFEST="tekton-dashboard-release"
+  if [ "$OPENSHIFT" == "true" ]; then
+    MANIFEST="openshift-$MANIFEST"
+  fi
+  if [ "$READONLY" == "true" ]; then
+    MANIFEST="$MANIFEST-readonly"
+  fi
+  MANIFEST="$MANIFEST.yaml"
+  DOWNLOAD_URL="https://github.com/tektoncd/dashboard/releases/download/v0.7.0/$MANIFEST"
```
AlanGreene, line 78: hardcoded version here, shouldn't it be using $TAG?

scripts/installer
```diff
@@ -0,0 +1,287 @@
...
+set -u
+
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    '--version'|-v)
+       shift
+       if [[ $# -ne 0 ]]; then
+           export DESIRED_VERSION="${1}"
+       else
+           echo -e "Please provide the desired version. e.g. --version v3.0.0 or -v canary"
```
AlanGreene, line 226: should probably use versions we actually have here
`e.g. --version v0.7.0`

scripts/installer
eddycharly: definitely !

scripts/installer
eddycharly: yeah, another arg ? something like `--channel stable/nightly/dev` ?

scripts/installer
eddycharly: fixed

eddycharly:
I think i will remove the calls to GitHub and make the `--version` arg mandatory for now, it seems too early for something more advanced.
I am going to add an installer overlay too, with placeholders to be used with `sed`, this should be a better approach with what we have now.
/hold

eddycharly:
I added an installer overlay and changed the installer to work with the local checkout (this is until we settle on the release strategy and implement it).

The installer overlay populates the manifests with placeholders that can easily be replaced.
To use it:
```bash
kustomize build --load_restrictor none overlays/installer/k8s/read-only | ko resolve -f - | ./scripts/installer --readonly
```

I will do openshift version this  evening.

eddycharly:
I added openshift overlays, and reworked the script so it's only for dev right now.
The plan could be:
- merge this (no big risk as it's only for dev)
- adapt and merge https://github.com/tektoncd/dashboard/pull/1443
- make the script work with a remote release
- adapt the release pipeline and create a pre-release
- if all goes well release the installer with docs etc...

What do you think ?

eddycharly:
I added support for `--no-csrf-secure-cookie` and separated the main command (`help`, `install`, `uninstall`, `build`) from the options.

AlanGreene:
The `csrf-secure-cookie` flag only controls the `Secure` attribute on the cookie, it does not disable setting the cookie entirely.

I haven't had a chance to review all of the recent changes since my last review, I'll try to get to it later today so we can get this in. I'd rather merge something that we can iterate on than try to cover everything in one go.

Can you also update your commit message to include the PR description (or a subset of it)?

eddycharly:
Added installer to the e2e tests

scripts/installer
```diff
@@ -0,0 +1,383 @@
...
+OPENSHIFT="false"
+READONLY="false"
+
+# configuration default values
+DEBUG="false"
+INSTALL_NAMESPACE="tekton-pipelines"
+PIPELINES_NAMESPACE="tekton-pipelines"
+TRIGGERS_NAMESPACE="tekton-pipelines"
+LOGOUT_URL=""
+CSRF_SECURE_COOKIE="true"
```
AlanGreene, line 24: should default to `false` for dev

scripts/installer
```diff
@@ -0,0 +1,383 @@
...
+
+initOS() {
+  OS=$(echo `uname`|tr '[:upper:]' '[:lower:]')
+
+  case "$OS" in
+    # Minimalist GNU for Windows
+    mingw*) OS='windows';;
+  esac
+
+  debug "Detacted OS: $OS"
```
AlanGreene, line 36:
```suggestion
  debug "Detected OS: $OS"
```

scripts/installer
```diff
@@ -0,0 +1,383 @@
...
+  'install'|i)
+    ACTION="install"
+    ;;
+  'uninstall'|u)
+    ACTION="uninstall"
+    ;;
+  'build'|b)
+    ACTION="build"
+    ;;
+  *)
```
AlanGreene, line 304: we should also fallback to this if no command provided, e.g. `./scripts/installer`
currently get an error: `./scripts/installer: line 290: $1: unbound variable`

scripts/installer
```diff
@@ -0,0 +1,383 @@
...
+  esac
+
+  debug "Detacted OS: $OS"
+}
+
+verifySupported() {
+  if ! type "curl" > /dev/null && ! type "wget" > /dev/null; then
+    echo "Either curl or wget is required"
+    exit 1
+  fi
```
AlanGreene, line 43: not used anymore

test/e2e-tests.sh
```diff
@@ -189,7 +205,12 @@ kustomize build overlays/dev-locked-down --load_restrictor=LoadRestrictionsNone
...
 
-test_overlay dev "proxy"
-test_overlay dev-locked-down "kubectl"
+# test overlays
+test_dashboard overlay proxy dev
+test_dashboard overlay kubectl dev-locked-down
+
+# test installer
+test_dashboard installer proxy
+test_dashboard installer kubectl --readonly
```
AlanGreene, line 214:
```suggestion
test_dashboard installer kubectl --read-only
```

test/e2e-tests.sh
```diff
@@ -189,7 +205,12 @@ kustomize build overlays/dev-locked-down --load_restrictor=LoadRestrictionsNone
 kustomize build overlays/dev-openshift --load_restrictor=LoadRestrictionsNone || fail_test "Failed to run kustomize on overlays/dev-openshift"
 kustomize build overlays/dev-openshift-locked-down --load_restrictor=LoadRestrictionsNone || fail_test "Failed to run kustomize on overlays/dev-openshift-locked-down"
 
-test_overlay dev "proxy"
-test_overlay dev-locked-down "kubectl"
+# test overlays
+test_dashboard overlay proxy dev
+test_dashboard overlay kubectl dev-locked-down
+
+# test installer
```
AlanGreene, line 212: Is the plan to remove the `overlay` tests eventually since the installer is doing the same thing under the covers?

test/e2e-tests.sh
eddycharly: the actual option is `--readonly` i can change it to `--read-only` if you want

AlanGreene: ah so it is, yes I think we should change it to be consistent with the dashboard flag

scripts/installer
eddycharly: ok, i will inverse the logic, false by default and an option to enable it ?

eddycharly: fixed

scripts/installer
eddycharly: fixed

scripts/installer
eddycharly: fixed

scripts/installer
eddycharly: removed

test/e2e-tests.sh
eddycharly: changed it

test/e2e-tests.sh
eddycharly: probably later if we promote the installer and deprecate the overlays. as long as we release manifests based on overlays we should keep the tests

test/e2e-tests.sh
AlanGreene: still needs to be updated here

eddycharly: thanks, good catch.
fixed

Issue #1499, opened 2020-06-09
The need to use flags when installing in different namespaces should be documented


Megan-Wright:
As part of: https://github.com/tektoncd/dashboard/pull/1398 we introduced flags for `triggers-namespace` and `pipelines-namespace`, which need to be used if you want to install the dashboard in a separate namespace to pipelines and triggers. However, this is not stated anywhere in the README and so users cannot be expected to know that these need to be used. I would suggest that we document this in the README as part of the dashboard install instructions.

eddycharly:
@Megan-Wright It's documented here https://github.com/tektoncd/dashboard#supported-backend-options
Once the dashboard compiled you can also `dashboard --help`.

I agree it's not very clear but we don't offer a lot of possibilities for the end user to customize the deployment, not sure if it belongs to the user docs or developer docs.

a-roberts:
> @Megan-Wright It's documented here https://github.com/tektoncd/dashboard#supported-backend-options
> Once the dashboard compiled you can also `dashboard --help`.
> 
> I agree it's not very clear but we don't offer a lot of possibilities for the end user to customize the deployment, not sure if it belongs to the user docs or developer docs.

Our overwhelming install base is going to be K8s based I believe though so perhaps @Megan-Wright's thinking more in terms of setting flags through the Kubernetes yaml? Or through deployment patches to apply through kustomize, that kinda thing?

eddycharly:
@a-roberts ah yes, at least overlays should cover the main use case.

Now it would be nice to be able to set these infos at install time (an making sure things are consistent e.g. if flag tells pipelines is in namespace X, we should apply RBAC for namespace X too).

I started writing an install script that could let us go in this direction hopefully see https://github.com/tektoncd/dashboard/pull/1494

PR #1503, opened 2020-06-09
Add pipelines and triggers namespace flags to openshift overlays

eddycharly:
# Changes

This PR adds pipelines and triggers namespace flags to openshift overlays.

Closes https://github.com/tektoncd/dashboard/issues/1499 (hopefully)

# Submitter Checklist

review them:

for more details._


overlays/openshift-patches/flags-patch.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value:
+    --pipelines-namespace=tekton-pipelines
```
a-roberts, line 5: Worth adding a comment here I think, so, and correct me if I'm wrong, we do this because

1) In the OpenShift Pipelines operator case, we'll be in `openshift-pipelines` alongside Triggers and Pipelines, and the lookup is done there by default
2) this really solves the non-operator case (as installing from release.notags.yaml for Pipelines or release.yaml for Triggers) as the deployments will end up in tekton-pipelines

Am I right in thinking that? It's a nuance I think we could capture here. Changes otherwise look great and are very much welcome

eddycharly: Ok for adding the comment but i have to admit i am a bit blind when it comes to openshift.

From what i understand openshift has two install path: operator and non-operator.

Do the manifest we generate in the releases apply to the operator install path ? or is the dashboard also deployed through the operator and the manifests are not used ?

For the non-operator path i suppose everything is deployed using the release manifests, hence the tekton-pipelines/openshift-pipelines mix.

a-roberts:
```
Do the manifest we generate in the releases apply to the operator install path ? or is the dashboard also deployed through the operator and the manifests are not used in this case ?
```

They're for the OpenShift Pipelines operator path, preferring the openshift-pipelines namespace - the Dashboard isn't installed as part of that. May be worth adding something into the main README to advise users on that (and mentioning how the namespace can be adjusted accordingly - which this PR demonstrates through the yaml changes).

```
For the non-operator path i suppose everything is deployed using the release manifests, hence the tekton-pipelines/openshift-pipelines mix.
```

yep, exactly

eddycharly: So when using OpenShift Pipelines operator, the operator installs pipelines (and triggers ?) in openshift-pipelines namespace and the dashboard is deployed manually using the release manifest ?

eddycharly: And OpenShift Pipelines operator is different than tekton operator ?

a-roberts: Yeah that's right, in both regards. Haven't used the Tekton operator on OpenShift but consider it beyond the scope of this work (by documenting how to override the lookup we're covered anyway)

eddycharly: I ask because i'm trying to implement the logic behind all this in https://github.com/tektoncd/dashboard/pull/1494
Thanks

eddycharly: I added a note in docs and comments in the patch file.
Please check it's not complete bullshit ;-)

docs/install.md
```diff
@@ -86,6 +86,10 @@ To install Tekton Dashboard on an OpenShift cluster:
 
 Congratulations! You have successfully installed Tekton Dashboard on your OpenShift cluster.
 
+**Note for users installing Tekton Pipelines and Triggers with the OpenShift Pipelines operator:**
+
+Tekton Pipelines and Triggers will be installed in the `openshift-pipelines` namespace and the Dashboard won't be able to lookup versions there out of the box. You will need to edit the deployment and remove `--pipelines-namespace=tekton-pipelines` and `--triggers-namespace=tekton-pipelines` args to make everything work correctly.
+
```
a-roberts, line 92: In the staged release yaml, these flags won't be present - so you could mention that in two ways.

1) in tekton/README.md, releasers need to know to remove these flags (short of doing it automatically)
2) that means the above is only applicable for developers right? Surely we should provide an overlay that uses those namespaces instead and we qualify that the openshift overlays are for the operator installs?

eddycharly: Not sure if we've cut a release with support for these flags yet, maybe there's no staged release that have them.

Before the flags came in we looked either in all namespaces or had a fallback to look in the `openshift-pipelines` namespace if not found in the install namespace. I changed the lookup logic to allow using `RoleBinding` instead of `ClusterRoleBinding`.

Do you suggest that openshift manifests we release are for the OpenShift Pipelines operator install path ?

I'm going to add support for `triggers-namespace` and `pipelines-namespace` in the installer script.

a-roberts: *Do you suggest that openshift manifests we release are for the OpenShift Pipelines operator install path ?"

Exactly this yeah as that's how it's always been and we can doc that. You're right there's no staged release yet with the flags and hence why I'd like for this issue to effectively result in that, even if I have to do a little manual work for the yaml in the short term

PR #1527, opened 2020-06-17
Write frontend build output directly to KO_DATA_PATH

AlanGreene:
# Changes

Write frontend build output (`npm run build_ko`) directly into
`KO_DATA_PATH` (`/var/run/ko` in the container by default,
`cmd/dashboard/kodata` in the project) instead of into a
subfolder `/var/run/ko/web`

This addresses an issue on some platforms (e.g. katacoda) where
the dashboard couldn't read the static files from that location
due to insufficient folder permissions when running as a
nonroot user.

Also update the `ko` version used as it contains a fix for
permissions on the `KO_DATA_PATH` folder itself.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -12,7 +12,7 @@ You must install these tools:
 
 1. [`go`](https://golang.org/doc/install): The language Tekton Dashboard is built in
 1. [`git`](https://help.github.com/articles/set-up-git/): For source control
-1. [`ko`](https://github.com/google/ko): For development. `ko` version v0.1 or higher is required for `dashboard` to work correctly.
+1. [`ko`](https://github.com/google/ko): For development. `ko` version v0.5.1 or higher is required for `dashboard` to work correctly.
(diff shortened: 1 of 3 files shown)


AlanGreene:
@a-roberts @eddycharly I'm not seeing any issues with this so far. What do you think?

eddycharly:
I dived at the generated image with ko v0.5.1 and it's not different than with v0.5.0.
Maybe the fix here is changing `/var/run/ko/web` by `/var/run/ko` or maybe dive is not showing permissions correctly.

I would like to understand better what is going on but the priority is fixing the issue.

AlanGreene:
0.5.0 and 0.5.1 should be the same (same git sha in the tag) although 0.5.0 was re-tagged at some point, so it's possible that some environments have a cached version from the original release. Using 0.5.1 avoids any potential issues.

0.5.0 should be fine for our CI as we're running in a clean environment with no previous cached version.

You're right that the real fix here is switching the path.

PR #1529, opened 2020-06-17
Adapt test script to easily test with another pipelines version

eddycharly:
# Changes

This PR adapts the test script to ease testing with another pipelines version.

It does so by allowing passing a version to the pipelines installer.
An end user can override the pipelines version deployed by defining the `PIPELINES_VERSION` environment variable:
```base
export PIPELINES_VERSION=v0.12.1
```

The default pipelines version used in our integration tests bumps from `v0.11.0` to `v0.13.2`.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -18,8 +18,6 @@
 
 source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/e2e-tests.sh
 
-pipeline_release=https://github.com/tektoncd/pipeline/releases/download/v0.11.0/release.yaml
-
 function print_diagnostic_info() {
   echo "Diagnostics:"
   resources=("pv" "pvc" "pods")
@@ -42,8 +40,10 @@ function install_kustomize() {
 }
 
 function install_pipeline_crd() {
-  echo ">> Deploying Tekton Pipelines"
-  kubectl apply --filename $pipeline_release  || fail_test "Tekton pipeline installation failed"
+  local version=$1
+
+  echo ">> Deploying Tekton Pipelines ($version)"
+  kubectl apply --filename "https://github.com/tektoncd/pipeline/releases/download/$version/release.yaml" || fail_test "Tekton pipeline installation failed"
 
   # Make sure thateveything is cleaned up in the current namespace.
   for res in pipelineresources tasks pipelines taskruns pipelineruns; do
@@ -55,8 +55,10 @@ function install_pipeline_crd() {
 }
 
 function delete_pipeline_crd() {
-  echo ">> Deleting Tekton Pipelines"
-  kubectl delete --filename $pipeline_release || fail_test "Tekton pipeline deletion failed"
+  local version=$1
+  
+  echo ">> Deleting Tekton Pipelines ($version)"
+  kubectl delete --filename "https://github.com/tektoncd/pipeline/releases/download/$version/release.yaml" || fail_test "Tekton pipeline deletion failed"
 }
 
 # Called by `fail_test` (provided by `e2e-tests.sh`) to dump info on test failure
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -46,25 +46,26 @@ initOS
 install_kustomize
 
 function test_dashboard() {
+  local pipelineVersion=$1
   # overlay or installer
-  installMode=$1
+  local installMode=$2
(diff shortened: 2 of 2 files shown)


eddycharly:
We can reduce the list too, leaving `v0.11.0` should be as long as before.
But it's easier to control the version used by the tests.
By the way, shouldn't we test with pipelines `v0.13.2` now ?

AlanGreene:
Yes we should test with the latest 👍 

Maybe it would be useful to support running against multiple versions in the local-e2e-tests script, so we can still easily run this if we're making changes we expect to impact older versions, but don't have to carry that burden on every PR.

eddycharly:
Changed to test only `v0.13.2`.

a-roberts:
I like the idea of running this against multiple versions but agree doing it on every PR is too much - so let's add some docs to inform users how to do this and when they would want to (e.g. when testing against a new version). And we can mention how to add a new version to the table. I think this, potentially combined with the Postman tests (if we deem that adds sufficient new coverage) is a good step in the right direction

eddycharly:
I agree that running it on every PR is a bit overkill but at least now we can.

eddycharly:
TBH, i don't see what to put in docs, this is just the same as before but just changed the implementation a little bit to allow specifying a pipelines version from the caller when installing tekton pipelines

a-roberts:
> TBH, i don't see what to put in docs, this is just the same as before but just changed the implementation a little bit to allow specifying a pipelines version from the caller when installing tekton pipelines

Sure, so what I'm thinking is: the title of this commit/PR is "Test with multiple pipelines versions", and I think that's really useful.

It would be helpful for us, as we have to deal with "currency" e.g. newer pipeline versions.

You added `for pipelineVersion in v0.13.2` in one of the scripts. 

I'm thinking: repurpose this PR to allow for the testing of multiple pipeline versions, and then as a follow-up, which anyone could do, make it easy for developers to add to that for loop, for their own local testing. This would be useful, say, when Pipelines 0.14 is announced, and we want to run something that tells us at least where to look for potential problems - instead of doing everything manually. We could also run it periodically, and against any given previous version.

I remember working on a similar issue over a year ago and I had text file that was parsed - we could do something similar or just except an arg list.

a-roberts:
Lemme know if you plan on doing anything further @eddycharly, I would also amend your commit message in the same way you've done for the PR title, then happy to lgtm

eddycharly:
Changed the PR title and updated description, more inline with the content of the PR.

Just an idea, we could run several integration tests instances, one per pipelines version `pull-tekton-dashboard-integration-tests-pipelines-0-11-0`, `pull-tekton-dashboard-integration-tests-pipelines-0-12-1`, `pull-tekton-dashboard-integration-tests-pipelines-0-13-2`.

a-roberts:
> Changed the PR title and updated description, more inline with the content of the PR.
> 
> Just an idea, we could run several integration tests instances, one per pipelines version `pull-tekton-dashboard-integration-tests-pipelines-0-11-0`, `pull-tekton-dashboard-integration-tests-pipelines-0-12-1`, `pull-tekton-dashboard-integration-tests-pipelines-0-13-2`.



> Changed the PR title and updated description, more inline with the content of the PR.
> 
> Just an idea, we could run several integration tests instances, one per pipelines version `pull-tekton-dashboard-integration-tests-pipelines-0-11-0`, `pull-tekton-dashboard-integration-tests-pipelines-0-12-1`, `pull-tekton-dashboard-integration-tests-pipelines-0-13-2`.

Yeah, and we could do that at release time especially (not sure how we'd turn that on in an easy way - so we could have links to additional scripts in the Prow config that are only picked up on some flag), or on demand locally through a follow-up PR I discussed above.

Either way, let's get this in and thanks 👍 

/lgtm

eddycharly:
In the end i added a `PIPELINES_VERSION` env variable to allow overriding the version.

Description updated.

eddycharly:
In prow, we could add jobs, one per pipelines version, each job would have a different `PIPELINES_VERSION` env var... do you want that i give it a try ?

a-roberts:
> In prow, we could add jobs, one per pipelines version, each job would have a different `PIPELINES_VERSION` env var... do you want that i give it a try ?

Yep, worth it to see what would happen 😸

eddycharly:
Here we go https://github.com/tektoncd/plumbing/pull/432

Issue #1526, opened 2020-06-16
Support Triggers 0.6


a-roberts:
I noticed @dibyom mentioned in Slack that Triggers 0.6.0 is scheduled for release on June the 22nd.

Here's the milestone https://github.com/tektoncd/triggers/milestone/7. So under this issue it would be useful to try master of the Dashboard with the latest version of Triggers (and potentially every time we see something important is changed), check that our Trigger pages all render OK, and look for new opportunities to either display new fields or provide enhancements.

a-roberts:
@Megan-Wright completed this a while back, while there's no official release yet, folks could pick up master and enjoy the benefits. Closing 😸

PR #1533, opened 2020-06-18
Update eventlistener bindings to use 'ref' instead of 'name'

Megan-Wright:
# Changes
For: https://github.com/tektoncd/dashboard/issues/1526
- Updates TriggerBindings within an EventListener to use `ref` instead of `name` as per https://github.com/tektoncd/triggers/pull/603
- This will mean the dashboard will work with the upcoming triggers 0.6.0 release, as well as triggers 0.5.0, however, it will not work with any triggers releases prior to this.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ It is **strongly recommended** to use the **v0.6.1.4** release or newer for Tekt
 
 | Version | Docs | Pipelines | Triggers |
 | ------- | ---- | --------- | -------- |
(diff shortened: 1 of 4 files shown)


a-roberts:
This does mean we'd only support Triggers 0.6 and later right? Can you update the readme and any tests we have to accomodate this please?

a-roberts:
Ah, noticed in the description we'll work with 0.5 too, so that's good. Yeah let's do a onceover that's all made clear in the readme that 0.4 won't be supported.

Issue #1384, opened 2020-05-15
Follow-up on release pipeline on OpenShift work (lockdown.py)


a-roberts:
# Expected behavior
I shouldn't have to manually lockdown image shas (say, for oauth-proxy) when I run the Dashboard release pipeline on OpenShift

# Actual behavior
I do, because a step was removed due to it relying on using a Docker socket

# Steps to reproduce the problem

1. Run the Dashboard release pipeline on OpenShift (see https://github.com/tektoncd/dashboard/pull/1336/files)
2. Check the yaml for :latest images - they're present

# Environment
- Kubernetes Platform:
OpenShift

- Tekton Pipelines version:
Use 0.11.x or later

- Tekton Triggers version:

- Tekton Dashboard version:

master

# Additional Info
Two immediately obvious options

1) bud pull and use those shas
2) docker in docker?

PR #1537, opened 2020-06-18
Replace overlays with installer

eddycharly:
# Changes

This PR is a replaces the various overlays by an installer script.

The manifests we release are now generated by the installer instead of using separate overlays, this would be easier to maintain.

Additionally, the installer script itself and the overlays it uses are released alongside pre baked manifests.

This allows an end user to use either the pree baked manifests as before, or the installer script if custom options are needed.

# Submitter Checklist

review them:

for more details._


eddycharly:
@a-roberts quick question, why do we have `/tekton/publish.yaml` and `/tekton/openshift/publish.yaml` ?

I need to modify both i guess ?

eddycharly:
I compared the generated manifests with the `v0.7.0` and they are identical.

a-roberts:
> @a-roberts quick question, why do we have `/tekton/publish.yaml` and `/tekton/openshift/publish.yaml` ?
> 
> I need to modify both i guess ?

The OpenShift one doesn't do the lockdown step (where we pin the images instead of using :latest), see https://github.com/tektoncd/dashboard/issues/1384

eddycharly:
Thanks, i modified both `/tekton/publish.yaml` and `/tekton/openshift/publish.yaml`.

I add to create a new `release` command to the installer in order to pass `--preserve-import-paths` to `ko` when doing a release build.

eddycharly:
I need to take care of the openshift with image stream case and it should be good.

I didn't keep the change we did with the services on openshift in this PR, will do that in another PR.

eddycharly:
I added support for `--image-stream` in the installer.

eddycharly:
i removed the tests based on generated manifests, it should be exactly the same as testing the installer
i would like to test when installing in a custom namespace though

eddycharly:
I updated the installer doc, it should be ready for review.

eddycharly:
Damn... postman ! :(

eddycharly:
I added installer overlays in release pipeline.
Would it be reasonable to make a dummy release to test it so that i can adapt the script to work with remote manifests ?

a-roberts:
> I added installer overlays in release pipeline.
> Would it be reasonable to make a dummy release to test it so that i can adapt the script to work with remote manifests ?

Yeah, you should be able to do a release then print the resulting yaml with a few tweaks, and skip the part where it pushes to the gcr bucket (both in terms of files and images), that way we needn't add a bunch of permissions - still waiting on @AlanGreene to have permissions 😢

eddycharly:
Either github or the gcr bucket.
Something where i can fetch the manifests used by the installer instead of having to clone the repo, invoking ko + kustomize, building and pushing the image myself, etc...

As an end user, I would like to be able to write something like this and get the dashboard up and running in my cluster, with my own options:
```bash
curl https://raw.githubusercontent.com/tektoncd/dashboard/master/scripts/installer | bash -s -- --read-only --openshift --version v1.2.3
```

eddycharly:
@a-roberts thanks, it works fine :
```bash
curl -sL https://raw.githubusercontent.com/eddycharly/dashboard/replace-overlays/scripts/installer | bash -s -- install --read-only --version mytestbucket --ingress-url tekton-dashboard.127.0.0.1.nip.io
```

eddycharly:
I tested it successfully yesterday with a dummy release.
Currently when working with a remote release, the script will fetch manifests from the gcs bucket (not using github at all), i will need to update the bash url as it's pointing to a dummy bucket path for now.

Next steps:
- add the script itself in the release
- add tenant namespace support
- change the way we handle service patches on openshift
- update docs

After that we should be good, i will do this in the current PR except the service patches change, i will take care of this in a follow up

eddycharly:
I added support for tenant namespace and adapted e2e/postman tests.
Will try to do the doc today and 🎉 🎉

eddycharly:
I updated docs, stating that it's an experimental feature.

eddycharly:
@AlanGreene @a-roberts this should be ready for review

afrittoli:
> # Changes
> 
> This PR is a work in progress for replacing the overlays by the installer.
> 
> The manifests we release can be generated by the installer instead of using separate overlays, this would be easier to maintain and ultimately we can promote the installer and deprecate completely the generated manifests.

I don't fully agree with the objective of deprecating the manifests completely. 
While it's good to have multiple options to deploy the dashboard, I think there would be be value in having at least one option that only involves applying the manifests. 

Right now the way we install the dashboard (and other Tekton projects) in dogfooding and robocat is as follows:
- a [task](https://github.com/tektoncd/plumbing/blob/master/tekton/resources/release/install_tekton_release.yaml) that installs a release from a release file in an object storage + overlays from git
- [cluster specific overlays](https://github.com/tektoncd/plumbing/tree/master/tekton/cd/dashboard/overlays) that for dashboard setup the service as nodePort and add the ingress for public access
- a [trigger template](https://github.com/tektoncd/plumbing/blob/master/tekton/resources/cd/tekton-template.yaml) for on-demand and nightly deployment
- a [script](https://github.com/tektoncd/plumbing/pull/417) for installation on demand
- a [cronjob](https://github.com/tektoncd/plumbing/tree/master/tekton/cronjobs/robocat-dashboard-deploy-latest-cron) for nightly deployment to the `robocat` cluster

As long as the script allows generating the manifests, it will still be possible to run the scripts with the preferred setup, apply environment specific overlays and install like before - but it would be nice to keep running the installer script as part of the release process, and generate at least a few of the manifests like today.

> # Submitter Checklist
> 
> These are the criteria that every PR should meet, please check them off as you
> review them:
> 
>     * [ ]  Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
> 
>     * [ ]  Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
> 
>     * [ ]  Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
> 
> 
> _See [the contribution guide](https://github.com/tektoncd/dashboard/blob/master/CONTRIBUTING.md) for more details._

AlanGreene:
With this change we still generate the existing 4 manifests as part of the publish pipeline, just using the installer script to generate them instead of the previous kustomize overlays which were becoming unwieldy.

I agree that we wouldn't want to get rid of the manifests entirely, having some predefined configurations that can just be `kubectl apply`-d is very useful for users.

@eddycharly @a-roberts I haven't been following this PR closely in recent days so maybe I've missed some discussion. I see references to running the installer directly from the GitHub repo and it appears to download some additional content itself, and I'm not entirely clear on the intention there.

eddycharly:
> it would be nice to keep running the installer script as part of the release process, and generate at least a few of the manifests like today

@afrittoli that's what we're doing right now, pre baked manifests for common use cases are generated by the installer and are part of the release. Deprecating them is not part of the plan.

AlanGreene:
@eddycharly I think we need to update the description and add a more detailed commit message to explain this change.

In particular this piece needs to be removed:
>  and deprecate completely the generated manifests.

eddycharly:
@AlanGreene the installer can work in two ways:
- with a local repository clone, in this case it will invoke `kustomize` and `ko` to build the docker image/manifests from source
- with a remote release, in this case no need to clone the repo, it will download the manifests from the gcs bucket

eddycharly:
I update the pr description & commit message

eddycharly:
@afrittoli @AlanGreene on the other hand we can invoke `installer build` instead of `installer install` to output the manifests in a file rather than passing the manifests to `kubectl`.

It would require an additional step to build the manifests, then calling `kubectl` in a subsequent step.
IMO it would be useful even in the context of robocat cluster, it would allow you to use new features that will not be part of the pre baked manifests (single namespace visibility for example).

eddycharly:
@Megan-Wright this is not supported on v0.7.0, in fact we have not cut a release compliant with the installer yet.
The release pipeline has been modified to produce installer specific assets and therefore v0.7.0 doesn't have those required assets :(

It will work from the next release, sorry :(

eddycharly:
Right now it can only be used with local checkout (thus not using --version option, `--version` means use a remote release instead of building from sources with `kustomize` and `ko`)

Megan-Wright:
> Right now it can only be used with local checkout (thus not using --version option, `--version` means use a remote release instead of building from sources with `kustomize` and `ko`)

Ahh ok no worries at all, that does make sense - probably worth adding a note somewhere to say that the installer script wont work with releases prior to 0.7.1 then :)

eddycharly:
> probably worth adding a note somewhere to say that the installer script wont work with releases prior to 0.7.1 then :)

thanks, i added the note in the docs

Megan-Wright:
@eddycharly Thanks for adding that :) As far as I can see, everything looking good on openshift as well, except when I try to uninstall I get this error:
```
Error from server (NotFound): error when deleting "/var/folders/4v/9tzhzwks4s1bg2r57rj7k9ww0000gn/T/tekton-dashboard-installer.XXXXXX.unhhqyh9/manifest.yaml": serviceaccounts "tekton-dashboard" not found
Error from server (NotFound): error when deleting "/var/folders/4v/9tzhzwks4s1bg2r57rj7k9ww0000gn/T/tekton-dashboard-installer.XXXXXX.unhhqyh9/manifest.yaml": services "tekton-dashboard" not found
Error from server (NotFound): error when deleting "/var/folders/4v/9tzhzwks4s1bg2r57rj7k9ww0000gn/T/tekton-dashboard-installer.XXXXXX.unhhqyh9/manifest.yaml": deployments.apps "tekton-dashboard" not found
Error from server (NotFound): error when deleting "/var/folders/4v/9tzhzwks4s1bg2r57rj7k9ww0000gn/T/tekton-dashboard-installer.XXXXXX.unhhqyh9/manifest.yaml": rolebindings.rbac.authorization.k8s.io "tekton-dashboard-dashboard" not found
```

eddycharly:
@Megan-Wright when uninstalling, did you pass the same options you passed when installing ?
it's critical to call uninstall with the same options else it won't uninstall the same things that were installed... it's in the docs ;)

Megan-Wright:
> @Megan-Wright when uninstalling, did you pass the same options you passed when installing ?
> it's critical to call uninstall with the same options else it won't uninstall the same things that were installed... it's in the docs ;)

Ah yep, I hadn't realised that, all seems to be working fine then :)
/lgtm

eddycharly:
@Megan-Wright thank you for testing !

docs/install.md
```diff
@@ -36,6 +37,26 @@ See the [docs on the main page](../README.md) to find the Tekton Dashboard versi
...
+While this is still an experimental feature, it should help making the install process customizable and allow you to provide options that suit your needs. We would definitely love to hear your feedback on this new install mode.
+
+You can refer to the dev docs for more infos [on how to use the installer](./dev/installer.md).
+
+**Important note:** this won't work with releases prior to `v0.7.1`, the installer needs specific assets to work and those assets weren't produced in past release thus making the installer impossible to use with versions before `v0.7.1`.
+
+In a nutshell, invoking the `installer` will look something like this:
+
+```bash
+DASHBOARD_VERSION=v0.7.1
```
AlanGreene, line 53: these should be v0.8.0 since we didn't include this for 0.7.1

eddycharly: fixed

Issue #1021, opened 2020-02-06
Full page refresh disrupts the user experience


afrittoli:
# Expected behavior
While I use the dashboard, only parts where content changes are re-rendered

# Actual behavior
At a fixed interval, the dashboard UI is fully re-rendered, which is rather frustrating, for instance while watching logs / progress on a pipeline run.

# Steps to reproduce the problem

1. Open the dashboard, wait 1min or so
2. See also attached gzipped video

# Environment
- Kubernetes Platform:
GKE (1.13.11-gke.14)

- Tekton Pipelines version:
v0.10.1

- Tekton Triggers version:
v0.2.1

- Tekton Dashboard version:
master + https://github.com/tektoncd/dashboard/pull/1017

# Additional Info
[Screen Recording 2020-02-06 at 21.06.25.mov.gz](https://github.com/tektoncd/dashboard/files/4167756/Screen.Recording.2020-02-06.at.21.06.25.mov.gz)

AlanGreene:
hrmm this sounds similar to https://github.com/tektoncd/dashboard/issues/736

Have you noticed this on other pages or just when watching an active PipelineRun?

AlanGreene:
OK seeing this on the dogfooding dashboard, pretty consistently every 34 or 35 seconds the content reloads. It's not a full page refresh, just the content section, but that doesn't make it any less disruptive or annoying for the user.

The websocket connection is being killed after 30 seconds, causing the client to reconnect and re-fetch data in case it's gone stale while disconnected. This is the same as a previous report where a proxy was killing long-lived connections causing this behaviour.

We should look at a more forgiving strategy for ensuring the content is fresh while tolerating frequent disconnects like this. Ideally we would keeping the existing content on screen while re-fetching in the background, so in cases like this it would be completely transparent to the user.

afrittoli:
> hrmm this sounds similar to #736
> 
> Have you noticed this on other pages or just when watching an active PipelineRun?

I've seen refresh happening on other pages too, like the list of eventlisteners - even though there it's less disruptive (possibly because I only have a few defined).

afrittoli:
Another aspect of this issue that makes it worse, is that if I'm looking at some details of an element of the pipeline run, the refresh brings me back to the default page for that pipelinerun, e.g.

<img width="1605" alt="image" src="https://user-images.githubusercontent.com/2205608/74149624-1a1c6180-4c00-11ea-8114-fc5d9c9c3c54.png">

[...]

PR #1578, opened 2020-06-28
Support both polling and streaming logs

eddycharly:
# Changes

This PR adds support for streaming logs instead of polling.

The proxy now supports streaming responses, this allows us to stream logs instead of polling.
Polling is still the default, the user has to enable streaming to use it.

The `Logs` component has been reworked to adapt to both streaming and polling use cases.
A flag was introduced to allow the frontend switching to streaming when enabled.
The installer and overlays were updated.
Doc was updated with the new flag.

I reverted the changes i did with the log lines limit.

/cc @a-roberts @AlanGreene 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "sass-loader": "^8.0.2",
     "storybook-react-router": "^1.0.8",
     "style-loader": "^0.23.1",
+    "web-streams-polyfill": "^2.1.1",
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.11",
     "webpack-dev-server": "^3.11.0",
--- a/packages/components/src/components/Log/Log.js
+++ b/packages/components/src/components/Log/Log.js
@@ -33,11 +33,11 @@ export class LogContainer extends Component {
 
   componentDidMount() {
     this.loadLog();
-    this.initPolling();
   }
 
   componentWillUnmount() {
     clearInterval(this.timer);
+    this.canceled = true;
   }
 
   getLogList = () => {
@@ -92,27 +92,57 @@ export class LogContainer extends Component {
     }
   };
 
-  /* istanbul ignore next */
-  initPolling = () => {
-    const { stepStatus, pollingInterval } = this.props;
-    if (!this.timer && stepStatus && !stepStatus.terminated) {
-      this.timer = setInterval(() => this.loadLog(), pollingInterval);
+  readChunks = ({ done, value }, decoder, text = '') => {
+    if (this.canceled) {
+      this.reader.cancel();
+      return undefined;
     }
-    if (this.timer && stepStatus && stepStatus.terminated) {
-      clearInterval(this.timer);
+    let logs = text;
+    if (value) {
+      logs += decoder.decode(value, { stream: !done });
+      this.setState({
+        loading: false,
+        logs: logs.split('\n')
+      });
+    } else {
+      this.setState(prevState => ({
+        loading: false,
+        logs: prevState.logs
+      }));
     }
+    if (done) {
+      return undefined;
+    }
+    return this.reader
+      .read()
+      .then(result => this.readChunks(result, decoder, logs));
   };
 
   loadLog = async () => {
-    const { fetchLogs, intl } = this.props;
+    const { fetchLogs, intl, stepStatus, pollingInterval } = this.props;
     if (fetchLogs) {
       try {
         const logs = await fetchLogs();
-        this.setState({
-          loading: false,
-          logs: logs ? logs.split('\n') : undefined
-        });
-      } catch {
+        if (logs instanceof ReadableStream) {
+          const decoder = new TextDecoder();
+          this.reader = logs.getReader();
+          await this.reader
+            .read()
+            .then(result => this.readChunks(result, decoder))
+            .catch(error => {
+              throw error;
+            });
+        } else {
+          this.setState({
+            loading: false,
+            logs: logs ? logs.split('\n') : undefined
+          });
+          if (stepStatus && !stepStatus.terminated) {
+            this.timer = setTimeout(() => this.loadLog(), pollingInterval);
+          }
+        }
+      } catch (error) {
+        console.error(error); // eslint-disable-line no-console
         this.setState({
           loading: false,
           logs: [
@@ -122,6 +152,9 @@ export class LogContainer extends Component {
             })
           ]
         });
+        if (stepStatus && !stepStatus.terminated) {
+          this.timer = setTimeout(this.loadLog, pollingInterval);
+        }
       }
     }
   };
--- a/src/api/comms.js
+++ b/src/api/comms.js
@@ -46,7 +46,10 @@ export function getPatchHeaders(headers = {}) {
   };
 }
 
-function parseBody(response) {
+function parseBody(response, stream = false) {
+  if (stream) {
+    return response.body;
+  }
   const contentLength = response.headers.get('content-length');
   if (contentLength === '0') {
     return null;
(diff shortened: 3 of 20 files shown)


eddycharly:
/hold
depends on #1576

eddycharly:
tests should be green ... 🤞

eddycharly:
I didn't think about the 30s kill, good point.

eddycharly:
I'm not sure the 30s thing is an issue, if i'm not mistaken, it's the load balancer that kills the request, not the ingress controller, and i feel like it should not kill this kind of request because it's only sending data (writing in the response stream).

At least that's what i observe on my AWS Application load balancer (idle timeout is 60s but the request keeps working)

AlanGreene:
We've seen this on dogfooding + robocat with the websocket connection being killed every 30s:
https://github.com/tektoncd/dashboard/issues/1021

I referenced the GKE Ingress settings for increasing the timeout in that issue and some threads in #plumbing, this is a common problem.

eddycharly:
I am thinking about adding a flag to make this optional.
Like `--stream-logs` in the backend to drive the frontend behavior, leaving it disabled by default, what do you think ?

eddycharly:
@AlanGreene I changed the approach, reverted the changes related to log lines limit and added a flag to enable streaming so that both polling and streaming are supported (polling being the default).

It should solve the issue with GKE ingress, the third party using the `Log` component should be able to continue using it without troubles, and search should be the same as before.

eddycharly:
/hold cancel

#1576 was merged

eddycharly:
Just rebased but the `needs-rebase` flag doesn't go away

package.json
```diff
@@ -52,7 +52,8 @@
     "reconnecting-websocket": "^4.4.0",
     "redux": "^4.0.5",
     "redux-thunk": "^2.3.0",
-    "regenerator-runtime": "^0.13.5"
+    "regenerator-runtime": "^0.13.5",
+    "web-streams-polyfill": "^2.1.1"
```
AlanGreene, line 56: this should be a dev dependency
`npm remove web-streams-polyfill && npm install -D web-streams-polyfill`

packages/components/src/components/Log/Log.js
```diff
@@ -92,27 +92,50 @@ export class LogContainer extends Component {
...
-      } catch {
+        if (logs instanceof ReadableStream) {
+          const decoder = new TextDecoder();
+          this.reader = logs.getReader();
+          this.reader
+            .read()
+            .then(result => this.readChunks(result, decoder))
+            .catch(error => {
+              throw error;
+            });
```
AlanGreene, line 127: this error isn't being handled
if you want it to be caught by the `catch` block below you'll need to do something like `await this.reader.read()...`

packages/components/src/components/Log/Log.js
```diff
@@ -92,27 +92,50 @@ export class LogContainer extends Component {
...
+          this.setState({
+            loading: false,
+            logs: logs ? logs.split('\n') : undefined
+          });
+          if (stepStatus && !stepStatus.terminated) {
+            this.timer = setTimeout(() => this.loadLog(), pollingInterval);
+          }
+        }
+      } catch (error) {
+        console.log(error); // eslint-disable-line no-console
```
AlanGreene, line 138: `console.error` would be appropriate here if we want to log it

packages/components/src/components/Log/Log.js
```diff
@@ -122,6 +145,9 @@ export class LogContainer extends Component {
             })
           ]
         });
+        if (stepStatus && !stepStatus.terminated) {
+          this.timer = setTimeout(() => this.loadLog(), pollingInterval);
```
AlanGreene, line 149: no need for the additional anonymous function here

```suggestion
          this.timer = setTimeout(this.loadLog, pollingInterval);
```

src/api/index.js
```diff
@@ -305,10 +305,16 @@ export function getCondition({ name, namespace }) {
...
-    queryParams = { container };
+  if (container || follow) {
+    queryParams = {};
+    if (container) {
+      queryParams.container = container;
+    }
+    if (follow) {
+      queryParams.follow = follow;
+    }
   }
```
AlanGreene, line 318: This can be condensed as:

```js
const queryParams = {
  ...(container && { container }),
  ...(follow && { follow })
};
```

Empty object for queryParams should be treated the same as null / undefined so it's safe to do this regardless of the values of `container` or `follow`, and it'll only include the keys with a truthy value.

src/api/comms.js
```diff
@@ -91,14 +94,18 @@ export async function request(uri, options = defaultOptions) {
...
-  }).then(checkStatus);
+  }).then(response => checkStatus(response, stream));
 }
 
-export function get(uri, headers) {
-  return request(uri, {
-    method: 'GET',
-    headers: getHeaders(headers)
-  });
+export function get(uri, headers, stream = false) {
```
AlanGreene, line 100: Could we have an `options` params here so we can name this flag where it's used?

```js
export function get(uri, headers, options = {}) {
```

```diff
const uri = getPodLogURL({ container, name, namespace, follow: 1 });
-return get(uri, { Accept: 'text/plain' }, true);
+return get(uri, { Accept: 'text/plain' }, { stream: true });
```

src/containers/TaskRun/TaskRun.js
```diff
@@ -38,14 +38,16 @@ import {
...
+
 import { LogDownloadButton } from '..';
 import {
   getSelectedNamespace,
   getTaskByType,
   getTaskRun,
   getTaskRunsErrorMessage,
-  isWebSocketConnected
+  isWebSocketConnected,
+  streamLogs
```
AlanGreene, line 50: can we rename this to make it clear it's a boolean? `isLogStreamingEnabled` or something like that

src/api/index.js
```diff
@@ -323,6 +329,11 @@ export function getPodLog({ container, name, namespace }) {
   return get(uri, { Accept: 'text/plain' });
 }
 
+export function followPodLog({ container, name, namespace }) {
+  const uri = getPodLogURL({ container, name, namespace, follow: 1 });
+  return get(uri, { Accept: 'text/plain' }, true);
+}
```
AlanGreene, line 335: could we combine this with `getPodLog` above?

```js
export function getPodLog({ container, name, namespace, stream }) {
  const uri = getPodLogURL({ container, name, namespace, follow: stream });
  return get(uri, { Accept: 'text/plain' }, { stream });
}
```

packages/components/src/components/Log/Log.js
eddycharly: Good catch !

eddycharly: Fixed

src/containers/TaskRun/TaskRun.js
eddycharly: Fixed, thanks

eddycharly:
@AlanGreene Please double check, but i think i fixed all your comments.

eddycharly:
There's something going wrong with the tests :(

eddycharly:
It's green ! 😅

eddycharly:
Gonna try, maybe i made a mistake when rebasing.

eddycharly:
Fixed, we exited too early (without setting the state to an empty string) when done came in the first chunk.

AlanGreene:
Not seeing the hanging skeleton anymore but it's not displaying the 'No log available' message with `stream-logs` enabled.

eddycharly:
Ah, 'No logs available' is the expected behavior ?
I thought it was when we couldn't fetch logs only.

AlanGreene:
'Unable to fetch log' for the error case

eddycharly:
Ok, so we have to set state to loading = false and logs = undefined. Doing it now, thanks.

eddycharly:
Should be fixed now... hopefully 🤞

AlanGreene:
@eddycharly there's a merge conflict here.

I'm also seeing an issue with `stream-logs` enabled:
image

I have a cancelled PipelineRun, and when I view the log tab for any of the cancelled steps it continually retries the `follow=true` request.


As far as I can tell all of the other cases seem to be working as expected.

eddycharly:
@AlanGreene just rebased.

Let's fix the canceling issue in another PR if it's ok. I will give a look to the issue once it's merged, it should not be hard, probably just a matter of checking the status.

PR #1604, opened 2020-07-08
Add release-installer script to ease install of dashboard release

eddycharly:
# Changes

This PR adds a release installer script to ease install of a dashboard release.

Using the installer script directly requires to specify the version twice (once to download the installer script itself from the gcs bucket, and also to pass the version option to the script).

This is error prone and not very pleasant.
The release installer script will take care of downloading the installer from gcs and will inject the version in the command line.
This script will stay in github and doesn't need to be embedded in the release.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/scripts/release-installer
+++ b/scripts/release-installer
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Tekton Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+BASE_RELEASE_URL="https://storage.googleapis.com/tekton-releases/dashboard/previous"
+
+verifySupported() {
+  if ! type "curl" > /dev/null 2>&1 && ! type "wget" > /dev/null 2>&1; then
+    echo "Either curl or wget is required"
+    exit 1
+  fi
+}
+
+install() {
+  VERSION=$1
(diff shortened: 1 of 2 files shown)


eddycharly:
That’s just an example, relying on an env variable set by the user is not a good idea IMO. I prefer something explicit where the user has to pass the version to the script, from an env car or directly on the command line.

AlanGreene:
I guess I'm just not seeing enough benefit of

`<script B> install $DASHBOARD_VERSION --read-only`
over
`<script A> install --version $DASHBOARD_VERSION --read-only`

to justify a separate script and the additional overhead that brings.

How about setting a default version in the script included in the release?

```
DASHBOARD_VERSION=v0.8.0
curl -sL https://storage.googleapis.com/tekton-releases/dashboard/previous/$DASHBOARD_VERSION/installer | \
   bash -s -- install --read-only
```

would be equivalent to

```
DASHBOARD_VERSION=v0.8.0
curl -sL https://storage.googleapis.com/tekton-releases/dashboard/previous/$DASHBOARD_VERSION/installer | \
   bash -s -- install --read-only --version $DASHBOARD_VERSION
```

If the user wanted to install a different version they could specify a different value for `--version`.

eddycharly:
The script should be part of the release, installer script for one version should not be guaranteed to work for another version.

The release-installer script doesn’t have this requirement, it’s just there to bootstrap the version specific install, can stay in github, and should work for any version.

AlanGreene:
> installer script for one version should not be guaranteed to work for another version

OK that makes sense, it would be very limiting if we wanted to make significant changes to the install in the future. I see the benefit now. Thanks 👍

scripts/release-installer
```diff
@@ -0,0 +1,77 @@
...
+    shift
+    VERSION="${1}"
+    shift
+    install $VERSION $@
+    ;;
+  'uninstall'|u)
+    shift
+    VERSION="${1}"
+    shift
+    install $VERSION $@
```
AlanGreene, line 71: shouldn't this be `uninstall`?

eddycharly: Hey sure ! Good catch !

eddycharly: Fixed, thanks !

PR #1605, opened 2020-07-08
Make uninstall not requiring install options

eddycharly:
# Changes

This PR make the uninstall command not requiring install options.

The uninstall command loops through all api resources and deletes the resources having the dashboard labels.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/scripts/installer
+++ b/scripts/installer
@@ -438,7 +438,9 @@ install() {
 # uninstall invokes kubectl delete with the built manifest.
 uninstall() {
   echo "Uninstalling ..."
-  kubectl delete -f $TMP_FILE
+  kubectl delete deployments -l=app.kubernetes.io/instance=default,app.kubernetes.io/part-of=tekton-dashboard --all-namespaces
+  kubectl api-resources --verbs=delete -o name | \
+    xargs -n 1 kubectl delete --ignore-not-found -l=app.kubernetes.io/instance=default,app.kubernetes.io/part-of=tekton-dashboard --all-namespaces
 }
 
 build() {
@@ -513,8 +515,8 @@ case $1 in
     shift
     ;;
   'uninstall'|u)
-    ACTION="uninstall"
-    shift
+    uninstall
+    exit 0
     ;;
   'build'|b)
     ACTION="build"
--- a/docs/dev/installer.md
+++ b/docs/dev/installer.md
@@ -189,18 +189,8 @@ The `installer` script can create an ingress for your Dashboard service, to enab
 
 To uninstall the Dashboard, use the `uninstall` instead of the `install` command.
 
-**IMPORTANT NOTE:** this is VERY important when you uninstall using the `installer` script that you provide the same options that were given to the script at install time. This is needed for the script to know what was created at install time and must be deleted by the `uninstall` command.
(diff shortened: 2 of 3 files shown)


scripts/installer
```diff
@@ -435,7 +435,8 @@ install() {
 # uninstall invokes kubectl delete with the built manifest.
 uninstall() {
   echo "Uninstalling ..."
-  kubectl delete -f $TMP_FILE
+  kubectl api-resources --verbs=list -o name | \
```
AlanGreene, line 438: Should we use `--verbs=list,delete` since for example `componentstatuses` cannot be deleted?

scripts/installer
```diff
@@ -435,7 +435,8 @@ install() {
...
+    xargs -n 1 kubectl delete --ignore-not-found -l=app.kubernetes.io/instance=default,app.kubernetes.io/part-of=tekton-dashboard --all-namespaces
```
AlanGreene, line 439: is the `instance` label necessary here?

scripts/installer
eddycharly: Yeah, maybe delete is enough.

scripts/installer
eddycharly: I think it is. In case of multiple instances, instance should be unique even across namespaces. If it’s not the case it will uninstall all instances.

eddycharly: I plan to add instance override in the installer.

scripts/installer
AlanGreene: oops that's what I meant 😆 👍

eddycharly:
We could delete the deployments first if it’s a concern.

scripts/installer
AlanGreene: It seems like another thing that could be forgotten by a user, and doesn't appear to give any benefit in this scenario over specifying the namespace for example.

If we're relying on it being unique across namespaces we would need to check for this in the installer, but then what about users not using the installer? Maybe you've already considered this and I'm missing something...

eddycharly: One namespace is not enough, resources are created into several namespaces.

True, would be nice to check for unicity at install time.

AlanGreene:
> We could delete the deployments first

I think we should, this way we don't cause unexpected churn on the cluster during uninstall.

eddycharly:
> > We could delete the deployments first
> 
> I think we should, this way we don't cause unexpected churn on the cluster during uninstall.

Yeah, i will rework that when adding instance support to the installer.

AlanGreene:
Spoke to @eddycharly on Slack, we'll update this to delete the deployments first. The change for instance support will follow separately.

/hold

eddycharly:
/hold cancel

@AlanGreene made the discussed changes.
It also fixes an issue in the prepare-kind-cluster (probably due to a bad rebase)

scripts/installer
eddycharly: fixed

scripts/installer
eddycharly: Will take care of multiple instances in another PR.
Deployment is deleted first, then remaining resources.

PR #1616, opened 2020-07-10
Fix modal + notifications for Import Resources and Create PipelineRun/TaskRun

AlanGreene:
# Changes

In Dashboard 0.7.1 and earlier we were not correctly setting the response status
code on proxied responses from the API server. This meant that requests to create
PipelineRuns were returning with a status code 200 on success, instead of the
expected 201 for this API.

With the updated proxy approach we're now correctly setting the status code
on responses, meaning that we triggered a different code path in the front end
for parsing the response. This code path was not expecting a response body with 201,
so the response body remained unread and code relying on it was not properly executed.

As a result, the dialogs were not dismissed and notifications were not displayed
on successful creation of the PipelineRun / TaskRun on these pages.

Update the frontend code to handle 201 responses both with and without response body,
and update unit tests to more accurately test the behaviour of the affected APIs.

Thanks to @eddycharly for helping to debug this one.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/api/comms.js
+++ b/src/api/comms.js
@@ -46,19 +46,31 @@ export function getPatchHeaders(headers = {}) {
   };
 }
 
+function parseBody(response) {
+  const contentLength = response.headers.get('content-length');
+  if (contentLength === '0') {
+    return null;
+  }
+
+  const contentType = response.headers.get('content-type');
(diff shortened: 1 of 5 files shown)


eddycharly:
tested with import resources and works fine
/lgtm

PR #1657, opened 2020-08-05
feat: Create an ansi text parser w/ links support

stevesoaress:
# Changes

- Creates an ANSI parser. 
- Convert all text urls to html links.

While investigating using linkify-it package, I noticed that the currently ansi-to-react package does not correctly parse ANSI codes. For example, `echo -e "\u001b[32mgreen and bold text\u001b[0m\n"` should be green text with a  bold style. https://en.wikipedia.org/wiki/ANSI_escape_code
Reference https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/gitlab/ci/ansi2html.rb#L163

Additionally, the new Ansi code utilizes linkify-it and tlds to convert text to clickable URLs.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5076,9 +5076,11 @@
         "d3-path": "^1.0.8",
         "elkjs": "^0.5.1",
         "js-yaml": "^3.13.0",
+        "linkify-it": "^3.0.2",
         "prop-types": "^15.7.2",
         "react-intl-formatted-duration": "^4.0.0",
-        "react-window": "^1.8.5"
+        "react-window": "^1.8.5",
+        "tlds": "^1.208.0"
       },
       "dependencies": {
         "@formatjs/intl-relativetimeformat": {
@@ -15778,6 +15780,14 @@
       "integrity": "sha1-HADHQ7QzzQpOgHWPe2SldEDZ/wA=",
       "dev": true
     },
+    "linkify-it": {
+      "version": "3.0.2",
+      "resolved": "https://registry.npmjs.org/linkify-it/-/linkify-it-3.0.2.tgz",
+      "integrity": "sha512-gDBO4aHNZS6coiZCKVhSNh43F9ioIL4JwRjLZPkoLIY4yZFwg264Y5lu2x6rb1Js42Gh6Yqm2f6L2AJcnkzinQ==",
+      "requires": {
+        "uc.micro": "^1.0.1"
+      }
+    },
     "load-json-file": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/load-json-file/-/load-json-file-2.0.0.tgz",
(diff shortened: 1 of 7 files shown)


packages/components/src/components/Ansi/Ansi.js
```diff
@@ -0,0 +1,268 @@
...
+      }
+    }
+    return null;
+  };
+
+  const getTermColor = (colorIndex, type) => {
+    if (colorIndex > 7) {
+      return null;
+    }
+    if (type === 'l') {
```
AlanGreene, line 69: what's `l` here? From context I'm guessing `light` or similar (referencing the set of alternative 'bright' colours?) It would be good to use a descriptive name to make this clear.

packages/components/src/components/Ansi/Ansi.js
```diff
@@ -0,0 +1,268 @@
...
+
+    90: () => setFGColor(0, 'l'),
+    91: () => setFGColor(1, 'l'),
+    92: () => setFGColor(2, 'l'),
+    93: () => setFGColor(3, 'l'),
+    94: () => setFGColor(4, 'l'),
+    95: () => setFGColor(5, 'l'),
+    96: () => setFGColor(6, 'l'),
+    97: () => setFGColor(7, 'l'),
+    99: () => setFGColor(9, 'l'),
```
AlanGreene, line 136: Is this needed? Similar for 109 below.

I can't find any reference to these codes being used. Looks like 39 and 49 would be used instead in these cases to reset the colours.

packages/components/src/components/Ansi/Ansi.js
```diff
@@ -0,0 +1,268 @@
...
+    }
+    const elements = [];
+    let offset = 0;
+    matches.forEach(match => {
+      if (match.index > offset) {
+        const string = str.substring(offset, match.index);
+        elements.push(<span style={styleObj}>{string}</span>);
+      }
+      elements.push(
+        <a href={match.url} style={styleObj}>
```
AlanGreene, line 180: open in new window and don't give child window access to the opener object

```suggestion
        <a href={match.url} style={styleObj} target="_blank" rel="noopener noreferrer">
```

AlanGreene:
Just noticed the update to package-lock.json is missing too for the new deps, can you run `npm run bootstrap` from the root of the project and add the package-lock.json updates in this PR too? Thanks

packages/components/src/components/Ansi/Ansi.js
```diff
@@ -0,0 +1,268 @@
...
+        const [[key, value]] = Object.entries(textStyles[style]);
+        styles[key] = value;
+      }
+    });
+  };
+
+  const parse = (ansi, index) => {
+    // eslint-disable-next-line no-control-regex
+    const ansiRegex = new RegExp('^\u001b([@-_])(.*?)([@-~])');
+    const characterRegex = new RegExp('.', 'm');
```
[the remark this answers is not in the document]
stevesoaress: I've defined these as a global variable.

packages/components/src/components/Ansi/Ansi.js
stevesoaress: Renamed to use 'light'

packages/components/src/components/Ansi/Ansi.js
stevesoaress: You're right. I couldn't find the codes in the linux man pageshttps://man7.org/linux/man-pages/man4/console_codes.4.html. I will remove them

packages/components/src/components/Ansi/defaults.js
```diff
@@ -0,0 +1,68 @@
...
+const textStyles = {
+  bold: {
+    fontWeight: 'bold'
+  },
+  italic: {
+    fontStyle: 'italic'
+  },
+  conceal: {
+    visibility: 'hidden'
+  },
```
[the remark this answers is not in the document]
stevesoaress: I agree. We can revisit this after the additional performance changes that are needed.

PR #1665, opened 2020-08-07
Update LogFormat conceal styles so text can still be selected / copied

AlanGreene:
# Changes

In environments where the 'conceal' style is supported it is typically
implemented by setting the foreground colour equal to the background
colour. This means that the content can still be viewed by selecting it,
and can also be copy/pasted if needed.

Switch from `visibility: hidden` (which means the content cannot be copied)
to `color: transparent` to match typical implementations.

# Submitter Checklist

review them:

for more details._


AlanGreene:
@stevesoaress I thought there'd be more involved here but looks like setting `color: transparent` does the trick and avoids having to maintain a stack of colours or anything like that.

PR #1673, opened 2020-08-11
Use version latest in installer docs

eddycharly:
# Changes

This PR uses `latest` version in docs when invoking the installer script.
It should be easier to maintain as we won't need to update the docs when releasing a new version.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -50,12 +50,18 @@ You can refer to the dev docs for more infos [on how to use the installer](./dev
 In a nutshell, invoking the `installer` will look something like this:
 
 ```bash
-DASHBOARD_VERSION=v0.8.0
 curl -sL https://raw.githubusercontent.com/tektoncd/dashboard/master/scripts/release-installer | \
-   bash -s -- install $DASHBOARD_VERSION --read-only
+   bash -s -- install latest --read-only
 ```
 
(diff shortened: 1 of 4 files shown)


docs/install.md
```diff
@@ -50,9 +50,8 @@ You can refer to the dev docs for more infos [on how to use the installer](./dev
 In a nutshell, invoking the `installer` will look something like this:
 
 ```bash
-DASHBOARD_VERSION=v0.8.0
 curl -sL https://raw.githubusercontent.com/tektoncd/dashboard/master/scripts/release-installer | \
-   bash -s -- install $DASHBOARD_VERSION --read-only
+   bash -s -- install latest --read-only
```
AlanGreene, line 54: using `latest` in the walkthroughs is fine and reduces the overhead of keeping the docs up to date, but here in the install docs we should at least give an example of installing a specific version so users know the format that's expected. Having `latest` as the main example is fine, but we should add a note on installing a specific version:

e.g. ```**Note**: to install a specific version: ... `bash -s -- install v0.8.0 --read-only` ...```

eddycharly: good point, i added a note. thanks

eddycharly:
easy cla down ?

PR #1679, opened 2020-08-12
Don't patch service on openshift but create a secure one for the Route

eddycharly:
# Changes

This PR changes the way services are managed on openshift.

Previously, we patched the original service to remove port 9097 and add port 443, this service was then used by the `Route`, then we created a second service with port 9097 to use inside the cluster.

This PR doesn't change the original service anymore, therefore it is to be used inside the cluster, instead it creates the secure service and use it in the `Route`.

This is simpler as it doesn't require modifying an existing resource, it leaves the original service as it is and creates two new resources that are openshift specific.

This should be ok with webhooks-extension as the service is looked up by labels (see https://github.com/tektoncd/experimental/pull/551).

/cc @a-roberts 
/cc @AlanGreene 

# Submitter Checklist

review them:

for more details._


eddycharly:
This should be ok with webhooks-extension as the service is looked up by labels (see https://github.com/tektoncd/experimental/pull/551).

eddycharly:
ping @a-roberts

AlanGreene (approved):
/lgtm

Issue #1697, opened 2020-08-26
Dashboard can be developed on Windows


yamaken1343:
# Is your feature request related to a problem ?

Currently, developing Tekton Dashboard on Windows requires some file modifications.
There is a Linux dependency on the npm script.
So when running on Windows, it gets stoped during the bootstrap process at `npm ci`.

# Describe the solution you'd like

I want to be able to develop the front end even on Windows.

Dependencies can be resolved by dynamically changing the script executed on Linux and Windows using [run-script-os](https://www.npmjs.com/package/run-script-os).

# Describe alternatives you've considered

Since the changes are few (2 lines), it is easy for the user to modify them.
In this case, they only need to describe the changing lines in the development documentation.

# Additional context

Git for Windows rewrites line feeds to `CRLF` by default on checkout.
Therefore, prettier throws up an error when you run `npm start`. (occurs `error Delete ␍ prettier/prettier`)
To get around this, you need to fuzzy up the prettier rules.
Specifically, add `endOfLine: "auto"` to `.prettierrc`.
Should I include this code in PR?

If you need this feature, I'll create PR.
Thanks.

AlanGreene:
Thanks for the issue @yamaken1343, can you share more information about the steps / commands that are failing on Windows?

Examples of the commands you're running and the error output would be very helpful.

yamaken1343:
## Errors in `npm ci`

Errors in `npm ci`, the following errors occurred.
The user only runs `npm ci`.
Then `lerna ERR! npm run prepare stderr: 'NODE_ENV' is not recognized as an internal or external command, operable program or batch file.` occurred.

We think the cause is as the following.
When `npm-ci` is running, `npm run bootstrap` is executed after the module installation is finished.  
In the `npm run bootstrap`, the `build` which is described in the `dashboard\packages\utilsutilspackage.json` is run.  
In the `build`, `rimraf dist && NODE_ENV=production babel src --out-dir dist --root-mode upward --ignore "src/**/*.test.js,src/**/*.stories.js" --no -copy-ignored` is executed.  
I think there is an error caused by an unrecognized format on Windows. (`NODE_ENV=production babel` command)

NOTE: A similar problem is contained in the file `dashboard\packages\componentspackage.json`.

Here is the full text.

```cmd
npm run bootstrap

> @tektoncd/dashboard@0.1.0 bootstrap C:\parentDirectory\dashboard
> lerna bootstrap

lerna notice cli v3.22.1
[...]

PR #1707, opened 2020-08-28
Ensure dashboard frontend can be built on Windows without change

AlanGreene:
# Changes

Fixes https://github.com/tektoncd/dashboard/issues/1697

Ensure npm scripts are using cross-platform safe commands.
Use `cross-env` to set environment variables since the `VAR=value`
syntax isn't valid in the standard Windows command prompt.

Also ensure line endings are set to enforce LF, preventing
git converting to CRLF on Windows.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/package-lock.json
+++ b/package-lock.json
@@ -10877,6 +10877,58 @@
         }
       }
     },
+    "cross-env": {
+      "version": "7.0.2",
+      "resolved": "https://registry.npmjs.org/cross-env/-/cross-env-7.0.2.tgz",
+      "integrity": "sha512-KZP/bMEOJEDCkDQAyRhu3RL2ZO/SUVrxQVI0G3YEQ+OLbRA3c6zgixe8Mq8a/z7+HKlNEjo8oiLUs8iRijY2Rw==",
+      "dev": true,
+      "requires": {
+        "cross-spawn": "^7.0.1"
+      },
+      "dependencies": {
+        "cross-spawn": {
+          "version": "7.0.3",
+          "resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-7.0.3.tgz",
+          "integrity": "sha512-iRDPJKUPVEND7dHPO8rkbOnPpyDygcDFtWjpeWNCgy8WP2rXcxXL8TskReQl6OrB2G7+UJrags1q15Fudc7G6w==",
+          "dev": true,
+          "requires": {
+            "path-key": "^3.1.0",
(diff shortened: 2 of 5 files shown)


eddycharly:
Unfortunately i don't have a windows box to test it...

docs/dev/README.md
```diff
@@ -32,6 +32,10 @@ In order to run the Tekton Dashboard, please make sure the requirements in [the
 You will also need the following tools in order to build the Dashboard locally and deploy it:
 1. [`go`](https://golang.org/doc/install): The language the Tekton Dashboard backend is built in
 1. [`git`](https://help.github.com/articles/set-up-git/): For source control
+
+   **Note:** On Windows make sure to set `core.autocrlf` to `true` in your git config to ensure line endings are handled correctly
```
yamaken1343, line 36:
```suggestion
   **Note:** On Windows make sure to set `core.autocrlf` to `input` or `false` in your git config to ensure line endings are handled correctly
```

docs/dev/README.md
```diff
@@ -32,6 +32,10 @@ In order to run the Tekton Dashboard, please make sure the requirements in [the
...
+
+   e.g. `git config --global core.autocrlf true`
```
yamaken1343, line 38:
```suggestion
   e.g. `git config --global core.autocrlf input`
```

yamaken1343 (changes requested):
You cannot build include CRLF in your code. Therefore, you need to make sure that it is not converted at checkout.
You also need to make sure that the CRLF is not included in the new lines.
If you don't like it, you need to fuzzy up the prettier rules.
Specifically, add `endOfLine: "auto"` to `.prettierrc`.

docs/dev/README.md
AlanGreene: we don't want to use `false` as that would allow users commit changes with carriage returns to the repo

`input` does seem to make more sense here, but we still have the problem with the prettier config
I'd prefer to keep our existing prettier config, let me see if there's something we can do here

AlanGreene: Added `.gitattributes` to enforce use of LF, git will auto-convert any CRLF to LF on commit (same as `core.autocrlf=input`), and will NOT convert to CRLF on checkout.

We already have `.editorconfig` which is supported by most of the popular editors, e.g. VS Code, Sublime Text, Atom, IntelliJ IDEA, Eclipse, WebStorm, NetBeans, etc. which sets line endings to LF, and our prettier config will catch any CRLF that manage to sneak through despite all of this.

yamaken1343 (approved):
I have confirmed that these work on Windows.
Using `.gitattributes` is a great idea!

steveodonovan (approved):
/approve
/lgtm

PR #1711, opened 2020-09-04
Remove OpenShift specific installer script manifests

AlanGreene:
~~**Note:** this is an exploration / WIP, I've added it to the agenda for the next WG call 
to discuss whether we're OK with this kind of change in principal, then we can work
out the specifics of what we do / don't want to change / remove.~~
[Update 2020/09/21] No objections after being raised on the WG call for the last few weeks

---

# Changes

The main difference between the default Dashboard install and the
OpenShift variant is the inclusion of the oauth-proxy and related
config. Since the oauth-proxy has a number of shortcomings (e.g.
lack of proper session timeout / invalidation) and better
alternatives exist, remove the oauth-proxy config from the
Dashboard install. We have walkthroughs in the Dashboard docs that
describe the steps required to set up oauth2-proxy or Envoy as
examples. Removing oauth-proxy allows us to simplify the release
pipeline and remove the lockdown.py script.

As well as removing the oauth-proxy config, remove the `/v1/endpoint`
and `/v1/ingress` APIs since they're not used by the Dashboard and
were for legacy integration. This allows us to remove our dependency
on the OpenShift client and reduce the amount of code that needs to be
maintained and tested.

Also remove the OpenShift release pipeline since it's incomplete,
requires manual changes, and we should be moving our release process
to the dogfooding cluster similar to other projects anyway. Since this
change removes the lockdown.py script which didn't work on OpenShift
there shouldn't be an issues with running the normal release pipeline
on OpenShift if we wish to do so.

# Submitter Checklist

review them:

for more details._


eddycharly:
The /v1/endpoint and /v1/ingress APIs are used by the webhooks extension. Are we going to stop supporting the extension ?

I don’t know openshift well but there was the Route resource too.

Even if it’s not perfect, how about openshift users ? Shouldn’t we provide at least some guidance on how to install on openshift ?

AlanGreene:
> The /v1/endpoint and /v1/ingress APIs are used by the webhooks extension. Are we going to stop supporting the extension ?

The webhooks-extension is already deprecated, the last Dashboard + webhooks-extension release was 0.7.0.

> Shouldn’t we provide at least some guidance on how to install on openshift ?

This PR is only removing the installer script manifests for OpenShift, not the release manifests. The installer script manifests aren't needed if we remove the oauth-proxy config since that was the only difference.

As I mentioned in the PR description I've added this to the WG agenda to discuss it more and decide exactly which pieces we do / do not want to keep.

bobcatfish:
Apologies, just investigating an issue with /retest:

/retest

steveodonovan (approved):
Talked though it with Al on slack

/approve
/lgtm

Issue #1759, opened 2020-10-13
Logs fail to display when they contain multiple consecutive carriage return characters


AlanGreene:
# Describe the bug

If log content contains multiple consecutive carriage return ('\r') characters
this causes the LogFormat component to fail with the following error:

```
TypeError: Cannot read property '0' of null
```

# Expected behaviour

Logs should display correctly.

# Steps to reproduce the bug

# Environment details

- Kubernetes Platform:

- Kubernetes or OpenShift version:

- Install mode (if on OpenShift):

- Cloud-provider/provisioner:

- Versions:
   - Tekton Dashboard:
   
   - Tekton Pipelines:
   
   - Tekton Triggers:
   
- Install namespaces:
   - Tekton Dashboard:
   
   - Tekton Pipelines:
   
   - Tekton Triggers:
   

# Additional Info

AlanGreene:
Fix merged, will be included in next published `@tektoncd/dashboard-components` package (likely tomorrow) and also included in next dashboard release.

/close

AlanGreene:
Fix is available in Dashboard v0.10.2

PR #1760, opened 2020-10-13
Fix for LogFormat component handling consecutive carriage returns

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1759

If log content contains multiple consecutive carriage returns ('\r')
this causes issues with the ANSI parser resulting in an error
displayed to the user and no logs output.

Specifically, the use of '.' and the 'm' (multiline) flag in the
regex for the ANSI parser fails in this case. Instead, it's
recommended to use '[^]' instead of '.' when operating in multiline
mode. This ensures we can match these and other escape sequences
correctly across lines. ES2018 introduces the 's' (dotAll) flag for
similar purpose, but this is not yet supported in all browsers.
# Submitter Checklist

review them:

for more details._


steveodonovan (approved):
/approve
/lgtm

Issue #1800, opened 2020-10-28
The menu entry for Extensions for cluster-scoped CRD should always be visible


Fabian-K:
# Describe the bug
We have registered a cluster-scoped CRD to the dashboard. The extension is displayed as long as the Namespace selector in the menu on the left is set to "All Namespaces" or to the namespace of the Extension CR instance.

# Expected behaviour
The menu entry for cluster-scoped resources should always be visible.

# Additional Info
This is a follow up for https://github.com/tektoncd/dashboard/issues/1731

AlanGreene:
This should be a simple fix. We already have code in place to detect when the Dashboard is installed in single namespace mode so it should be safe to make the non-namespaced API call if the `tenantNamespace` is not set instead of falling back to the globally selected namespace from the UI.

I'll test this out now and open a PR. We'll include the fix in the release that's due in the next couple of days.

/assign

AlanGreene:
Fix is merged, it'll be included in the next nightly build (~3am UTC) and also the next release.

/close

AlanGreene:
Fix is available in Dashboard v0.10.2

PR #1802, opened 2020-10-28
Fix loading of CRD extensions when a specific namespace is selected

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1800

The side nav should always display extensions, regardless of the
globally selected namespace. The resulting UI produced by the extension
will be subject to the filter, but its presence/absence in the nav
should not depend on this.

Since we already check for the single namespace mode, we can safely
change the fallback to query all namespaces (i.e. use the non-namespaced
API call) instead of using the globally selected namespace.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/App/App.js
+++ b/src/containers/App/App.js
@@ -30,7 +30,12 @@ import {
   LogoutButton,
   PageErrorBoundary
 } from '@tektoncd/dashboard-components';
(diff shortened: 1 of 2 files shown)


steveodonovan (approved):
/approve
/lgtm

Issue #1727, opened 2020-09-25
Implement i18n for tekton dashboard


kobaji:
Hi.
We believe that it's meaningful to internationalize the Dashboard, which will make Tekton easier to use for non-native people.
For this reason, we [proposed internationalization](https://app.slack.com/client/TJ45YV83X/CHTHGRQBC/thread/CHTHGRQBC-1597822056.045700) in Slack before.
Thank you for your response at that time.
We create this issue according to your reply.

There are two things we want to confirm.

1. Does the Dashboard project define the translation process? If not, We'd like to discuss it with you.

   For reference, we researched several famous projects.

   - Kubernetes(<https://kubernetes.io/docs/contribute/localization/>)
   - Jenkins(<https://wiki.jenkins.io/display/JENKINS/Internationalization>)
   - harbor(<https://goharbor.io/docs/1.10/build-customize-contribute/developer-guide-i18n/>)

   The Kubernetes example is helpful. For each language, they seem to have one team responsible for reviewing.
   The other two have written translations, but there is no process within the community. They may not make it public.
   However, it would be grateful if you extended the current <https://github.com/tektoncd/dashboard/tree/master/docs/dev#i18n> as they do.

2. We think we can add a translation by modifying the code as follows, is this correct?

   - create `src/nls/messages_ja.json`
   - add `ja` to `locales.build` in `config_frontend/config.json`
   - run `npm run nls`
   - generated `src/nls/messages_ja.json` filled with only key
   - add translate to `src/nls/messages_ja.json`
   - load `src/nls/messages_ja.json` on `src/containers/App/App.js`

We will help you with the translation into Japanese.
Thanks.

AlanGreene:
1. we don't have a formal translation process documented for the community but can certainly improve the existing development docs to include more information about the expected format etc. Thank you for the links to those other processes, I'll review them and see what makes sense to include for the dashboard.

2. we already have https://github.com/tektoncd/dashboard/blob/master/src/nls/messages_ja.json partially populated with translations, mostly for the PipelineRuns and PipelineRun components. Additional translations can be added to that file by replacing the empty string values.

   The translations are not exposed in the UI yet, so as you stated `src/containers/App/App.js` would need to be modified to load the correct file based on the browser locale. We should allow developers to view even partial translations while working on the dashboard / translations, and once a given locale is complete (or at an acceptable level) we would add it to `locales.supported` in `config.json` so it would be loaded for a production build too.

kobaji:
[...]

PR #1808, opened 2020-10-29
Add Japanese translation

Martsings:
Hello,

We have translated the dashboard into Japanese, regarding to #1727. 
Please let me know if there is problem with modified `app.js`.
For those lines with no value (value in `messages_*.json` is with an empty string), the default language (English) of the message will be displayed. 

# Changes

We added Japanese translations to `messages_ja.json`, with `App.js` loading `messages_ja.json` directly. 

# Submitter Checklist

review them:

for more details._


src/nls/messages_ja.json
```diff
@@ -1,304 +1,304 @@
...
-    "dashboard.about.version": "",
-    "dashboard.accessTokenField.accessToken": "",
-    "dashboard.accessTokenField.accessTokenRequired": "",
-    "dashboard.accessTokenField.hidePasswordLabel": "",
-    "dashboard.accessTokenField.showPasswordLabel": "",
-    "dashboard.actions.createButton": "",
-    "dashboard.actions.deleteButton": "",
-    "dashboard.app.loadingConfigError": "",
+    "carbon.table.batch.item.selected": "{totalSelected, plural, =0 {0 項目} one {1 項目}}が選択されました",
+    "carbon.table.batch.items.selected": "{totalSelected, plural, other {# 項目}}が選択されました",
```
AlanGreene, line 13: can the plural syntax be removed from these if it's not needed?

```suggestion
    "carbon.table.batch.item.selected": "項目が選択されました",
    "carbon.table.batch.items.selected": "項目が選択されました",
```

src/containers/App/App.js
```diff
@@ -81,10 +81,13 @@ import {
   isReadOnly,
   isWebSocketConnected
 } from '../../reducers';
-import messages from '../../nls/messages_en.json';
+import enMessages from '../../nls/messages_en.json';
+import jaMessages from '../../nls/messages_ja.json';
```
AlanGreene, line 85: We should only load the one file required for the user's language, as we would end up loading several thousand lines of JSON otherwise when we add more languages.

As I mentioned in the issue we should add any language with completed translations to `config.locales.supported` and use that to determine which files can be loaded, or fallback to English as a default if the user's language is not yet supported.
https://github.com/tektoncd/dashboard/blob/aeaee0be5a1bbadcc59d98151d968f2fe3698a01/config_frontend/config.json#L6

The existing code already handles using English as a fallback for any untranslated strings in the selected language so there's no need to include `message_en.json` when loading another language.

AlanGreene (changes requested):
Thank you for your PR. This will be a fantastic update for any Japanese speaking users.

Can you please squash your commits so there's just one, and add a descriptive commit message? https://github.com/tektoncd/community/blob/master/standards.md#commit-messages

If you would like, you can remove the change in `App.js` for now so we can merge the change to `message_ja.json` and follow up with the `App.js` change in a separate PR. Or we can wait and do them both in this PR. Let me know which you prefer.

src/containers/App/App.js
kobaji: Thank you @AlanGreene and all, for kind advice.
We are wondering how we can load a specific language file such as messages_ja.json dynamically.
`require` requires a literal and will be specified at compile time, `fs` cannot load modules.
Should you have any suggestion, please let us know.

AlanGreene: For dynamically loading additional content in the client you can use a [dynamic import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Dynamic_Imports).

We're already using it in the [Extension component](https://github.com/tektoncd/dashboard/blob/7c74fcc2161a6117d615cdc7bf7bb0c356718140/src/containers/Extension/Extension.js#L29) for example.

Martsings: Thank you for your advice.

Since it will take some time to deal with the modification of `App.js`, we would like to  remove the change in `App.js` and just let you merge the changes to message_ja.json. 

And because it's also needed to squash the commits, I'm going to close this PR and create a new one afterwards. Or if there is better way to do with that, please let me know.

AlanGreene: Opening a new PR will be fine 👍 Thank you

Issue #1775, opened 2020-10-15
Update TaskTree design


AlanGreene:
Extracted from https://github.com/tektoncd/dashboard/issues/1445#issuecomment-672984164

Update the design of the TaskTree component to:
- reduce use of heavy block colours
- make it easier to see status of multiple TaskRuns / steps
- etc.

Here's a rough mockup of the proposed design:
![image](https://user-images.githubusercontent.com/2829095/96123551-e14d2e80-0eea-11eb-8f53-ffad4a3c0a14.png)

There are still some details to be worked out but I think this would be a nice improvement.

AlanGreene:
`StatusIcon` component has been updated and provides 2 styles: default and 'inverse'.
The inverse style currently uses a white fill and is intended for use on the block colour backgrounds.
With the new design this can switch to the outline style instead. May need a minor update to remove the [exclusions](https://github.com/tektoncd/dashboard/blob/36c0d2f1ecfb938c50373585cff719175018dd63/packages/components/src/components/Task/Task.js#L88).

AlanGreene:
I have the initial updates mostly complete but will hold off on the PR until after the next Dashboard release to avoid holding up getting some important fixes out (e.g. rerunning cancelled runs https://github.com/tektoncd/dashboard/issues/1788)

AlanGreene:
Some focus states still need to be updated, e.g. selected step, but otherwise it's nearly there.
![image](https://user-images.githubusercontent.com/2829095/97310451-282c1400-185b-11eb-83b3-13897881eb6b.png)

I've made some changes to StatusIcon size and positioning.

This first pass maintains the existing behaviour of only having a single TaskRun expanded at a time. A future update will add the chevron to allow multiple to be expanded at the same.

AlanGreene:
PR #1806 with the initial design update is merged, will check it out on robocat after the next nightly deploy.

The next step is to support expanding multiple TaskRuns at the same time so the user can see summary of step status without having to switch between them.

PR #1822, opened 2020-11-10
Display duration on TaskRun and step details

AlanGreene:
# Changes

Resolves https://github.com/tektoncd/dashboard/issues/1775

Add duration below the name header on TaskRun and step details
views so the user doesn't have to compute the value themselves
from the start / end timestamps in the 'Status' tab.

image

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/DetailsHeader/DetailsHeader.js
+++ b/packages/components/src/components/DetailsHeader/DetailsHeader.js
@@ -16,11 +16,44 @@ import { ChevronRight24 as DefaultIcon } from '@carbon/icons-react';
 import { injectIntl } from 'react-intl';
 import { getStatus } from '@tektoncd/dashboard-utils';
 
-import { StatusIcon } from '..';
+import { FormattedDuration, StatusIcon } from '..';
 
 import './DetailsHeader.scss';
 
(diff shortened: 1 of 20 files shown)


packages/components/src/components/DetailsHeader/DetailsHeader.js
```diff
@@ -16,11 +16,45 @@ import { ChevronRight24 as DefaultIcon } from '@carbon/icons-react';
 import { injectIntl } from 'react-intl';
 import { getStatus } from '@tektoncd/dashboard-utils';
 
-import { StatusIcon } from '..';
+import { FormattedDuration, StatusIcon } from '..';
 
 import './DetailsHeader.scss';
 
 class DetailsHeader extends Component {
+  getDuration() {
```
steveodonovan, line 24: Followed up in DM but might be worth making this configurable ?

AlanGreene: good point, I also noticed some odd behaviour with the duration being revised downwards when a step completes so I'll update this to only display on completion

steveodonovan (approved):
/approve
/lgtm

PR #1852, opened 2020-11-30
Add additional handling to prevent possible UI error selecting default Task / step

AlanGreene:
# Changes

Add additional handling to the PipelineRun container to ensure we
don't incorrectly select the wrong TaskRun in the case where there's
no `pipelineTaskName` selected and a TaskRun does not have a
condition check label. This would result in the comparison
`undefined == undefined`, potentially leading to the wrong TaskRun
being expanded by default, and not reflected in the URL.

This is unlikely to occur, but is easily avoided by checking that
a condition check label is present.

# Submitter Checklist

review them:

for more details._


steveodonovan (approved):
/approve
/lgtm

Issue #1304, opened 2020-04-20
Overhaul the namespace dropdown global filter UX


AlanGreene:
Our current UX for the global namespace filter has lead to a number of issues that cause user frustration, bugs, or nasty workarounds in the code.

Tracking them here under a single issue so we can attempt to address these issues in a better and more consistent manner that's easier to maintain without the need for messy workarounds.

Current:
- ~~Can't see created secret depending on globally selected vs. target namespace: https://github.com/tektoncd/dashboard/issues/1271~~
- ~~Changing namespace to All Namespaces while on create secret cancels create secret: https://github.com/tektoncd/dashboard/issues/1300~~
- ~~Changing namespaces on namespaced Secrets url changes from access token to password: https://github.com/tektoncd/dashboard/issues/1299~~
- ~~Create PipelineResource not using globally selected namespace: https://github.com/tektoncd/dashboard/issues/1031~~
- ~~Create Secret not using globally selected namespace: https://github.com/tektoncd/dashboard/issues/975~~
- We have a lot of duplicate code in SideNav that handles switching to 'All Namespaces' and redirects to the correct list view depending on the details page the user was on. We could remove this if the namespaces dropdown wasn't visible on the details pages.

There may be more, I'll add them here as I find them, but these are the most obvious remaining issues.

Also including some historical issues for broader context:
- Selected namespace reset when cancelling Create Secret: https://github.com/tektoncd/dashboard/issues/1186 and https://github.com/tektoncd/dashboard/issues/1241
- Switching namespace on details page redirects to PipelineRuns (default page)
  - Task: https://github.com/tektoncd/dashboard/issues/1179
  - Secret: https://github.com/tektoncd/dashboard/issues/1160
  - PipelineRun (error instead of redirect): https://github.com/tektoncd/dashboard/issues/1070
  - ServiceAccount: https://github.com/tektoncd/dashboard/issues/1026
  - TaskRun: https://github.com/tektoncd/dashboard/issues/759
  - PipelineResource: https://github.com/tektoncd/dashboard/issues/758
- Selected nav item issues:
  - Triggers: https://github.com/tektoncd/dashboard/issues/868
- Interaction between global namespace filter + URL / modals:
  - Create Secret: https://github.com/tektoncd/dashboard/issues/549

Design:
_*TBD:* include details of proposed design updates to address these_
(placeholder for now)
[...]

Issue #805, opened 2019-12-04
Dashboard UI Implementation for Specs


carrenelloyd:
# Expected Behavior
As a user I need a UI that uses Carbon components and is consistent with other business applications that IBM produces. 

# Actual Behavior
The current implementation should use consistent components across all pages of the application. 

# Additional Info

### Heading and Breadcrumbs
A consistent heading and breadcrumbs can be used throughout all pages in Tekton. The use of existing code can be found here:
[The IDE Page Header Component with breadcrumbs and heading ](https://pages.github.ibm.com/cloud-integration-design/ide-components/?path=/story/production-ready-ide-navigation-components-ide-page-header-component--with-breadcrumb-and-heading)

### Button styles and placement
Buttons and actions can be anchored along the Heading or Page Title/breadcrumb area to the right side of the grid. In this case, shown below a Secondary (black), Ghost button (blue outline) and a Primary button have been used.
![image](https://user-images.githubusercontent.com/53481064/70175572-f9d2a280-16a4-11ea-85e9-e16ccc842d16.png)

The buttons are centered along the Page Title 
Layout-04 Token is used to separate the buttons above and below the Filter
![image](https://user-images.githubusercontent.com/53481064/70175905-8f6e3200-16a5-11ea-884c-96c557afe959.png)

Layout-04 Token is used below the button btwn the Tabbed area
![image](https://user-images.githubusercontent.com/53481064/70176212-191dff80-16a6-11ea-92f8-4e7ac94d42d8.png)

AlanGreene:
We no longer use breadcrumbs in the Dashboard headers, and much of the common layout is handled by the application shell and the ResourceDetails component.

Button styles and placement for buttons (and other controls) in the header will need to be taken into account for https://github.com/tektoncd/dashboard/issues/1304

AlanGreene:
Addressing this as part of https://github.com/tektoncd/dashboard/issues/1304

PR #1853, opened 2020-12-01
Move global namespaces dropdown to list page header

AlanGreene:
# Changes

Resolves https://github.com/tektoncd/dashboard/issues/1304
Resolves https://github.com/tektoncd/dashboard/issues/805

Move the global namespaces dropdown from the side nav to the
header of the list pages.

To avoid duplication and to help ensure consistency, introduce
a new ListPageLayout container which is responsible for providing
the common layout and functionality for the list pages.

Moving the namespaces dropdown allows us to resolve a number of
long-standing issues:
- remove code duplication required to redirect to list pages
  when changing namespace from details page as this scenario
  is no longer possible
- a namespaced view will now always contain the namespace
  in the URL, providing consistency as it no longer behaves
  differently when navigating from a namespaced / non-namespaced
  view to a namespaced list. This also means that URLs are
  reliable and can be shared to produce the same view the user
  is currently viewing

It also allows us to remove a large amount of custom styling
in SideNav.scss which was required to style the dropdown to
match the dark theme used in the nav.

One final change that was required to avoid potentially
expensive API calls to build the nav for extensions is the
addition of an optional `namespaced` field (defaults to `true`)
on the kubernetes-resource extensions. This is used to ensure
we generate the correct URL in the side nav (i.e. namespaced /
non-namespaced) regardless of the current URL / globally
selected namespace.

# Submitter Checklist

review them:

for more details._


src/containers/ListPageLayout/ListPageLayout.js
```diff
@@ -0,0 +1,104 @@
...
+      );
+      return;
+    }
+
+    let prefix = '';
+    if (!match?.params?.namespace) {
+      prefix = paths.byNamespace();
+    }
+
+    const newURL = generatePath(prefix + match.path, {
```
steveodonovan, line 59: `${prefix}${match.path}` instead?

src/containers/ListPageLayout/ListPageLayout.js
```diff
@@ -0,0 +1,104 @@
...
+      selectNamespace(selectedNamespace);
+      setPath(
+        generatePath(match.path.replace(paths.byNamespace(), ''), {
+          ...match.params
+        })
+      );
+      return;
+    }
+
+    let prefix = '';
```
steveodonovan, line 54: `const prefix = !match?.params?.namespace ? paths.byNamespace() : "";`

src/containers/ListPageLayout/ListPageLayout.js
```diff
@@ -0,0 +1,104 @@
...
+  history,
+  location,
+  match,
+  namespace,
+  selectNamespace,
+  tenantNamespace,
+  title
+}) => {
+  function setPath(path) {
+    history.push(path + location.search);
```
steveodonovan, line 38: up to you on the whole using template strings or not

src/containers/PipelineResources/PipelineResources.js
```diff
@@ -220,7 +220,7 @@ export /* istanbul ignore next */ class PipelineResources extends Component {
         ];
 
     return (
-      <>
+      <ListPageLayout title="PipelineResources" {...this.props}>
```
steveodonovan, line 223: Should the titles be externalised strings ? Probably not given theyre types but just to double check

steveodonovan (commented):
A few nits, looks good overall

src/containers/PipelineResources/PipelineResources.js
AlanGreene: not for these since we're using the resource `kind` names from Kubernetes and don't translate them

src/containers/ListPageLayout/ListPageLayout.js
AlanGreene: yeah we use them in most places so I'll update these for consistency 👍

steveodonovan (approved):
/approve
/lgtm

PR #1864, opened 2020-12-09
Update react-testing-library to latest major version

AlanGreene:
# Changes

Update from react-testing-library@6 to @testing-library/react@11 (package is under the
@testing-library scope since v8.x)

Changes:
- find/replace `react-testing-library` for `@testing-library/react`
- remove use of `cleanup-after-each` as this happens automatically since v9
- a number of breaking changes in the DOM testing library (now @testing-librar/dom v7)
  since @testing-library/react v9
  - `getBy*` / `queryBy*`: now throw an error if more than one element returned. Use
    `getAllBy` etc. instead.
  - `getByValue` is replaced by `getByDisplayValue`
  - `getByLabelText` changes behaviour, added quick fix for now but we should revisit
    this in a separate behaviour. It now returns more complete results, so in a number
    of cases it's returning multiple matches instead of the single expected node.
    I've updated the affected tests making the minimal reasonable change to get the
    tests passing

Future changes:
  - review use of `getByLabelText` and improve to follow the testing library
    recommendations such as specifying a selector
  - `wait`, `waitForElement`, etc. are now deprecated. Use of `waitForElement` should
    be possible to trivially replace with `waitFor`, but alternatives are recommended.
    I've left these as they are for now so there will be some warnings in the logs for
    a short period. I'll follow up in a separate PR to address these

Note: default timeout for async utilities changes from 4.5s to 1s in the DOM testing
library v7, so watch out for any resulting flakiness.

# Submitter Checklist

review them:

for more details._


steveodonovan (approved):
/approve
/lgtm

Issue #775, opened 2019-11-25
Dashboard UI Implementation for Pipelines and Tasks


kimholmes:
# Expected Behavior
The portion of the Dashboard UI for Secrets should use design patterns and components for Carbon 10 and reflect the user experience created through collaboration between Development, Design and User Research

# Actual Behavior
The actual behavior is found in the current implementation of the Dashboard UI

# Design Specs for Pipelines:
At the following site, please find these three pages with specs for Pipelines:
https://kimholmes.github.io/Exported-Specs/PipelinesAndTasks/Pipelines/
Use the left nav to access the following pages:
<img width="241" alt="image" src="https://user-images.githubusercontent.com/21316329/69561221-05261f80-0f7b-11ea-96cb-41780c428eaf.png">

**Pipeline Main List View:**
![Piplelines](https://user-images.githubusercontent.com/21316329/69561493-78c82c80-0f7b-11ea-9dda-768986dbbe3f.png)

**Pipeline Row Selected:**
![Piplelines_Selected](https://user-images.githubusercontent.com/21316329/69561523-87aedf00-0f7b-11ea-9e87-e167e2c7cdcc.png)

**Pipeline Filtered List:**
If the user clicks on a Pipeline in the above page, the result is a filtered list of PipelineRuns
![PipelineRuns_NonSelected](https://user-images.githubusercontent.com/21316329/69561628-ba58d780-0f7b-11ea-932f-956f3ede1e14.png)

# Design Specs for Tasks:
At the following site, please find these three pages with specs for Tasks:
https://kimholmes.github.io/Exported-Specs/PipelinesAndTasks/Tasks/
Use the left nav to access the following pages:
<img width="241" alt="image" src="https://user-images.githubusercontent.com/21316329/69561978-4834c280-0f7c-11ea-9eb0-5a8f8e668875.png">

**Tasks Main List View:**
![Tasks](https://user-images.githubusercontent.com/21316329/69562136-8df18b00-0f7c-11ea-97f9-72ff29dab34b.png)

**Tasks Row Selected:**
![TasksSelected](https://user-images.githubusercontent.com/21316329/69562211-afeb0d80-0f7c-11ea-9720-d255af124283.png)

**Tasks Filtered List:**
If the user clicks on a Task in the above page, the result is a filtered list of TaskRuns
![TaskRuns](https://user-images.githubusercontent.com/21316329/69562267-c85b2800-0f7c-11ea-80f9-6697b7fb585c.png)

[...]

PR #1877, opened 2020-12-16
Add ability to delete Pipelines, Tasks, and ClusterTasks

AlanGreene:
# Changes

Resolves https://github.com/tektoncd/dashboard/issues/775

Add the ability to delete Pipelines, Tasks, and ClusterTasks, both
individually and in bulk via the Carbon Table batch actions toolbar.

There's one slight difference in the user experience for these resource
types compared to the existing delete functionality for PipelineRuns
and TaskRuns. That is, the delete action for a single resource is
exposed as a button inline on the table row rather than being nested
in an overflow menu. This is because we only have at most 2 actions
on a row, one being the link to the details page for the resource.
Carbon guidelines recommend keeping the actions inline when there are
fewer than 3 actions.

# Submitter Checklist

review them:

for more details._


src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -345,20 +348,22 @@ export /* istanbul ignore next */ class PipelineRuns extends Component {
             lowContrast
           />
         )}
-        {this.state.submitError && (
+        {this.state.deleteError && (
```
steveodonovan, line 351: Not for this PR but just spotted that were missing error handling for cancelling a run :see_no_evil:

https://github.com/tektoncd/dashboard/blob/266f8030f92a3d321ba648b1d97512f8c2280dfe/src/containers/PipelineRuns/PipelineRuns.js#L115

steveodonovan (approved):
Discussed on follow up item in slack

/approve
/lgtm

src/containers/PipelineRuns/PipelineRuns.js
AlanGreene: so we are... created an issue to track that: https://github.com/tektoncd/dashboard/issues/1884

PR #1885, opened 2021-01-06
Reduce code duplication for delete modals and disambiguate resources in different namespaces

AlanGreene:
# Changes

Follow up to https://github.com/tektoncd/dashboard/pull/1877

Add a new DeleteModal component to handle the common layout / styling
for all of our views that use it. Update all containers that provide
delete functionality to use the new component.

The new modal uses a Table instead of UnorderedList and will display
the namespace of the selected resources (when in 'All Namepsaces' mode
and viewing namespaced resources)

As part of this change I've also updated the Secrets container to be
more consistent with the use of complete resources (referenced by id)
we have for other resource types and to avoid string manipulation to
pass name/namespace around. This also removes the need for the
SecretsDeleteModal as it's now using the common pattern and can use
the generic DeleteModal.

image

# Submitter Checklist

review them:

for more details._


AlanGreene:
EasyCLA not updating again...

/close

src/actions/secrets.js
```diff
@@ -57,13 +57,15 @@ export function deleteSecret(secrets, cancelMethod) {
     // from error handling and additional notifications on success/error
     const namespacesToSecretsMap = new Map();
     secrets.forEach(secret => {
-      let foundSecretInfo = namespacesToSecretsMap.get(secret.namespace);
+      let foundSecretInfo = namespacesToSecretsMap.get(
```
steveodonovan, line 60: For this and the lines below just pull out name and namespace at the top ?

src/containers/Tasks/Tasks.js
```diff
@@ -287,45 +286,13 @@ export /* istanbul ignore next */ class Tasks extends Component {
...
-            </p>
-            <UnorderedList nested>
-              {toBeDeleted.map(task => {
-                const { name, namespace } = task.metadata;
-                return <ListItem key={`${name}:${namespace}`}>{name}</ListItem>;
-              })}
-            </UnorderedList>
-          </Modal>
+          <DeleteModal
+            kind="Tasks"
```
steveodonovan, line 290: Dont we have constants somewhere for the kind ? doesn't have to change, just if they're there might be better.

steveodonovan (commented):
Some nits, overall great!

src/actions/secrets.js
AlanGreene: good call, done 👍

src/containers/Tasks/Tasks.js
AlanGreene: There are no constants for these since they're usually only used in a single container, 'Tasks' for example is only used here in the Tasks container and in 1 test.

steveodonovan (approved):
/approve
/lgtm

PR #1887, opened 2021-01-07
Fix styling of inline table actions in read only mode

AlanGreene:
# Changes

When in read only mode there's currently only a single inline action
on table rows. Align the inline actions to the right so that they're
always at the end of the row instead of being positioned at the left
of their containing column.

Before:
image

After:
image

There should be no visual difference when multiple inline actions are present:
image

# Submitter Checklist

review them:

for more details._


steveodonovan (approved):
/approve
/lgtm

PR #1899, opened 2021-01-14
Simplify CSRF protection to work with client bundle approach

AlanGreene:
# Changes

Related to 1388

Remove the masked cookie approach for a simpler solution relying
only on the presence of a custom request header. These can only
be set by JavaScript meaning same origin restrictions apply. As
long as the app is not vulnerable to XSS attacks this approach
is safe, and if it's vulnerable to XSS pretty much all bets are
off anyway...

This change eliminates the need to make an additional request
before any mutation to retrieve the CSRF token, meaning that
the front end code will also be compatible with the kubectl
proxy client approach without needing any additional config
or detection for CSRF. This uniform approach to making requests
regardless of the back end (Dashboard golang back end or
Kubernetes API server via kubectl proxy) keeps maintenance
relatively simple.

It also has the added benefit of eliminating the need for special
consideration of HTTPS connections. Users no longer have to
modify the CSRF secure cookie flag in their deployment to match
the scheme used to access the application. This has caused
confusion in the past, with a number of users leaving the setting
disabled in a production environment for convenience.

# Submitter Checklist

review them:

for more details._


steveodonovan (approved):
/approve
/lgtm

dghubble:
Maybe deserves a v0.14.0 changelog note, since a flag was removed :+1: Got here bc of crashloop.

AlanGreene:
@dghubble thanks for the suggestion. I've updated the release notes to mention removal of the `--csrf-secure-cookie` argument.

PR #641, opened 2019-10-15
Base64 encode input data for secret creation

a-roberts:
# Changes

Fixes https://github.com/tektoncd/dashboard/issues/639

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
^^ I'd love to

- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/dashboard/blob/master/CONTRIBUTING.md)
for more details._

PR #1905, opened 2021-01-15
Fix buffer undefined error in create new secret page

ziheng:
# Changes

When in Create new Secret page, users are not able to create Secret type Password. In dev console it throws error `Uncaught ReferenceError: Buffer is not defined
    at <anonymous>:1:1`

Before:
Screenshot 2021-01-15 at 15 06 42

After:
no error

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9171,8 +9171,7 @@
     "base64-js": {
       "version": "1.5.1",
       "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.5.1.tgz",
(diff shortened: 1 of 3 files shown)


AlanGreene (changes requested):
This is strange, the use of `Buffer` here is not new, it was added in https://github.com/tektoncd/dashboard/pull/641
My current best guess is that a change somewhere in the webpack / babel / corejs stack is no longer bundling a polyfill it used to...

Can you `npm install buffer` and include the package.json/package-lock.json updates too so we're not relying on a transitive dependency? Should be good to go then 👍

ziheng:
Thanks Al, update the commit with package.json

AlanGreene (approved):
/lgtm
/meow

PR #1967, opened 2021-03-04
Remove unnecessary tags for LogFormat component

stevesoaress:
# Changes

Goal is to reduce the size of the DOM tree to hopefully improve the rendering time of large logs.  

If text isn't formatted, we return the string instead of wrapping the string in a <span> tag. 

**Example:** 
```
<div>
  <span class="tkn--ansi--color-fg--cyan">INFO</span>
  <span>[0015] Taking snapshot of files...</span>
</div>
```
_becomes_
```
<div>
  <span class="tkn--ansi--color-fg--cyan">INFO</span>
  [0015] Taking snapshot of files...
</div>
```

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/LogFormat/LogFormat.js
+++ b/packages/components/src/components/LogFormat/LogFormat.js
@@ -36,6 +36,18 @@ const getXtermColor = commandStack => {
   return null;
 };
 
+const createFormattedString = (str, styleObj, className) => {
+  const hasStyles = styleObj.color || styleObj.backgroundColor || className;
+  if (hasStyles) {
+    return (
+      <span style={styleObj} className={className}>
+        {str}
+      </span>
+    );
+  }
+  return str;
+};
+
 const linkify = (str, styleObj, classNameString) => {
   const className = classNameString || undefined;
   if (!str) {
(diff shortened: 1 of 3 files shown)


AlanGreene:
Thanks for the PR @stevesoaress. Do you have any data illustrating the impact of the change? It would be good to understand at what point the number of DOM nodes becomes an issue and the improvement this change brings, even just a rough indication.

/ok-to-test

packages/components/src/components/LogFormat/LogFormat.js
```diff
@@ -36,29 +36,33 @@ const getXtermColor = commandStack => {
   return null;
 };
 
+const createFormatedString = (str, styleObj, className) => {
```
AlanGreene, line 39:
```suggestion
const createFormattedString = (str, styleObj, className) => {
```

packages/components/src/components/LogFormat/LogFormat.js
```diff
@@ -36,29 +36,33 @@ const getXtermColor = commandStack => {
...
     return null;
   }
   const matches = linkifyIt.match(str);
   if (!matches) {
-    return (
-      <span style={styleObj} className={className}>
-        {str}
-      </span>
-    );
+    return createFormatedString(str, styleObj, className);
```
AlanGreene, line 58:
```suggestion
    return createFormattedString(str, styleObj, className);
```

AlanGreene (commented):
Small typo

stevesoaress:
@AlanGreene I unfortunately only have anecdotal evidence. Perhaps the best way to check this out is by the "Performance "section the Chrome dev tools which will take time to gather

AlanGreene:
Testing this out now, assuming no performance regression it's good to go 👍

AlanGreene:
Tested using the Storybook in both dev and production modes on the `LogFormat` 'performance' story using Chrome dev tools performance pane for both this PR (rebased on `main`) and the `main` branch.

Average results over 20 iterations each, standard deviation across all results ~0.1

dev
38% reduction in scripting time
20% reduction in rendering time
~26% reduction overall (~6.5s reduced to ~4.8s)

prod
negligible change in scripting time
20% reduction in rendering time
~15% reduction overall (~5.4s reduced to ~4.6s)

Not seeing any change in visual output so we're good to go 👍 Thanks @stevesoaress 

/lgtm
/approve

PR #1968, opened 2021-03-04
Fix display of EventListeners with null triggers field

AlanGreene:
# Changes

Handle missing triggers field in EventListener spec

Display NamespaceSelector if specified

Before:
image

After:
local dev frontend connected to dogfooding cluster, so it's the same resource
image

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/EventListener/EventListener.js
+++ b/src/containers/EventListener/EventListener.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2019-2020 The Tekton Authors
+Copyright 2019-2021 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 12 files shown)


src/containers/EventListener/EventListener.js
```diff
@@ -93,14 +97,25 @@ export /* istanbul ignore next */ class EventListenerContainer extends Component
             {serviceType}
           </p>
         )}
+        {namespaceSelector?.matchNames?.length ? (
```
steveodonovan, line 100: Can we use `&&` instead of the conditional ?

AlanGreene: when dealing with a numeric condition it's safer to use the ternary otherwise if you're not careful you can end up with a rogue '0' appearing on the page

steveodonovan (approved):
/approve
/lgtm

Issue #1842, opened 2020-11-23
Investigate removing most use of redux in the front end


AlanGreene:
Redux is not really necessary for most of our state since most slices of state are only used on a small number of pages in very specific contexts. Very little state is common to the bulk of the application, we could probably reduce this to just the install properties, extensions, namespaces, and maybe one or two other slices.

Our current use leads to performance issues when a page has been left open for a long time connected to a busy cluster, e.g. TaskRun resources are only required on 3 pages (TaskRuns list, TaskRun details, and PipelineRun details), but all pages bear the cost of processing and storing at least some TaskRun data, i.e. when we receive a websocket message containing a TaskRun update even when not currently viewing anything TaskRun-related. We store this data indefinitely in memory, until we either receive a delete update, or the page is refreshed / closed, leading to excessive memory usage and gradual slow down as the volume of data / updates increases.

Investigate alternative approaches to managing what is essentially server state mirrored in the frontend. One interesting option to consider would be [React Query](https://react-query.tanstack.com/docs/overview. There may also be other packages that are more suited to our use cases, we should be able to run some small scale experiments to compare them.

AlanGreene:
This could also simplify our efforts if we want to consume historical data from https://github.com/tektoncd/results in future. Something to consider / discuss...

AlanGreene:
This should be one of the next big changes we make as it enables a number of interesting possibilities and integrations.
/lifecycle frozen

AlanGreene:
This is currently blocked by #2000 which is in progress.

AlanGreene:
Tracking progress of remaining work…

Resources to migrate to react-query:
- [ ] install properties 🚧 #2091
- [ ] ClusterInterceptors 🚧 
- [ ] ClusterTasks
- [ ] ClusterTriggerBindings 🚧 
- [ ] Conditions 🚧 
- [ ] EventListeners 🚧 
- [ ] Extensions
- [ ] Namespaces
- [ ] PipelineResources 🚧 
- [ ] PipelineRuns
- [ ] Pipelines
- [ ] ServiceAccounts 🚧 
- [ ] TaskRuns 🚧 
- [ ] Tasks
- [ ] Triggers 🚧 
- [ ] TriggerBindings 🚧 
- [ ] TriggerTemplates 🚧 

Actions, reducers, selectors, and associated tests can be deleted for each of these.
Once all of the above are done we can also remove a number of utils that are no longer needed.

- [ ] actionCreators
[...]

PR #1984, opened 2021-03-15
Refactor containers to use hooks

AlanGreene:
# Changes

Related to https://github.com/tektoncd/dashboard/issues/1842
### **Note:** ignore whitespace for an easier review

Refactor the App, About, Conditions, and Condition containers to
be functional components using React hooks instead of being class
components.

This change is a prerequisite for adopting react-query and other
libraries which provide a hooks-based API.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/About/About.js
+++ b/src/containers/About/About.js
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React, { Component } from 'react';
+import React, { useEffect } from 'react';
 import { connect } from 'react-redux';
 import { injectIntl } from 'react-intl';
 import { InlineNotification } from 'carbon-components-react';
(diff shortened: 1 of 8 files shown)


src/containers/App/App.js
```diff
@@ -197,281 +161,299 @@ export /* istanbul ignore next */ class App extends Component {
...
+    } catch (error) {
+      console.error(error); // eslint-disable-line no-console
+      setLoadingConfigError(error);
+      setIsLoadingConfig(false);
+      setShowLoadingState(false);
+    }
   }
+
+  useEffect(() => {
+    if (webSocketConnected !== false) {
```
steveodonovan, line 183:
```
if (webSocketConnected) {
...
```

src/containers/Condition/Condition.js
```diff
@@ -21,116 +21,95 @@ import {
...
-    } = prevMatch.params;
-
-    if (
-      namespace !== prevNamespace ||
-      conditionName !== prevConditionName ||
-      (webSocketConnected && prevWebSocketConnected === false)
-    ) {
-      this.fetchData();
-    }
+function AdditionalContent({ condition, intl }) {
```
steveodonovan, line 29: Naming nit, additional content isn't obvious, `ConditionParameters` or similar maybe ?

src/containers/About/About.js
```diff
@@ -84,143 +93,139 @@ export /* istanbul ignore next */ class About extends Component {
...
+      header: intl.formatMessage({
+        id: 'dashboard.tableHeader.value',
+        defaultMessage: 'Value'
+      })
+    }
+  ];
+
+  const getRow = (property, value) => {
+    const displayValue = getDisplayValue(value);
+    return displayValue
```
steveodonovan, line 115: `displayValue && `{ id: property, property, value: displayValue }`

src/containers/App/App.js
AlanGreene: By default `useEffect` is called after every render (including the first one after the component is mounted). We're limiting that here by passing `webSocketConnected` in runs after the first render and then only after subsequent renders where the value of `webSocketConnected` has changed.

[`webSocketConnected` is initially `null`](https://github.com/tektoncd/dashboard/blob/4d23fc20367beb5e4370ca89f28dbef0486d0d12/src/reducers/notifications.js#L15) until we attempt to connect. We want to call `fetchConfig` when the `App` component mounts, regardless of the websocket connection status, and later only on reconnect.

As part of the switch to `react-query` we'll also be ditching these status booleans and manual checks, letting react-query handle most of the reconnect / retry logic for us.

src/containers/Condition/Condition.js
AlanGreene: updated

src/containers/About/About.js
AlanGreene: updated

steveodonovan (approved):
/approve
/lgtm

AlanGreene:
First part of https://github.com/tektoncd/dashboard/issues/2000

Issue #2000, opened 2021-04-12
Refactor containers to be functional components


AlanGreene:
Refactor the [containers](https://github.com/tektoncd/dashboard/tree/main/src/containers) to be functional components so that we can use the React Hooks API.

This is a pre-requisite for [adopting `react-query`](https://github.com/tektoncd/dashboard/issues/1842) and other libraries which only provide a Hooks-based API.

- [x] About
- [x] App
- [x] ClusterTasks
- [x] ClusterTasksDropdown
- [x] ClusterTriggerBinding
- [x] ClusterTriggerBindings
- [x] Condition
- [x] Conditions
- [x] CreatePipelineResource
- [x] CreatePipelineRun
- [x] CreateTaskRun
- [x] CustomResourceDefinition
- [x] EventListener
- [x] EventListeners
- [x] ~~Extension~~ N/A
- [x] ImportResources
- [x] PipelineResource
- [x] PipelineResources
- [x] PipelineResourcesDropdown
- [x] PipelineRun
- [x] PipelineRuns
- [x] Pipelines
- [x] PipelinesDropdown
- [x] ResourceList
- [x] ServiceAccountsDropdown
- [x] SideNav
- [x] TaskRun
- [x] TaskRuns
- [x] Tasks
- [x] TasksDropdown
- [x] TriggerBinding
- [x] TriggerBindings
- [x] TriggerTemplate
- [x] TriggerTemplates

This was started in https://github.com/tektoncd/dashboard/pull/1984

AlanGreene:
@wingyplus that would be great, thank you

wingyplus:
@AlanGreene I just see that Tasks container lack of tests and storybook. Do we need to add tests (and storybook) before begin refactoring it?

AlanGreene:
The storybook is primarily focused on developing and testing the presentational components (from `packages/components`) in isolation, so there's no need to add an entry for a page-level container.

Many of the containers don't currently have tests as they follow a very similar pattern and usually just interact with the redux code which is already well tested. If there are particular pieces you feel would benefit from additional tests feel free to add those.

[...]

PR #2013, opened 2021-04-27
Add custom hook to refetch data when websocket reconnects

AlanGreene:
# Changes

## NOTE: ignore whitespace for easier review

https://github.com/tektoncd/dashboard/issues/2000

Add a custom React hook that can be used to perform an action
when the websocket is reconnected (but not on initial connection).
This is useful in our containers so we can avoid refetching when
the websocket disconnects (default behaviour if we use `webSocketConnected`
as a dependency for `useEffect`).

This avoid unnecessary and unwanted requests while keeping the container
logic relatively simple. In future with `react-query` or similar libraries
we can easily change the central `useWebSocketReconnected` hook as needed
without having to update each container.

Continue refactoring containers to be functional components, using
the new hook as appropriate.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/ClusterTriggerBinding/ClusterTriggerBinding.js
+++ b/src/containers/ClusterTriggerBinding/ClusterTriggerBinding.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Tekton Authors
+Copyright 2020-2021 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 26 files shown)


src/containers/ClusterTriggerBinding/ClusterTriggerBinding.js
```diff
@@ -11,123 +11,107 @@ See the License for the specific language governing permissions and
...
-      >
-        <Table
-          title={intl.formatMessage({
-            id: 'dashboard.parameters.title',
-            defaultMessage: 'Parameters'
-          })}
-          headers={headersForParameters}
-          rows={rowsForParameters}
-          size="short"
-          selectedNamespace={selectedNamespace}
```
briangleeson, line 124: Question, was passing this prop actually a bug?

AlanGreene: The `selectedNamespace` prop was never actually set (because it's a cluster-scoped resource) so this was always undefined. Removed it here to avoid confusion. In fact I'll probably update the `Table` component in a future PR to accept a single `emptyText` prop since it shouldn't really be aware of namespaces at all.

PR #2020, opened 2021-05-05
Refactor Task component to be functional component

wingyplus:
# Changes

This changes is a part of #2000 that convert `Task` component to use React hook (functional component) instead of class component and passes all current component tests.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/Task/Task.js
+++ b/packages/components/src/components/Task/Task.js
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React, { Component } from 'react';
+import React, { useEffect, useState } from 'react';
 import {
   PendingFilled20 as DefaultIcon,
   ChevronDown20 as ExpandIcon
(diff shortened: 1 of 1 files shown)


AlanGreene:
Thanks for the PR @wingyplus. Is there a particular issue / problem this change is trying to address? It would be really helpful to understand why this change is being proposed.

wingyplus:
> Thanks for the PR @wingyplus. Is there a particular issue / problem this change is trying to address? It would be really helpful to understand why this change is being proposed.

I'm sorry that I forget to mention the issue https://github.com/tektoncd/dashboard/issues/2000 in the PR description (just mentioned in git commit message).

AlanGreene:
Thanks for the update. As mentioned in the description of that issue the refactoring only relates to the containers (i.e. page components in `src/containers`). This is for some planned changes that require use of React Hooks in those components for interacting with our APIs and application state. We're not currently planning on updating the presentational components (in `packages/components`) as they are unaware of the API/state level changes and do not need to be modified.

There's a list in the issue description of the containers remaining to be refactored. If you would like to refactor one of those the help would be very welcome. Please add a comment on the issue identifying which container you're updating so we don't end up duplicating effort.

I'll close this PR since this particular change is not something we want to do right now.

wingyplus:
Thanks for your clarification. I will take a look and pick one of them. :)

PR #2022, opened 2021-05-07
Refactor Tasks container component to be functional component

wingyplus:
# Changes

Refactor `Tasks` container component to use React hook instead. 

Updates #2000

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/Tasks/Tasks.js
+++ b/src/containers/Tasks/Tasks.js
@@ -11,18 +11,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React, { Component } from 'react';
+import React, { useEffect, useState } from 'react';
 import { Link } from 'react-router-dom';
 import { connect } from 'react-redux';
 import { injectIntl } from 'react-intl';
-import isEqual from 'lodash.isequal';
 import keyBy from 'lodash.keyby';
 import { Button } from 'carbon-components-react';
 import {
   ALL_NAMESPACES,
   getFilters,
   getTitle,
-  urls
+  urls,
+  useWebSocketReconnected
 } from '@tektoncd/dashboard-utils';
 import {
   DeleteModal,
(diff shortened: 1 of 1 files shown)


src/containers/Tasks/Tasks.js
```diff
@@ -35,46 +35,44 @@ import {
...
+    filters,
+    namespace,
+    fetchTasks,
+    error,
+    intl,
+    tasks,
+    loading,
+    isReadOnly,
+    namespace: selectedNamespace,
+    webSocketConnected
```
AlanGreene, line 61: nit: these should be in alphabetical order so they're easier to read / quickly scan

src/containers/Tasks/Tasks.js
```diff
@@ -35,46 +35,44 @@ import {
...
-  componentDidMount() {
-    document.title = getTitle({ page: 'Tasks' });
-    this.fetchData();
-  }
+  const [deleteError, setDeleteError] = useState(null);
+  const [showDeleteModal, setShowDeleteModal] = useState(false);
+  const [toBeDeleted, setToBeDeleted] = useState([]);
+  const [cancelSelection, setCancelSelection] = useState(() => {
+    return () => {};
+  });
```
AlanGreene, line 69: the initial value can be `null` here

src/containers/Tasks/Tasks.js
```diff
@@ -85,198 +83,197 @@ export /* istanbul ignore next */ class Tasks extends Component {
...
   }
 
-  render() {
-    const { loading, tasks, intl, namespace: selectedNamespace } = this.props;
-    const { showDeleteModal, toBeDeleted } = this.state;
+  function openDeleteModal(selectedRows, handleCancelSelection) {
+    const resourcesById = keyBy(tasks, 'metadata.uid');
+    setShowDeleteModal(true);
+    setToBeDeleted(selectedRows.map(({ id }) => resourcesById[id]));
+    setCancelSelection(() => handleCancelSelection);
```
AlanGreene, line 138: can just past the reference to the function here
`setCancelSelection(handleCancelSelection);`

src/containers/Tasks/Tasks.js
```diff
@@ -35,46 +35,44 @@ import {
...
+  const {
+    filters,
+    namespace,
+    fetchTasks,
+    error,
+    intl,
+    tasks,
+    loading,
+    isReadOnly,
+    namespace: selectedNamespace,
```
AlanGreene, line 60: no need for the alias here, we can leave it as `namespace`

AlanGreene (changes requested):
Thanks for the PR @wingyplus. I've left a few small comments, let me know if you have any questions.

src/containers/Tasks/Tasks.js
wingyplus: @AlanGreene I have been try your approach but I found the error:

Screen Shot 2564-05-10 at 19 55 27

I've been try deleting task in `ClusterTasks` and found the same error also:

Screen Shot 2564-05-10 at 19 52 28

I think it's because of `React.useState` expecting to return the next state if we pass a function to the first argument.  That's why I calls `setCancelSelection` with a function that return a callback. And the reason that I set the initial state to an empty function is because of we can prevent nullable check. 

This is reference that I read about `useState`: https://reactjs.org/docs/hooks-reference.html#functional-updates.

Please correct me if I'm wrong. ^^

src/containers/Tasks/Tasks.js
wingyplus: Fix

src/containers/Tasks/Tasks.js
wingyplus: @AlanGreene I've explained below. :)

src/containers/Tasks/Tasks.js
AlanGreene: ah of course, you're right 👍 I forgot about the lazy initialisation with `useState`

src/containers/Tasks/Tasks.js
AlanGreene: It should still be safe to set the initial value to null here, this would match the previous behavior of the class-based component where there was no initial value for this state value.

src/containers/Tasks/Tasks.js
wingyplus: Do you want me to send a PR to fix in `ClusterTasks`?

AlanGreene: yes that would be great, thank you!

wingyplus: The PR is here https://github.com/tektoncd/dashboard/pull/2026. ^^

AlanGreene (approved):
/lgtm
/meow

PR #2037, opened 2021-05-13
Refactor 'create run' containers to functional components

AlanGreene:
# Changes

#2000 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/CreateTaskRun/CreateTaskRun.js
+++ b/src/containers/CreateTaskRun/CreateTaskRun.js
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { connect } from 'react-redux';
 import keyBy from 'lodash.keyby';
 import {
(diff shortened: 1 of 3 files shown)


src/containers/CreateTaskRun/CreateTaskRun.js
```diff
@@ -223,29 +257,31 @@ class CreateTaskRun extends React.Component {
...
-  handleClose = () => {
-    const { defaultNamespace: namespace, history } = this.props;
-    const { kind, taskName } = this.getTaskDetails();
+  function handleClose() {
+    const { kind: taskKind, taskName } = getTaskDetails();
     let url = urls.taskRuns.all();
-    if (taskName && namespace !== ALL_NAMESPACES) {
-      url = urls.taskRuns[kind === 'ClusterTask' ? 'byClusterTask' : 'byTask']({
-        namespace,
+    if (taskName && defaultNamespace !== ALL_NAMESPACES) {
```
briangleeson, line 269: I notice here that we are comparing `defaultNamespace` from props in the if statement, whereas in [CreatePipelineRun in the handleClose function](https://github.com/tektoncd/dashboard/pull/2037/files?diff=split&w=1#diff-a76903232a0bbce3293cb530f3aac343e2b5d6dda21048219bb2467ce9b5027dR255) we check against `namespace` from the state.

Should these match each other? The old code suggests we should be checking against `defaultNamespace` in both cases?

briangleeson (changes requested):
One query about handleClose functions

src/containers/CreateTaskRun/CreateTaskRun.js
AlanGreene: nice catch, thanks. Fixed

Issue #2047, opened 2021-05-21
Remove ClusterRole aggregationRule from default installs


AlanGreene:
[`base/200-clusterrole-extensions.yaml`](https://github.com/tektoncd/dashboard/commit/282b291375d6b40bb9e8e6f82e00bbaa41893e31#diff-5914806ab704c5e9f84cb0f412680c12d2cab0b6938ce4243f67999206961949) defines a `ClusterRole` with an `aggregationRule` that was intended to make it easier for users of Dashboard extensions to add the additional roles required by their extension without having to edit the existing Dashboard roles / bindings.

Having this as part of the default manifests means that users must have access similar to a cluster-admin role in order to install the Dashboard. This is also true of the Tekton Operator (see https://github.com/tektoncd/operator/issues/131) and other projects.

Instead, we should:
- remove this file from the default manifests
- remove references to it from [`scripts/installer`](https://github.com/tektoncd/dashboard/blob/c0e5a4fe05ead7e40c6f9345861fa8822b34d8f0/scripts/installer)
- and document the required changes in [`docs/extensions.md`](https://github.com/tektoncd/dashboard/blob/facf7b949ed49d99bf1c6d931ed7d41e7fdad80d/docs/extensions.md) for those users who are using extensions

AlanGreene:
Instead of removing the content we can move it behind an installer flag so there's still an easy way for users to 
generate the required rbac resources, something like `--extensions-rbac`. Documentation should highlight this and describe the requirements if users want to handle it manually instead.
/assign

AlanGreene:
Change ready and tested locally, will open PR after https://github.com/tektoncd/dashboard/pull/2053 is merged to avoid conflicts since it also modifies similar parts of the installer script.

PR #2059, opened 2021-05-24
Remove the ClusterRole aggregation from the default manifests

AlanGreene:
# Changes

Closes https://github.com/tektoncd/dashboard/issues/2047

ClusterRole aggregation is only used to support Dashboard extensions
and requires additional permissions on the cluster. This causes
problems in many scenarios and currently affects all users even if
they're not using extensions.

Move the aggregation rule behind an installer flag so users who
need it can still easily generate the required config but other
users benefit from the reduced permissions required for a default
install.

Document the flag in the extensions doc and describe the requirements
for achieving the same result manually for users not using the
installer script.

Also update the installer docs and help command to describe the
new option.

# Submitter Checklist

review them:

for more details._


docs/dev/installer.md
```diff
@@ -63,24 +63,25 @@ Accepted commands:
 
 Accepted options:
         [--debug]                               Prints additional messages in the console
+        [--extensions-rbac]                     Enable ClusterRole aggregation for easier management of extensions RBAC
+        [--external-logs <logs-provider-url>]   External url to fetch logs from when not available in the cluster
```
briangleeson, line 67: I feel like this sentence doesnt read right: `External url to fetch logs from when not available in the cluster`

Better? `External url from which to fetch logs, when logs are not available in the cluster`

docs/extensions.md
```diff
@@ -94,6 +94,8 @@ rules:
 EOF
 ```
 
+Alternatively, if the `--extensions-rbac` is not used, ommit the label from the ClusterRole, and define the appropriate RoleBinding or ClusterRoleBinding yourself with the subject set to the `tekton-dashboard` service account.
```
briangleeson, line 97: Typo for `omit`

scripts/installer
```diff
@@ -476,24 +498,25 @@ help () {
   echo -e ""
   echo -e "Accepted options:"
   echo -e "\t[--debug]\t\t\t\tPrints additional messages in the console"
+  echo -e "\t[--extensions-rbac]\t\t\tEnable ClusterRole aggregation for easier management of extensions RBAC"
+  echo -e "\t[--external-logs <logs-provider-url>]\tExternal url to fetch logs from when not available in the cluster"
```
briangleeson, line 502: Same comment as before with wording of this sentence

briangleeson (changes requested):
Couple of comments on Strings

docs/extensions.md
AlanGreene: 🙈

Issue #1972, opened 2021-03-05
Syntax highlighting for YAML display


AlanGreene:
# Is your feature request related to a problem ?

Large blocks of YAML can be difficult to scan quickly.

# Describe the solution you'd like

Add syntax highlighting for the `ViewYAML` component to make it easier to read.

We could use something like [`react-syntax-highlighter`](https://www.npmjs.com/package/react-syntax-highlighter) to achieve this. It's already used by [tektoncd/hub](https://github.com/tektoncd/hub/blob/eae5fc9461c451ea77b6c9745af5460d4db9c4c7/ui/src/components/Yaml/index.tsx).

Can we adapt our existing colour palette from the `LogFormat` component to work here or is there a suitable preset theme we can use that matches the Dashboard reasonably well (with or without tweaking)?

# Describe alternatives you've considered

https://www.npmjs.com/package/highlight.js
https://www.npmjs.com/package/prismjs

`react-syntax-highlighter` has support for both of these and provide a nice React API.

# Additional context

wingyplus:
I've tried it with the atom dark theme.

<img width="1667" alt="Screen Shot 2564-04-28 at 22 20 13" src="https://user-images.githubusercontent.com/484530/116429267-ee6eb100-a86f-11eb-8058-0c1edc5f54f4.png">

After I try to fix the background color to match with previous output. It become:

<img width="1673" alt="Screen Shot 2564-04-28 at 22 50 43" src="https://user-images.githubusercontent.com/484530/116433965-3263b500-a874-11eb-9bb2-ab5fec525119.png">

This is diff that I try (I'm not expert at JS and CSS but I try my best):

```diff
diff --git a/packages/components/package.json b/packages/components/package.json
index 87e3fa3..a14045e 100644
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -4,7 +4,12 @@
   "author": {
     "name": "The Tekton Authors"
   },
-  "keywords": ["tekton", "tektoncd", "components", "react"],
+  "keywords": [
+    "tekton",
+    "tektoncd",
+    "components",
+    "react"
+  ],
   "license": "Apache-2.0",
   "private": false,
   "main": "./dist/components/index.js",
@@ -26,6 +31,7 @@
     "linkify-it": "^3.0.2",
     "prop-types": "^15.7.2",
     "react-intl-formatted-duration": "^4.0.0",
+    "react-syntax-highlighter": "^15.4.3",
[...]

PR #2060, opened 2021-05-25
Add syntax highlight to ViewYAML

wingyplus:
# Changes

Replacing the current code tag with SyntaxHighlight from
react-syntax-highlight package. The SyntaxHighlight needs to set
codeTagProps to empty object to prevent SyntaxHighlight add inline
style. And add style in ViewYAML.scss to handle light & dark theme.

The code is still need to cleanup. Just open to see visibility and tweaking
color style.

Closes #1972

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/main/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/main/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/main/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/dashboard/blob/main/CONTRIBUTING.md)
for more details._

PR #2071, opened 2021-05-29
Add syntax highlight to ViewYAML component

wingyplus:
# Changes

Add `react-syntax-highlighter` to the `src/components` and use it as
code tag in ViewYAML. The syntax highlighter will colorize a string in
YAML when `enableSyntaxHighlighter` set it to `true`.

Updates #1972

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/ViewYAML/ViewYAML.js
+++ b/packages/components/src/components/ViewYAML/ViewYAML.js
@@ -15,29 +15,54 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import jsYaml from 'js-yaml';
 import classNames from 'classnames';
+import { Prism as SyntaxHighlight } from 'react-syntax-highlighter';
 
-const ViewYAML = props => {
-  const { className, dark, resource } = props;
-
+function YAMLHighlighter({ children, className }) {
   return (
-    <div
-      className={classNames('bx--snippet--multi', className, {
-        'tkn--view-yaml--dark': dark
-      })}
+    <SyntaxHighlight
+      className={className}
+      language="yaml"
+      useInlineStyles={false}
+      codeTagProps={{}}
     >
+      {children}
+    </SyntaxHighlight>
+  );
+}
+
+function YAMLRaw({ children, className }) {
+  return (
+    <div className={className}>
       <code>
-        <pre>{jsYaml.dump(resource)}</pre>
+        <pre>{children}</pre>
       </code>
     </div>
   );
-};
+}
+
+function ViewYAML({ className, dark, enableSyntaxHighlighting, resource }) {
+  const clz = classNames('bx--snippet--multi', className, {
+    'tkn--view-yaml--dark': dark
+  });
+  const yaml = jsYaml.dump(resource);
+
+  if (enableSyntaxHighlighting) {
+    return <YAMLHighlighter className={clz}>{yaml}</YAMLHighlighter>;
+  }
+  return <YAMLRaw className={clz}>{yaml}</YAMLRaw>;
+}
 
 ViewYAML.propTypes = {
   resource: PropTypes.oneOfType([
     PropTypes.array,
     PropTypes.shape({}),
     PropTypes.string
-  ]).isRequired
+  ]).isRequired,
+  enableSyntaxHighlighting: PropTypes.bool
+};
+
+ViewYAML.defaultProps = {
+  enableSyntaxHighlighting: false
 };
 
 export default ViewYAML;
--- a/packages/components/src/components/ViewYAML/ViewYAML.scss
+++ b/packages/components/src/components/ViewYAML/ViewYAML.scss
@@ -15,3 +15,12 @@ limitations under the License.
   background-color: $gray-90; // TODO: $inverse-02 - see Log.scss
   color: $gray-10;            //       $inverse-01
 }
+
+// TODO: Colorize YAML syntax to conform carbon design. See
+// https://github.com/carbon-design-system/carbon-website/issues/965 for more
+// details.
+.bx--snippet--multi {
+  .token.string {
+    color: rgb(0, 187, 0);
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -12661,6 +12661,7 @@
         "linkify-it": "^3.0.2",
         "prop-types": "^15.7.2",
         "react-intl-formatted-duration": "^4.0.0",
+        "react-syntax-highlighter": "^15.4.3",
         "react-window": "^1.8.6",
         "tlds": "^1.208.0"
       },
@@ -12793,6 +12794,18 @@
           "resolved": "https://registry.npmjs.org/is-number/-/is-number-7.0.0.tgz",
           "integrity": "sha512-41Cifkg6e8TylSpdtTpeLVMqvSBEVzTttHvERD741+pnZ8ANv0004MRL43QKPDlK9cGvNp6NZWZUBlbGXYxxng==",
           "optional": true
+        },
+        "react-syntax-highlighter": {
+          "version": "15.4.3",
+          "resolved": "https://registry.npmjs.org/react-syntax-highlighter/-/react-syntax-highlighter-15.4.3.tgz",
+          "integrity": "sha512-TnhGgZKXr5o8a63uYdRTzeb8ijJOgRGe0qjrE0eK/gajtdyqnSO6LqB3vW16hHB0cFierYSoy/AOJw8z1Dui8g==",
+          "requires": {
+            "@babel/runtime": "^7.3.1",
+            "highlight.js": "^10.4.1",
+            "lowlight": "^1.17.0",
+            "prismjs": "^1.22.0",
+            "refractor": "^3.2.0"
+          }
         },
         "react-window": {
           "version": "1.8.6",
@@ -13193,7 +13206,6 @@
       "version": "2.3.1",
       "resolved": "https://registry.npmjs.org/@types/hast/-/hast-2.3.1.tgz",
       "integrity": "sha512-viwwrB+6xGzw+G1eWpF9geV3fnsDgXqHG+cqgiHrvQfDUW5hzhCyV7Sy3UJxhfRFBsgky2SSW33qi/YrIkjX5Q==",
-      "dev": true,
       "requires": {
         "@types/unist": "*"
       }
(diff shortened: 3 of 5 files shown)


wingyplus:
Currently, the `ViewYAML` test will fail because of `react-syntax-highlighter`. The `getByText` function looks hard to test its separates by `SyntaxHighlight` component. Is a test snapshot match for us in this case?

AlanGreene:
Apologies for the delay reviewing this @wingyplus, I thought it was https://github.com/tektoncd/dashboard/pull/2060 which was marked as a work in progress so hadn't looked too closely.

/ok-to-test

packages/components/src/components/ViewYAML/ViewYAML.js
```diff
@@ -15,20 +15,23 @@ import React from 'react';
...
-  const { className, dark, resource } = props;
+  const { className, dark, resource, enableSyntaxHighlighting = false } = props;
 
   return (
-    <div
+    <SyntaxHighlight
       className={classNames('bx--snippet--multi', className, {
-        'tkn--view-yaml--dark': dark
+        'tkn--view-yaml--dark': dark,
+        'tkn--view-yaml--enable-syntax-highlighting': enableSyntaxHighlighting
```
AlanGreene, line 27: no need for this line as the additional classes will only be added to the content when highlighting is enabled anyway

packages/components/src/components/ViewYAML/ViewYAML.js
```diff
@@ -15,20 +15,23 @@ import React from 'react';
...
       })}
+      codeTagProps={{}}
```
AlanGreene, line 29: from the docs:

> `codeTagProps` - props that will be spread into the `<code>` tag that is the direct parent of the highlighted code elements. Useful for styling/assigning classNames.

Seems like it should be safe to omit this. Did you run into problems if you left it out?

packages/components/src/components/ViewYAML/ViewYAML.js
```diff
@@ -15,20 +15,23 @@ import React from 'react';
...
+      language="yaml"
+      useInlineStyles={false}
     >
-      <code>
-        <pre>{jsYaml.dump(resource)}</pre>
-      </code>
-    </div>
+      {jsYaml.dump(resource).trim()}
```
AlanGreene, line 33: Was the `trim()` added for a particular reason? It would be good to understand if this is necessary.

packages/components/src/components/ViewYAML/ViewYAML.scss
```diff
@@ -15,3 +15,12 @@ limitations under the License.
   background-color: $gray-90; // TODO: $inverse-02 - see Log.scss
   color: $gray-10;            //       $inverse-01
 }
+
+// TODO: Colorize YAML syntax to conform carbon design. See
+// https://github.com/carbon-design-system/carbon-website/issues/965 for more
+// details.
+.bx--snippet--multi.tkn--view-yaml--enable-syntax-highlighting {
```
AlanGreene, line 22: see comment above, the `enable-syntax-highlighting` class shouldn't be required.

AlanGreene (changes requested):
Thanks for the PR @wingyplus, I've left a few comments below.

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: @AlanGreene I found that if we omit this. SyntaxHighlight will add some inline style to the element. It can cause we cannot be styling syntax highlight in sass world.

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: I found that it has some extra bottom padding compare to the others. It fixes after I trim it. oTZ

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: Thank you for your review. This would be a good choice. ^^

Let me take a look tomorrow.

packages/components/src/components/ViewYAML/ViewYAML.js
AlanGreene: It's probably a trailing newline character which we should leave if possible, so we're accurately rendering the content provided.

wingyplus:
@AlanGreene The unit tests fail because the text content was change from plain YAML code to the code that tokenized by SyntaxHighlight component. It might be hard to match the content like the existing one. Do you have any suggestions?

wingyplus:
> Apologies for the delay reviewing this @wingyplus, I thought it was #2060 which was marked as a work in progress so hadn't looked too closely.
> 
> /ok-to-test

@AlanGreene I'm sorry that my PR make you confuse. I close that one because the implementation seems different.

AlanGreene:
> The unit tests fail because the text content was change from plain YAML code to the code that tokenized by SyntaxHighlight component. It might be hard to match the content like the existing one. Do you have any suggestions?

If you follow my suggestion about only using the SyntaxHighlighter component when the flag is enabled the existing tests should pass again. We would then need to add some additional tests for the ViewYAML component to test when the flag is enabled.

In future if we switch the default behaviour we would need to update the other tests, but that's expected as we would be changing behaviour.

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: Ok. So I will remove the trim out of it.

wingyplus:
> > The unit tests fail because the text content was change from plain YAML code to the code that tokenized by SyntaxHighlight component. It might be hard to match the content like the existing one. Do you have any suggestions?
> 
> If you follow my suggestion about only using the SyntaxHighlighter component when the flag is enabled the existing tests should pass again. We would then need to add some additional tests for the ViewYAML component to test when the flag is enabled.
> 
> In future if we switch the default behaviour we would need to update the other tests, but that's expected as we would be changing behaviour.

Ahh, I'm sorry that I didn't see it in detail. Your suggestion should work as expected. ^^

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: This is the one issue if I doesn't add `codeTagProps: {{}}`:

Screen Shot 2564-06-09 at 22 47 37

The syntax highlighter add some inline styling even set `useInlineStyles` to `false`.

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: @AlanGreene After I try, I found that we cannot replace the container element easily because the child structure is not the same. So I refactoring it to 2 render functions and select render function base on `enableSyntaxHighlighter` like this https://github.com/tektoncd/dashboard/pull/2071/files#diff-7d3c10c7cf56e92d73518e8412053ccc766adc892693aedc504249e8c78c50d5R23-R44. 

What do you think?

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: @AlanGreene I think it's not because of final new line in yaml string but it's because of syntax highlight itself. Try this example from storybook:

Raw yaml code:

Screen Shot 2564-06-09 at 23 02 47

Syntax highlight enable:

Screen Shot 2564-06-09 at 23 03 12

And if we try react syntax highlighter demo site https://react-syntax-highlighter.github.io/react-syntax-highlighter/demo/. We'll see the same (note that they have final new line in editor, we must remove it first and see it).

wingyplus:
@AlanGreene the code review was addressed. The test added. Please kindly review.

packages/components/src/components/ViewYAML/ViewYAML.js
```diff
@@ -15,29 +15,54 @@ import React from 'react';
...
+  const clz = classNames('bx--snippet--multi', className, {
+    'tkn--view-yaml--dark': dark
+  });
+
+  const yaml = jsYaml.dump(resource);
+
+  if (enableSyntaxHighlighting) {
+    return renderSyntaxHighlight(clz, yaml);
+  } else {
+    return renderRaw(clz, yaml);
```
AlanGreene, line 55: test are failing due to a lint error here:

```
/home/prow/go/src/github.com/tektoncd/dashboard/packages/components/src/components/ViewYAML/ViewYAML.js
  54:10  error  Unnecessary 'else' after 'return'  no-else-return
```

packages/components/src/components/ViewYAML/ViewYAML.js
```diff
@@ -15,29 +15,54 @@ import React from 'react';
...
+      <SyntaxHighlight
+        className={clz}
+        language="yaml"
+        useInlineStyles={false}
+        codeTagProps={{}}
+      >
+        {yaml}
+      </SyntaxHighlight>
+    );
+  };
```
AlanGreene, line 44: Since these functions receive all of their required inputs as params anyway, they can be defined outside the component to avoid unnecessarily redefining them on every render

They could also be modified slightly to be their own React components:

```
function YAMLHighlighter({ children, className }) {
  return (
    <SyntaxHighlight
      className={className}
      language="yaml"
      useInlineStyles={false}
      codeTagProps={{}}
    >
      {children}
    </SyntaxHighlight>
  );
}

function ViewYAML(props) {
…
  if (enableSyntaxHighlighting) {
    return <YAMLHighlighter className={className}>{yaml}</YAMLHighlighter>;
  }
…
};
```

packages/components/src/components/ViewYAML/ViewYAML.js
```diff
@@ -15,29 +15,54 @@ import React from 'react';
...
-        <pre>{jsYaml.dump(resource)}</pre>
-      </code>
-    </div>
-  );
+  const { className, dark, resource, enableSyntaxHighlighting = false } = props;
+
+  const renderRaw = (clz, yaml) => {
+    return (
+      <div>
+        <code className={clz}>
```
AlanGreene, line 26: the classes were applied to the div before, is there a reason for moving them to the code element instead?

AlanGreene (changes requested):

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: Oh, I'm sorry, my bad.

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: @AlanGreene The render functions already extracted and I change `ViewYAML` component to be function instead of const variable. Please kindly review again.

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: Thank you for your help. I won't forget to run lint next time. :)

packages/components/src/components/ViewYAML/ViewYAML.js
wingyplus: Fixed

packages/components/src/components/ViewYAML/ViewYAML.js
AlanGreene: No worries, it can be easy to miss. Depending on the editor you're using it might be useful to install a linter plugin that's compatible with our eslint / prettier config so you get feedback while you're editing. I find these very helpful.

I'll take a look at our webpack config too as we should be catching lint error during the dev build.

AlanGreene (approved):
This looks great, thanks @wingyplus 
/meow

PR #2084, opened 2021-06-09
Update DataTableSkeleton to support `size` prop

AlanGreene:
# Changes

Update DataTableSkeleton to support a `size` prop with the same
values supported by the Carbon `Table` component, and update our
`Table` wrapper component to pass the value through.

Add control for the supported size values to the storybook.

This will be used on the updated About page loading state as part
of the react-query migration.

# Submitter Checklist

review them:

for more details._


LyndseyBu:
looks good, only noticed one thing with the sizes, they seem to be removing the previous terms (compact, short....) in the next major release...

image

AlanGreene:
Thanks @LyndseyBu, I'll remove those so we don't start using them. A Carbon 11 prerelease should be available for testing in the next few weeks but I don't expect it to GA for a while yet.

PR #2085, opened 2021-06-09
Add initial react-query utilities

AlanGreene:
# Changes

~~Depends on https://github.com/tektoncd/dashboard/pull/2084 due to dependency updates.~~ merged

https://github.com/tektoncd/dashboard/issues/1842

Add a custom hook for fetching a list of namespaced resources
similar to our existing `fetchNamespacedCollection` action creator.
The difference with this is that we won't need action creators,
selectors, or the reducer backing it, all of the caching functionality
is handled internally by react-query while giving us control over
config (e.g. cache time, stale time, retries) if desired.

Add a custom hook to handle websocket events, updating the react-query
cache directly for any matching queries. This is a replacement for
the existing websocket middleware and events handlers we've added with
redux.

These are not yet wired into the application, this will be handled
in another PR.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/api/utils.test.js
+++ b/src/api/utils.test.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2019-2020 The Tekton Authors
+Copyright 2019-2021 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


src/api/utils.test.js
```diff
@@ -143,3 +148,223 @@ describe('checkData', () => {
...
+  const kind = 'TaskRun';
+  it('should handle Created events', () => {
+    const queryClient = getQueryClient();
+    const webSocket = getWebSocket();
+
+    const wrapper = ({ children }) => (
+      <APIWrapper queryClient={queryClient} webSocket={webSocket}>
+        {children}
+      </APIWrapper>
+    );
```
wingyplus, line 162: How about we extract it to a function like:

```
// return a wrapper function.
function createWrapper(queryClient, WebSocket) {
  return ({children}) => <APIWrapper queryClient={queryClient} webSocket={webSocket}>{children}</APIWrapper>;
}
```

And then call `createWrapper(queryClient, WebSocket)` in renderHook instead?

AlanGreene: I like this, I've extracted it to a function as you suggested 👍

PR #2093, opened 2021-06-18
Convert Triggers resources to use react-query

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1842

Update all Triggers resources pages (except for ClusterInterceptors)
to use react-query, removing their redux-related code.

From an end-user point of view functionality should be the same,
however we're now deduplicating requests, avoiding unnecessary
requests on page navigation, and pruning the cache of stale data
in the background. As a result, users may notice improved
performance for these pages and lower browser memory consumption
over long-lived sessions.

# Submitter Checklist

review them:

for more details._


src/api/utils.js
```diff
@@ -118,40 +168,32 @@ export function useWebSocket(resourceType) {
...
-export function useNamespacedCollection(resourceType, api, params) {
-  const query = useQuery([resourceType, params], () => api(params));
-  useWebSocket(resourceType);
+export function useCollection(kind, api, params) {
+  const query = useQuery([kind, params], () => api(params));
+  useWebSocket({ kind });
+  return query;
+}
+
+export function useResource(kind, api, params) {
```
LyndseyBu, line 195: useCollection and useResource seem to be doing the same thing, could we use one?

AlanGreene: yes we could use one now, but in the near future `useCollection` will be updated to handle automatic chunked loading of large lists (using something like https://github.com/AlanGreene/dashboard/commit/b16612948c62b1ec5d9cb681e14898d7a7a87473#diff-6d1caeaa7a4f194561101db7781e6ff562dd9a95c9cb07852bc69ce6443af15eR165) so splitting them now seemed to make sense since we'll only have to update it here and any container calling `useCollection` will automatically benefit while any using `useResource` will be unaffected

Issue #327, opened 2019-07-09
Show latest logs and/or follow feature


castlemilk:
In the Pipeline/Task log view a large number of logs can be produced, which can become unwieldy as a user scrolls.

Would be nice from a UX perspective to either have a fixed height & scrollable window for logs that can be set to follow mode (showing latest logs as they are recieved), or a tick-box to show latest logs first.

Additionally, it might be worth investigating an optimised mechanism for rendering a sub-set window of logs, instead of fetching the entire log and attempting to render. Could possible default to showing last 1000 lines and then as user scrolls, fetch more by increase the sinceTimestamp value. Potentially utilise `react-virtualized` for this optimization.

I've seen some page freezing on large logs (>5MB), which can then be triggered to re-occur as you change between Logs|Status|Details.

AlanGreene:
Thanks for the feedback and suggestions @castlemilk, they're much appreciated.

We're currently working on some significant design updates for the details view (which includes the logs) and will keep these in mind.

tomhau01:
We have users viewing logs live so having a "follow" feature would be super convenient. Think of it like a console type of view :)

eddycharly:
There was some changes/improvements around logs recently.
It would be nice to get this one too !

AlanGreene:
Agreed this would be really nice, and with the streaming changes should also give a better UX.

@stevesoaress is currently working on some [changes to the processing and display of the log content](https://github.com/tektoncd/dashboard/issues/895) so I think it would be a good idea to hold off on implementing this until those changes stabilise to avoid conflicts / rework.

AlanGreene:
Design notes:
- default behaviour: should it follow logs by default, and stop when the user scrolls or should it be opt-in?
- should there be some UI control to toggle the following?
- if there is a UI control, when should it be displayed and where should it be located? Should it affect just the current logs or persist globally?
- review common patterns in existing solutions for reference.

Investigate existing libraries such as `react-lazylog` which provide follow functionality and determine if they're suitable as replacements for our existing logs component or whether we should keep our own and enhance it with the scroll / follow functionality.

AlanGreene:
We should keep this simple and follow by default, stopping when the user scrolls. This shouldn't require any additional dependencies.

AlanGreene:
Mostafa will be working on this.

AlanGreene:
[...]

PR #2100, opened 2021-06-23
Add scroll buttons and auto-scroll to logs

MA9521:
Related to https://github.com/tektoncd/dashboard/issues/327

# Changes

First commit (auto-scroll):
Logs of step runs can get very big. While the step run has not terminated, the p
age will keep scrolling down the page as needed. This behaviour can be stopped once the user scrolls up so that the bottom of the log is no longer visible. This behaviour can be resumed once the user scrolls back down to the log bottom.

This feature is activated by a boolean flag (which is turned to true in the dashboard ui) not to affect the users of the reusable component `<Log` in edge cases (the component followed by other elements in the page,component inside a non-window scrollable parent.....)

Second commit( Scroll buttons):
Step-runs with huge logs require the user to manually scroll all the way down the page. A click/press-to-scroll-to-page-top button will now appear when the top of the logs is above the viewport and a click/press-to-scroll-to-page-bottom button will now appear when the bottom of the logs is below the viewport.

The button presence is updated at each Log component update or scroll. For a logically-complete solution, we would need to use a ResizeObserver. This feature is behind a flag to not affect the users of the reusable `<Log` component with edge cases (a non-window-scrollable parent).

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/Log/Log.js
+++ b/packages/components/src/components/Log/Log.js
@@ -15,6 +15,7 @@ import React, { Component } from 'react';
 import { SkeletonText } from 'carbon-components-react';
 import { FixedSizeList as List } from 'react-window';
 import { injectIntl } from 'react-intl';
+import utils from './utils';
 
 import Ansi from '../LogFormat';
 
@@ -28,16 +29,76 @@ const itemSize = 15; // This should be kept in sync with the line-height in SCSS
 const defaultHeight = itemSize * 100 + itemSize / 2;
 
 export class LogContainer extends Component {
-  state = { loading: true };
+  constructor(props) {
+    super(props);
+    this.state = { loading: true };
+    this.logRef = React.createRef();
+    this.textRef = React.createRef();
+  }
 
   componentDidMount() {
     this.loadLog();
+    window.addEventListener("scroll", this.handleLogScroll, true);
+    if (this.shouldAutoScroll(this.props.stepStatus)) {
+      this.scrollToBottomLog();
+      return; // the scroll event listener will call handleLogScroll
+    }
+    this.handleLogScroll();
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (
+      prevState.logs?.length !== this.state.logs?.length ||
+      prevProps.isLogsMaximized !== this.props.isLogsMaximized
+    ) {
+      if (this.shouldAutoScroll(this.props.stepStatus, prevProps.stepStatus)) {
+        this.scrollToBottomLog();
+        return; // the scroll event listener will call handleLogScroll
+      }
+      this.handleLogScroll();
+    }
   }
 
   componentWillUnmount() {
     clearInterval(this.timer);
+    window.removeEventListener("scroll", this.handleLogScroll, true);
     this.cancelled = true;
   }
+
+  handleLogScroll = () => {
+    const isLogBottomUnseen = this.isLogBottomUnseen();
+
+    if (
+      isLogBottomUnseen !== this.state.isLogBottomUnseen
+    ) {
+      this.setState({
+        isLogBottomUnseen
+      });
+    }
+  };
+  
+  isLogBottomUnseen = () => {
+    return utils.isElementEndBelowViewBottom(this.logRef?.current)
+      || utils.hasElementPositiveScrollBottom(this.textRef?.current?.firstElementChild);
+  };
+
+  scrollToBottomLog = () => {
+    const longTextElement = this.textRef?.current?.firstElementChild;
+    if (utils.hasElementPositiveScrollBottom(longTextElement)) {
+      longTextElement.scrollTop = longTextElement.scrollHeight - longTextElement.clientHeight;
+    }
+    const rootElement = document.documentElement;
+    rootElement.scrollTop = rootElement.scrollHeight - rootElement.clientHeight;
+  };
+
+  shouldAutoScroll = (stepStatus, prevStepStatus) => {
+    return this.props.enableLogAutoScroll && !this.state.isLogBottomUnseen &&
+      (
+        (stepStatus && !stepStatus.terminated) ||
+        (stepStatus?.terminated && prevStepStatus && !prevStepStatus.terminated)
+      ) &&
+      this.isLogBottomUnseen();
+  };
 
   getLogList = () => {
     const { stepStatus, intl } = this.props;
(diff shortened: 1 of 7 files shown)


AlanGreene:
@MA9521 and I spoke about this earlier today. We're going to try a different approach with `position: sticky` and `IntersectionObserver` to see if we can simplify the code while also accounting for the case where the scroll parent is not the window.

AlanGreene:
Thanks for the PR @MA9521, I'll take a look at this today. Let me know if you have any problems with the CLA.

MA9521:
Closing then will reopen to recheck CLA

packages/components/src/components/Log/Log.scss
```diff
@@ -53,6 +53,20 @@ pre.tkn--log {
     right: 0;
   }
 
+  .bx--copy-btn.scroll-btn {
```
AlanGreene, line 56: I'm not seeing this style applied, instead the buttons just appear at the bottom of the log.

packages/components/src/components/Log/Log.js
```diff
@@ -28,17 +30,147 @@ const itemSize = 15; // This should be kept in sync with the line-height in SCSS
...
+  constructor(props) {
+    super(props);
+    this.state = { loading: true };
+    this.logRef = React.createRef();
+    this.textRef = React.createRef();
+  }
 
   componentDidMount() {
     this.loadLog();
+    window.addEventListener("scroll", this.handleLogScroll, true);
```
AlanGreene, line 42: Ideally we would only register the scroll listener when one of the features relying on it is enabled, similar with the other code here.

packages/components/src/components/Log/Log.js
```diff
@@ -28,17 +30,147 @@ const itemSize = 15; // This should be kept in sync with the line-height in SCSS
...
+      || utils.hasElementPositiveScrollTop(this.textRef?.current?.firstElementChild);
+  };
+
+  scrollToBottomLog = () => {
+    const longTextElement = this.textRef?.current?.firstElementChild;
+    if (utils.hasElementPositiveScrollBottom(longTextElement)) {
+      longTextElement.scrollTop = longTextElement.scrollHeight - longTextElement.clientHeight;
+    }
+    const rootElement = document.documentElement;
+    rootElement.scrollTop = rootElement.scrollHeight - rootElement.clientHeight;
```
AlanGreene, line 101: rather than setting the `scrollTop` directly can we use the `scrollTo` API instead?
i.e. `rootElement.scrollTo({ top: … })`, this also lets us control the scroll behaviour (smooth / auto).

packages/components/src/components/Log/Log.js
```diff
@@ -28,17 +30,147 @@ const itemSize = 15; // This should be kept in sync with the line-height in SCSS
...
+    const logElementRight = (window.innerWidth || document.documentElement.clientWidth) -
+      this.logRef?.current?.getBoundingClientRect().right;
+    document.documentElement.style.setProperty("--log-element-right",
+      `${logElementRight.toString()}px`);
+  };
+
+  getScrollButtons = () => {
+    const { enableLogScrollButtons, intl } = this.props;
+    if (!enableLogScrollButtons) {
+      return <Fragment />;
```
AlanGreene, line 131: no need to return a fragment here, `null` is fine

packages/components/src/components/Log/Log.js
```diff
@@ -28,17 +30,147 @@ const itemSize = 15; // This should be kept in sync with the line-height in SCSS
...
+    }
+
+    return (
+      <Fragment>
+        {this.state.isLogTopUnseen ? (
+          <button
```
AlanGreene, line 137: Can we use a Carbon `Button` component here? This will ensure consistent styling across future updates.

packages/components/src/components/Log/Log.js
```diff
@@ -28,17 +30,147 @@ const itemSize = 15; // This should be kept in sync with the line-height in SCSS
...
+            key="scroll-to-top-btn"
+            className="scroll-btn bx--copy-btn scroll-to-top-btn"
+            onClick={this.scrollToTopLog}
+            type="button"
+          >
+            <UpToTop16>
+              <title>
+                {intl.formatMessage({
+                  id: 'dashboard.pipelineRun.scrollToTop',
```
AlanGreene, line 146: `dashboard.log.scrollToTop` since it's not just for PipelineRuns

packages/components/src/components/Log/utils.js
```diff
@@ -0,0 +1,57 @@
+const scrollRegex = /(auto|scroll)/;
```
AlanGreene, line 1: missing copyright / license header

packages/components/src/components/Log/utils.js
```diff
@@ -0,0 +1,57 @@
...
+        return false;
+    }
+
+    const horizontalScroll = includeHorizontalScroll ? "" : getStyle(el, "overflow-x");
+    return scrollRegex.test(
+        getStyle(el, "overflow") +
+        getStyle(el, "overflow-y") + horizontalScroll);
+}
+
+const utils = {
```
AlanGreene, line 51: we can just export/import the individual functions directly, no need to wrap them in an object here

`export function isElementScrollable(…) {` etc.

src/containers/App/App.js
```diff
@@ -205,6 +205,8 @@ export function App({
   }, [isFetchingConfig, tenantNamespace]);
 
   const logoutButton = <LogoutButton getLogoutURL={() => logoutURL} />;
+  const enableLogAutoScroll = true;
```
AlanGreene, line 208: no need for these here, they can just be set in the PipelineRun / TaskRun containers.

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -284,6 +286,8 @@ export /* istanbul ignore next */ function PipelineRunContainer(props) {
         />
       )}
       <PipelineRun
+        enableLogAutoScroll={enableLogAutoScroll}
```
AlanGreene, line 289: no need to accept this as a prop, just set the attribute here
```jsx
<PipelineRun
  enableLogAutoScroll
  enableLogScrollButtons
  error={error}
…
```


same for the TaskRun container

AlanGreene (changes requested):
This doesn't seem to be working for me, I'm not seeing the scroll buttons appear at all in the Dashboard application.

If I edit the Log storybook entries I can see them on the windowed / performance stories but the positioning is not as expected.

Let's split this into 2 PRs as I mentioned before so we can deliver the auto-scroll functionality, then follow up on the scroll buttons separately.

There's also quite a bit of logic involved in determining when to show certain elements / trigger certain behaviour, it would be great to have some unit tests even if it's just for the utils so we have some coverage to give us confidence we're not going to break this in future.

packages/components/src/components/Log/Log.js
MA9521: I used the `button` tag because it is already used in https://github.com/tektoncd/dashboard/blob/aec426b0c0d7004dcff1dec627be6ee004b9ea37/packages/components/src/components/LogsToolbar/LogsToolbar.js#L27

src/containers/App/App.js
MA9521: my idea was to centralize the flag for the entire app instead of having the potential of a different behaviour in PipelineRun vs TaskRun

PR #2104, opened 2021-06-25
Restrict X-Frame-Options header with new flag

leiyanggz:
The value of the response header "X-Frame-Options" is "deny" by default in http requests because of security concerns.
However in some use cases people want to load the dashboard in their internal web applications via iframe.

# Changes
Add a string flag xframe-options to restrict the header "X-Frame-Options".

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/cmd/dashboard/main.go
+++ b/cmd/dashboard/main.go
@@ -47,6 +47,7 @@ var (
 	logFormat          = flag.String("log-format", "json", "Format for log output (json or console)")
 	streamLogs         = flag.Bool("stream-logs", false, "Enable log streaming instead of polling")
 	externalLogs       = flag.String("external-logs", "", "External logs provider url")
+	xFrameOptions      = flag.String("x-frame-options", "DENY", "Value for the X-Frame-Options response header, set '' to omit it")
 )
 
 func main() {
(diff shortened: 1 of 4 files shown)


cmd/dashboard/main.go
```diff
@@ -47,6 +47,7 @@ var (
 	logFormat          = flag.String("log-format", "json", "Format for log output (json or console)")
 	streamLogs         = flag.Bool("stream-logs", false, "Enable log streaming instead of polling")
 	externalLogs       = flag.String("external-logs", "", "External logs provider url")
+	xFrameOptions      = flag.String("xframe-options", "DENY", "Whether or not a browser should be allowed to render a page in a <frame>, <iframe>, <embed> or <object>")
```
AlanGreene, line 50: `flag.String("x-frame-options"…` so it's consistent with the header name and less likely to cause confusion / error

we should update the description too, maybe something like `Value to set for the X-Frame-Options response header ("" to omit it)`

AlanGreene (changes requested):
This looks good thanks. One small suggestion and then I think it's ready to merge.

cmd/dashboard/main.go
leiyanggz: Just updated in a new commit.

AlanGreene (approved):
Thanks @leiyanggz. Please squash your commits then we can merge this.

leiyanggz:
> Thanks @leiyanggz. Please squash your commits then we can merge this.

Done.

PR #2106, opened 2021-06-26
Enable Start and Stop for PipelineRunPending

lilley2412:
# Changes

This PR implements tektoncd#2069 to enable the 'stop' action on PipelineRuns with status 'PipelineRunPending'.  The stop action transitions a pending PipelineRun to the cancelled state.  

This PR also adds a new action named 'start' that is only enabled for PIpelineRuns with status 'PipelineRunPending'.  The start action patches the PipelineRun to remove the pending status; removing the PipelineRunPending status causes the PipelineRun to start.

The stop functionally will need to be revisited in the future for
pipeline [TEP-0058](https://github.com/tektoncd/community/blob/main/teps/0058-graceful-pipeline-run-termination.md) which introduces new status for cancellation.

# Submitter Checklist

review them:

- [X] Includes [tests](https://github.com/tektoncd/community/blob/main/standards.md#principles) (if functionality changed/added)
- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/main/standards.md#commit-messages)

for more details._


lilley2412:
Marked this as draft as this is my first PR and wasn't sure if something of this magnitude requires a design document / discussion.  Please let me know if it's good to proceed!

AlanGreene:
Thanks for the PR @lilley2412. From a first glance this seems like a reasonable change. Once suggestion, the Start and Rerun components are very similar, would it be possible to refactor to use a common RunAction component or similar since we're likely to have additional actions in future. We may also consider using an overflow menu instead of inline buttons if we end up with 3 or more items in future.

/ok-to-test

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -187,6 +193,40 @@ export /* istanbul ignore next */ function PipelineRuns(props) {
...
+        actionText: intl.formatMessage({
+          id: 'dashboard.startPipelineRun.actionText',
+          defaultMessage: 'Start'
+        }),
+        action: start,
+        disable: resource => {
+          const { reason, status } = getStatus(resource);
+          return !isPending(reason, status);
+        },
+        modalProperties: {
```
AlanGreene, line 206: I think we can leave out the `modalProperties` here, no need to confirm with the user since they're not deleting anything or performing an action they can't cancel or otherwise recover from.

lilley2412:
> the Start and Rerun components are very similar, would it be possible to refactor to use a common RunAction component 

Thanks for the feedback.  Makes sense to me, will make this change along w/ removing the modal confirmation.

packages/components/src/components/RunAction/RunAction.js
```diff
@@ -0,0 +1,139 @@
...
+        });
+        showNotification({
+          message: triggerMsg,
+          kind: 'success',
+          logsURL: finalURL
+        });
+      })
+      .catch(error => {
+        let msg = null;
+        switch (action) {
```
lilley2412, line 63: @AlanGreene so .. I'm actually _not_ a frontend dev, but trying to learn as I go w/ this one.  Is this the best way to manage the intl id's?  It appears they must be static.  If there is a better pattern / example for me to replicate I'm happy to change if you can point me in the right direction.

AlanGreene: That's right, they must be static. This approach is fine, if we end up with many more actions in future we might consider refactoring it but it's not a problem for now. 👍

lilley2412: ok sounds good, I will leave it this way for now

lilley2412:
@AlanGreene hopefully last issue ... this unit test fails https://github.com/tektoncd/dashboard/blob/93dfbbfae5cc65a27352ca7dc428a196aedff53c/src/containers/PipelineRuns/PipelineRuns.test.js#L404, I've tried numerous things to get it working (changing the data store input, dumping out the matched rerun button which looks ok, changing the pipelineRuns.js code to always enable Rerun just to get it to work).  First I suspected it was related to the renaming of rerun to runaction, but that doesn't appear to need any additional config in the routes.

This is the error:
```
    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 1
    Received number of calls: 0

      427 |     await waitFor(() => getByText(/Rerun/i));
      428 |     fireEvent.click(getByText('Rerun'));
    > 429 |     expect(PipelineRunsAPI.rerunPipelineRun).toHaveBeenCalledTimes(1);
```

I'm sure I'm missing something really simple here, would appreciate any tips!  Once I understand this I'll add new tests for the start button.

AlanGreene:
With your update the rerun menu item is only enabled if the run is not currently pending (makes sense 👍), it was previously always enabled. The way the `isPending` util is written it expects to find a PipelineRun status condition with `type === 'Succeeded'` but the test data isn't providing that.

If you add something like this it should pass:
```diff
       'pipelineRunWithSingleLabel-id': {
         status: {
-          startTime: '1'
+          startTime: '1',
+          conditions: [
+            { reason: 'Completed', status: 'True', type: 'Succeeded' }
+          ]
```

lilley2412:
> If you add something like this it should pass:
>  ...

Ok so I _thought_ I tried all variations of that but ... clearly I didn't b/c this suggestion fixed it :) Thanks.  Will work on some more tests then remove wip status.

AlanGreene:
Glad that worked, it can be tricky to get things _just right_. Don't hesitate to reach out here or Slack if you have any other questions.

packages/components/src/components/RunAction/RunAction.js
```diff
@@ -0,0 +1,139 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React, { Component } from 'react';
+import { injectIntl } from 'react-intl';
+import { Button } from 'carbon-components-react';
+import { Play32 as Play, Restart32 as Restart } from '@carbon/icons-react';
```
AlanGreene, line 17: Let's use PlayOutline32 here, this is more visually consistent with other icons already used and also consistent with use of some of these components in other applications.

packages/components/src/components/RunAction/RunAction.js
```diff
@@ -0,0 +1,139 @@
...
+    let triggerMsg = null;
+    switch (action) {
+      case 'rerun':
+        triggerMsg = intl.formatMessage({
+          id: 'dashboard.rerun.triggered',
+          defaultMessage: 'Triggered rerun'
+        });
+        break;
+      case 'start':
+        triggerMsg = intl.formatMessage({
```
AlanGreene, line 40: I'm not seeing this notification appear when clicking the Start button. Looks like there's an error being thrown from `runaction` and the error handling in turn is failing on a missing `error.response.status` for the error message.

src/api/pipelineRuns.js
```diff
@@ -121,3 +121,12 @@ export function rerunPipelineRun(pipelineRun) {
...
   return post(uri, payload).then(({ body }) => body);
 }
+
+export function startPipelineRun(pipelineRun) {
+  const { name, namespace } = pipelineRun.metadata;
+
+  const payload = [{ op: 'remove', path: '/spec/status' }];
+
+  const uri = getTektonAPI('pipelineruns', { name, namespace });
+  return patch(uri, payload).then(({ body }) => body);
```
AlanGreene, line 131: We've just recently (i.e. Tuesday) updated the API proxy in the backend and it's now correctly handling the request/response headers, in particular the `Accept` / `Content-Type` headers.

It looks like our original implementation of `patch` wasn't quite right. In api/comms.js `getPatchHeaders` you'll need to update the value of the `Accept` header to `application/json` to resolve the 406 error.

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -87,7 +89,9 @@ export /* istanbul ignore next */ function PipelineRunContainer(props) {
 
   const maximizedLogsContainer = useRef();
   const [loading, setLoading] = useState(true);
-  const [showRerunNotification, setShowRerunNotification] = useState(null);
+  const [showRunActionNotification, setshowRunActionNotification] = useState(
```
AlanGreene, line 92: nit: `setShowRunActionNotification`

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -174,6 +176,10 @@ export /* istanbul ignore next */ function PipelineRuns(props) {
     rerunPipelineRun(pipelineRun);
   }
 
+  function start(pipelineRun) {
```
AlanGreene, line 179: Looks like this isn't actually required since it's just passing the params directly to `startPipelineRun` so we could just call that directly. My guess is that we used to have more logic in the `rerunPipelineRun` and never removed the function after changing it.

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -195,7 +216,7 @@ export /* istanbul ignore next */ function PipelineRuns(props) {
         action: cancel,
         disable: resource => {
           const { reason, status } = getStatus(resource);
-          return !isRunning(reason, status);
+          return !isRunning(reason, status) && !isPending(reason, status);
```
AlanGreene, line 219: 👍

AlanGreene (changes requested):
Thanks for the PR. Overall it looks great, just a few minor issues below, and a small case of unfortunate timing with our rewrite of the backend code but easily fixed (left a comment below with details).

packages/components/src/components/RunAction/RunAction.js
lilley2412: It was due to no newRun / no need for the link to new run, as startPipelineRun doesn't return a new run.  I modified to make the link conditional:

https://github.com/tektoncd/dashboard/pull/2106/files#diff-95ec2ca7d82a45cf2f79f910de125bff1d3dfc12d870d5f261967ec47bd92e6aR52-R60

src/containers/PipelineRun/PipelineRun.js
lilley2412: 👍

src/containers/PipelineRuns/PipelineRuns.js
lilley2412: removed both methods

lilley2412:
Thanks for all the feedback @AlanGreene.  I think I have addressed everything; left comments on some of the resolved items.  Please lmk if you see any other issues.  Thanks!

AlanGreene (approved):
This looks great! Thanks again @lilley2412

/lgtm
/meow

Issue #2103, opened 2021-06-24
Display pod information to help users debug problems with their TaskRuns or PipelineRuns


AlanGreene:
# Is your feature request related to a problem ?

There are a number of classes of error where we don't provide relevant information crucial to debugging.

These include a number of problems at the pod level. Surfacing this information will make it much easier for users to debug in the Dashboard UI without having to switch to `kubectl` or other tools.

# Describe the solution you'd like

Display the equivalent of `kubectl get pod <podName> -n <namespace> -oyaml` in the UI to help the user with debugging issues at the pod level.

If possible, also include relevant `events`.

![image](https://user-images.githubusercontent.com/2829095/123273128-45080c80-d4fa-11eb-8490-1203a1cb043c.png)

# Describe alternatives you've considered

# Additional context

- UI components will need to be updated to support displaying optional pod information as a tab in the TaskRunDetails component
- RBAC will need to be updated to give the Dashboard ServiceAccount permissions to get and watch pods
- Current websocket connection probably isn't suitable as we likely don't want to subscribe to all updates for all pods. Ideally we would be able to just watch a single pod for updates when it's being viewed in the UI. @AlanGreene did some investigation into options for a new websocket approach ~1 year ago, might be worth revisiting for this.

AlanGreene:
UI components were updated to support displaying Pod and Event information in #2110 but this is not wired up yet.

I have a change ready to switch our websocket approach from a single central firehose to individual connections per kind, created and managed by each page as needed. Once https://github.com/tektoncd/dashboard/issues/1842 is complete I'll open the PR for review (needs to be rebased and a few small changes after https://github.com/tektoncd/dashboard/pull/2154 is merged).

This new approach will allow us to watch only the required pod for updates, as well as making it easier to add user-based auth in future.

PR #2113, opened 2021-07-03
Update proxy to support websockets

AlanGreene:
# Changes

Related to https://github.com/tektoncd/dashboard/issues/1842 and https://github.com/tektoncd/dashboard/issues/2103

Our current proxy does not support connection upgrades, which means we do not
have a clean way to access the built-in support for websockets provided by the
Kubernetes API server. Instead, the Dashboard provides a single websocket
connection which provides updates for all supported resources.

This has a number of drawbacks such as:
- not being able to selectively register to receive updates for:
  - resources in a specific namespace
  - a particular kind of resource
  - a single specific resource
- not supporting real time updates for resource-based extensions

This means that all pages in the Dashboard incur an unnecessary overhead by
processing updates for resources they may never require.

Add a new proxy implementation which handles the connection upgrades, while
also providing the ability to support more fine-grained permissions and
user-based auth which we want to add in future.

Migrate the existing endpoints to the new server/proxy setup, including the
existing websocket connection as it's still in use but will be replaced in the
near future.

Remove the service-based extension support as the webhooks-extension was
deprecated after v0.7.0 and has since been deleted. Similar functionality could
be added via a small modification to the resource-based extensions and using
the new proxy to access the services.

# Submitter Checklist

review them:

for more details._


briangleeson (approved):
/lgtm

No issues jumping out at me after looking over the changes :shipit:

PR #2119, opened 2021-07-08
Add autoscroll to logs of running steps

MA9521:
# Changes

A flag-activated feature to autoscroll the logs of a running step (running = was running at any point after the component mounting). The logs will auto-scroll down when there is new text added and the bottom of the "old" logs was already seen. This behaviour can be stopped at any time if the user scrolls up so that the bottom of the logs is out of view. This behaviour can be resumed if the user scrolls down so that the bottom of the logs is in view.

This feature is flag-activated not to to affect the behaviour of the Log pure-component users which may be using it as an element followed down the page by other elements or as a pure-component in a scrollable parent which is not the browser window.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/Log/Log.js
+++ b/packages/components/src/components/Log/Log.js
@@ -15,6 +15,11 @@ import React, { Component } from 'react';
 import { SkeletonText } from 'carbon-components-react';
 import { FixedSizeList as List } from 'react-window';
 import { injectIntl } from 'react-intl';
+import { getStepStatusReason, isRunning } from '@tektoncd/dashboard-utils';
+import {
+  hasElementPositiveVerticalScrollBottom,
+  isElementEndBelowViewBottom
+} from './domUtils';
 
 import Ansi from '../LogFormat';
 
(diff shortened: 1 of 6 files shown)


MA9521:
Jest is not a full-fledged browser so functions like getBoundingClientRect (which are the bread and butter of this feature) do not work in the test environment. I forked https://github.com/tektoncd/pipeline and changed 2 steps (https://github.com/MA9521/pipeline/commit/34f322e72bbdbe94cff7ded721efc1b86cf47cc3) so that they display long logs with an interval between new logs (so that you can see how the feature works).

Follow the instructions in this section but replace the repo url with `https://github.com/MA9521/pipeline`. The 2 steps in questions are  `read-docs-old` and `write-new-stuff` in the task `create-file` . The second task allows to verify that the feature works for steps with very long logs where the logs are rendered as a virtualized list (and contain a scroll inside the `<Log` component

packages/components/src/components/Log/Log.js
```diff
@@ -28,17 +33,92 @@ const itemSize = 15; // This should be kept in sync with the line-height in SCSS
...
+    rootElement.scrollTop = rootElement.scrollHeight - rootElement.clientHeight;
+  };
+
+  wasRunningAfterMounting = () => {
+    if (this.alreadyWasRunningAfterMounting) {
+      return true;
+    }
+    const { reason, status } = getStepStatusReason(this.props.stepStatus);
+    if (isRunning(reason, status)) {
+      this.alreadyWasRunningAfterMounting = true;
```
AlanGreene, line 116: this should be set in state

packages/components/src/components/Log/Log.test.js
```diff
@@ -67,6 +67,58 @@ describe('Log', () => {
...
+      .mockReturnValueOnce({ bottom: 0 });
+    const spiedFn = jest.spyOn(Element.prototype, 'scrollTop', 'set'); // the scrollTop value is changed in scrollToBottomLog
+    render(
+      <Log
+        stepStatus={{ running: true }}
+        fetchLogs={() => 'Line 1'}
+        enableLogAutoScroll
+      />
+    );
+    await waitFor(() => expect(spiedFn).toBeCalled());
```
AlanGreene, line 83: nit: `toHaveBeenCalled()` for consistency with existing tests

packages/components/src/components/Log/Log.test.js
```diff
@@ -67,6 +67,58 @@ describe('Log', () => {
...
+  });
+
+  it('auto-scrolls down, for a running step, when new logs are added and the Log component is scrollable', async () => {
+    const long = Array.from(
+      { length: 60000 },
```
AlanGreene, line 88: 20,001 or more is fine here, no need for massive logs in the test as long as we're triggering the right condition

AlanGreene (changes requested):
Thanks @MA9521, overall this looks good.

One thing I found a little jarring though was that when I click on a running TaskRun in the list, I'm immediately scrolled to the bottom of the page so if I'm trying to see the status / details I have to scroll back up. I would have expected it only to automatically scroll if the bottom was in view when I loaded it, or if I scrolled to the bottom myself.

packages/components/src/components/Log/Log.js
MA9521: As discussed, I added a comment in the amended commit explaining that this is a class variable because it does not require a subsequent re-rendering

packages/components/src/components/Log/Log.test.js
MA9521: changed as requested in the amended commit

packages/components/src/components/Log/Log.test.js
MA9521: changed as requested

MA9521:
In the amended commit, I replaced the condition `this.state.isLogBottomUnseen` by `this.state.isLogBottomUnseen === false` in the method `shouldAutoScroll` . In the method `handleLogScroll`, I added an initial condition `!this.state.loading` before setting the state. These 2 changes prevent the behaviour of clicking on a running step then being scrolled down without being able to see the step details at the top.
This means, that autoscroll now works at the conjunction of these 3 conditions:
 1) a step that was running at any moment after mounting
 2) the current bottom is below view
3) the previous state/render had
a) the bottom not below view and
b) the logs not loading

I should also mention that this does not solve the use case of a "sitting duck" user who has selected a not-yet-running step with "No logs available" displayed then the step starts running and the page is auto-scrolled down because 19900 lines (each with 35 to 45 characters) were added.
This auto-scroll down did not occur however for a sitting-duck user when the step went from to "No logs available" to display 20000 lines (each with a single character). So it seems that there is some loading behaviour when the virtualized list is used.

I updated the unit tests to force a re-rendering with a change in the number of log lines as `state.loading=true` prevents updating the state variable `isLogBottomUnseen`

AlanGreene (approved):
Looks good, thanks for making those changes 👍 The 'sitting-duck' use case you described is fine and for the most part would be expected.

/lgtm

Issue #2049, opened 2021-05-21
Skip to main content link navigates to different page


AlanGreene:
# Describe the bug

Using the keyboard, tab from the browser's address bar to the page content and a 'Skip to main content' link should be displayed. Click this link and the browser is navigated to the PipelineRuns page.

![image](https://user-images.githubusercontent.com/2829095/119158122-ca0c8b80-ba4d-11eb-88fb-a38abaedb533.png)

# Expected behaviour

Clicking the link should focus the main content of the current page, bypassing the navigation.

# Steps to reproduce the bug

- navigate to any page on the Dashboard, e.g. https://dashboard.robocat.tekton.dev/#/about
- press the tab key until the 'Skip to main content' link appears
- press enter to activate the link
- see the page is redirected to the PipelineRuns list

# Environment details

- Versions:
   - Tekton Dashboard: `main`

# Additional Info

The target of the link is `#main-content` which references the id of the element wrapping the main page content. Normally this would be OK and clicking the link would move focus to the referenced element, but for the Dashboard we're using the fragment to identify the current page. Since `main-content` isn't a recognised page in the app, we end up redirecting to the default page which is PipelineRuns.

We need to find a different strategy for this skip link, some suggestions to consider:
- `element.scrollIntoView()`
- `element.focus()`
- …

We cannot switch from HashRouter (e.g. `/#/about`) to BrowserRouter (e.g. `/about`) as the latter will not work in all cases depending on how the Dashboard is accessed.

PR #2127, opened 2021-07-12
update the skip to main content button

LyndseyBu:
# Changes
Fixes https://github.com/tektoncd/dashboard/issues/2049
Updating the Skip to main content link so that it skips the navigation, and sets the focus on the main content

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/Header/Header.js
+++ b/packages/components/src/components/Header/Header.js
@@ -22,6 +22,12 @@ import {
   SkipToContent
 } from 'carbon-components-react';
 
+function skipToContentClick(event) {
+  event.preventDefault();
+  window.scrollTo(0, 0);
+  document.getElementById('main-content').focus();
+}
+
 function Header({
   intl,
   isSideNavExpanded,
@@ -30,7 +36,7 @@ function Header({
 }) {
   return (
     <CarbonHeader aria-label="Tekton Dashboard" className="tkn--header">
-      <SkipToContent>
+      <SkipToContent href="#" onClick={skipToContentClick}>
         {intl.formatMessage({
           id: 'dashboard.skipToContent',
           defaultMessage: 'Skip to main content'
--- a/src/scss/App.scss
+++ b/src/scss/App.scss
@@ -116,3 +116,8 @@ a {
     min-height: 100%;
   }
 }
+
+//height of the main header, added for accessibility of skip to main content link
+#main-content {
+  scroll-margin-top: 3rem;
+}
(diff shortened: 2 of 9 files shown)


packages/components/src/components/Header/Header.js
```diff
@@ -28,14 +27,20 @@ function Header({
...
+  function skipToContentClick(event) {
+    event.preventDefault();
+    window.scrollTo(0, 0);
+    document.getElementById('main-content').focus();
+  };
+
   return (
     <CarbonHeader aria-label="Tekton Dashboard" className="tkn--header">
-      <SkipToContent>
+      <a href="#" className="bx--skip-to-content" onClick={skipToContentClick}>
```
AlanGreene, line 38: Can we continue to use the `SkipToContent` component here so we don't have to rely on knowing the classname etc.? Looks like it allow overriding the `href` and adding additional props such as the `onClick`: https://github.com/carbon-design-system/carbon/blob/e7ef150ba1d74e2e864a9d6d67e612e824dcbf51/packages/react/src/components/UIShell/SkipToContent.js#L18

packages/components/src/components/Header/Header.js
```diff
@@ -28,14 +27,20 @@ function Header({
   logoutButton,
   onHeaderMenuButtonClick
 }) {
+  function skipToContentClick(event) {
```
AlanGreene, line 30: This can be moved above, outside the component function since it doesn't rely on any state or props.

src/scss/App.scss
```diff
@@ -116,3 +116,7 @@ a {
     min-height: 100%;
   }
 }
+
+#main-content {
+  scroll-margin-top: 3rem;
```
AlanGreene, line 121: Can you add a comment noting where the value comes from? If we change the header height or other aspects of the layout future it could save some searching and avoid weirdness.

AlanGreene (changes requested):
Thanks @LyndseyBu, this looks good. Just a few small comments below. Let me know if you have any questions.

packages/components/src/components/Header/Header.js
```diff
@@ -28,14 +27,20 @@ function Header({
...
+    event.preventDefault();
+    window.scrollTo(0, 0);
+    document.getElementById('main-content').focus();
+  };
```
AlanGreene, line 34: the failing check is due to a linting error here:

```
/home/prow/go/src/github.com/tektoncd/dashboard/packages/components/src/components/Header/Header.js
  34:4  error  Delete `;`  prettier/prettier
```

src/scss/App.scss
LyndseyBu: comment added

AlanGreene (approved):
Thanks for fixing this @LyndseyBu 

/lgtm

Issue #2128, opened 2021-07-14
Readonly dashboard still has delete button


rouke-broersma:
# Describe the bug

When we deploy the tekton dashboard as readonly, TaskRuns have a delete button when you select one or more items. This delete button actually deletes the TaskRuns.

# Expected behaviour

In readonly mode nothing should be modifyable using the dashboard.

# Environment details

<html>
<body>
[...]

PR #2129, opened 2021-07-14
Fix UI bug for TaskRuns in read-only mode

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/2128

Batch actions should not be enabled in read-only mode. A bug introduced
during recent refactoring meant that the read-only flag wasn't being read
correctly in this case causing the checkboxes to appear on the TaskRuns
table. This gave the impression that TaskRuns could be deleted, however
the action is still blocked in environments where RBAC is enforced as
the Dashboard ServiceAccount does not have permissions to delete these
resources when deployed in read-only mode.

Update the test to prevent a similar regression in future, as well as
updating the equivalent test for PipelineRuns in the same manner.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/PipelineRuns/PipelineRuns.test.js
+++ b/src/containers/PipelineRuns/PipelineRuns.test.js
@@ -399,7 +399,12 @@ describe('PipelineRuns container', () => {
       url: '/pipelineruns'
     };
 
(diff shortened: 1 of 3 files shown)


briangleeson (approved):
/lgtm

PR #2130, opened 2021-07-14
Move Extensions to use react-query instead of redux

briangleeson:
Remove the unused Extensions-related redux code.

# Changes

Move Extensions to use react-query instead of redux. Remove the unused Extensions-related redux code.

# Submitter Checklist

review them:

for more details._


src/containers/App/App.js
```diff
@@ -184,10 +186,17 @@ export function App({
...
 
   const showLoadingState = isPropertiesPlaceholder || isMessagesPlaceholder;
-  const isFetchingConfig = isFetchingProperties || isFetchingMessages;
-  const loadingConfigError = propertiesError || messagesError;
+  const isFetchingConfig =
+    isFetchingProperties || isFetchingMessages || isFetchingExtensions;
+  const loadingConfigError =
+    propertiesError || messagesError || extensionsError;
+
+  function fetchData() {
```
briangleeson, line 194: @AlanGreene wasnt 100% sure if this it the correct approach here for fetching the data via websocket

src/reducers/extensions.js
```diff
@@ -1,78 +0,0 @@
...
-import keyBy from 'lodash.keyby';
-
-import {
-  createErrorMessageReducer,
-  createIsFetchingReducer
-} from './reducerCreators';
-
-const type = 'Extension';
-
-export function mapResourceExtension(extension) {
```
briangleeson, line 24: @AlanGreene is it safe to remove the `mapResourceExtension` and `byName` functions in this reducer?

briangleeson:
There are some failures in the SideNav tests locally. Will need to figure out how to fix those

src/containers/App/App.js
AlanGreene: yes this is good for now, after all resources have been fully migrated to react-query we'll be centralising this using [`invalidateQueries`](https://react-query.tanstack.com/reference/QueryClient#queryclientinvalidatequeries) and letting react-query take care of refetching all active queries automatically.

src/containers/App/App.js
```diff
@@ -184,10 +186,17 @@ export function App({
   });
 
   const showLoadingState = isPropertiesPlaceholder || isMessagesPlaceholder;
-  const isFetchingConfig = isFetchingProperties || isFetchingMessages;
-  const loadingConfigError = propertiesError || messagesError;
+  const isFetchingConfig =
+    isFetchingProperties || isFetchingMessages || isFetchingExtensions;
```
AlanGreene, line 190: no need to include extensions in this, it's only for holding off on setting global config such as the selected namespace which may depend on deployment settings (e.g. single namespace mode)

src/containers/SideNav/SideNav.js
```diff
@@ -52,6 +63,10 @@ function SideNav(props) {
 
   const { namespace } = match?.params || {};
 
+  const { data: extensions = [], refetch } = useExtensions();
+
+  useWebSocketReconnected(refetch, webSocketConnected);
```
AlanGreene, line 68: no need for this here since it's already handled in App

src/reducers/extensions.js
AlanGreene: Once the logic in src/containers/Extensions/Extension.js that was relying on the `extensionType` is updated it should be safe to remove 👍

src/containers/SideNav/SideNav.js
briangleeson: Ok, so `extensions` should be passed in as a prop to SideNav from App?

AlanGreene: sorry just meant the websocket piece, they're already being refetched on reconnect in App, so we can just call the `useExtensions` hook here and will benefit from the caching etc. already being done by react-query

src/api/utils.js
```diff
@@ -186,8 +186,8 @@ export function useWebSocket({ kind: _ }) {
   }, [queryClient, webSocket]);
 }
 
-export function useCollection(kind, api, params) {
-  const query = useQuery([kind, params], () => api(params));
+export function useCollection(kind, api, params, queryConfig) {
+  const query = useQuery([kind, params], () => api(params, queryConfig));
```
AlanGreene, line 190:
```suggestion
  const query = useQuery([kind, params], () => api(params), queryConfig);
```

`queryConfig` should be passed as the third param to `useQuery`

src/containers/App/App.js
```diff
@@ -185,9 +182,25 @@ export function App({
...
+    data: extensions = [],
+    error: extensionsError,
+    refetch: refetchExtensions
+  } = useExtensions(
+    { namespace: tenantNamespace || ALL_NAMESPACES },
+    { enabled: !isFetchingConfig }
+  );
+
+  const loadingConfigError =
+    propertiesError || messagesError || extensionsError;
```
AlanGreene, line 196: no need to include extensions here as they don't block / affect the rest of the app, if we fail to load them we just won't show the nav items for them but everything else in the app works as expected.

src/containers/Extensions/Extensions.js
```diff
@@ -19,28 +19,31 @@ import {
...
   useTitleSync({
     page: intl.formatMessage({
       id: 'dashboard.extensions.title',
       defaultMessage: 'Extensions'
     })
   });
 
+  const { data: extensions = [], error, isFetching, refetch } = useExtensions();
+
+  useWebSocketReconnected(refetch, webSocketConnected);
```
AlanGreene, line 45: no need for the refetch here since it's handled centrally in App.js `fetchData`

AlanGreene (changes requested):
Thanks @briangleeson. Just a few small changes then it's ready to go.

AlanGreene (approved):
Looks good, thanks @briangleeson 
/meow

PR #2133, opened 2021-07-15
Add scroll buttons to logs of steps

MA9521:
# Changes

Related to https://github.com/tektoncd/dashboard/issues/327

A flag-activated feature to add scroll buttons when the bottom (or the top) of the logs is out of view. The buttons will appear inside the right padding of the Log component. Clicking on the scroll to bottom ( or top) button) scrolls you automatically to the bottom (or top) of the logs. This feature also accounts for presence of page headers (up to a height of 3.125 rem).

This feature is flag-activated not to to affect the behaviour of the Log pure-component users which may be using it as an element followed down the page by other elements or as a pure-component in a scrollable parent which is not the browser window.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/Log/Log.js
+++ b/packages/components/src/components/Log/Log.js
@@ -12,13 +12,16 @@ limitations under the License.
 */
 
 import React, { Component } from 'react';
-import { SkeletonText } from 'carbon-components-react';
+import { Button, SkeletonText } from 'carbon-components-react';
 import { FixedSizeList as List } from 'react-window';
 import { injectIntl } from 'react-intl';
 import { getStepStatusReason, isRunning } from '@tektoncd/dashboard-utils';
+import { DownToBottom16, UpToTop16 } from '@carbon/icons-react';
 import {
   hasElementPositiveVerticalScrollBottom,
-  isElementEndBelowViewBottom
+  hasElementPositiveVerticalScrollTop,
+  isElementEndBelowViewBottom,
+  isElementStartAboveViewTop
 } from './domUtils';
 
 import Ansi from '../LogFormat';
@@ -42,7 +45,7 @@ export class LogContainer extends Component {
 
   componentDidMount() {
     this.loadLog();
-    if (this.props.enableLogAutoScroll) {
+    if (this.props.enableLogAutoScroll || this.props.enableLogScrollButtons) {
       this.wasRunningAfterMounting();
       window.addEventListener('scroll', this.handleLogScroll, true);
       this.handleLogScroll();
@@ -51,7 +54,7 @@ export class LogContainer extends Component {
 
   componentDidUpdate(prevProps, prevState) {
     if (
-      this.props.enableLogAutoScroll &&
+      (this.props.enableLogAutoScroll || this.props.enableLogScrollButtons) &&
       (prevState.logs?.length !== this.state.logs?.length ||
         prevProps.isLogsMaximized !== this.props.isLogsMaximized)
     ) {
@@ -72,10 +75,18 @@ export class LogContainer extends Component {
   handleLogScroll = () => {
     if (!this.state.loading) {
       const isLogBottomUnseen = this.isLogBottomUnseen();
+      const isLogTopUnseen =
+        this.props.enableLogScrollButtons && this.isLogTopUnseen();
+      this.updateScrollButtonCoordinates();
 
-      if (isLogBottomUnseen !== this.state.isLogBottomUnseen) {
+      if (
+        isLogBottomUnseen !== this.state.isLogBottomUnseen ||
+        (this.props.enableLogScrollButtons &&
+          isLogTopUnseen !== this.state.isLogTopUnseen)
+      ) {
         this.setState({
-          isLogBottomUnseen
+          isLogBottomUnseen,
+          isLogTopUnseen
         });
       }
     }
@@ -99,6 +110,15 @@ export class LogContainer extends Component {
     );
   };
 
+  isLogTopUnseen = () => {
+    return (
+      isElementStartAboveViewTop(this.logRef?.current) ||
+      hasElementPositiveVerticalScrollTop(
+        this.textRef?.current?.firstElementChild
+      )
+    );
+  };
+
   scrollToBottomLog = () => {
     const longTextElement = this.textRef?.current?.firstElementChild;
     if (longTextElement) {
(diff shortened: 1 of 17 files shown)


MA9521:
The PR now includes the unit tests and the messages files

packages/components/src/components/Log/Log.js
```diff
@@ -122,6 +150,105 @@ export class LogContainer extends Component {
...
+      const scrollButtonTop = Math.max(0, logRectangle.top);
+
+      const scrollButtonBottom = Math.max(
+        0,
+        document.documentElement?.clientHeight - logRectangle.bottom
+      );
+
+      this.updateCssStyleProperty(
+        logElementRight,
+        Object.keys({ logElementRight })[0]
```
AlanGreene, line 168: this will always yield the string `logElementRight`, can we use that directly here, or even use the CSS custom property name instead so we don't have to transform between formats later?

similar for the other values

packages/components/src/components/Log/Log.scss
```diff
@@ -53,6 +53,31 @@ pre.tkn--log {
...
+    width: 1.6rem;
+    height: 1.6rem;
+    min-height: 1.6rem; //to override Carbon's .bx--btn
+    right: var(--log-element-right);
+  }
+
+  .button-container {
+    position: absolute;
+    clip-path: inset(0); // ensure the children with position:fixed are not shown outside this element.
+    top: 3.125rem; //equals the maximum between padding-top of pre.tkn--log and betwwen the page header height
```
AlanGreene, line 67:
```suggestion
    top: 3.125rem; //equals the maximum between padding-top of pre.tkn--log and the page header height
```

packages/components/src/components/Log/Log.scss
```diff
@@ -53,6 +53,31 @@ pre.tkn--log {
...
+    bottom: 0;
+    right: 0;
+    width: 1.6rem; //equals the padding-right of pre.tkn--log
+  }
+  
+  .scroll-to-top-btn {
+    top: calc(var(--scroll-button-top) + 3.125rem); //3.125 rem is the maximum between padding-top of pre.tkn--log and betwwen the page header height
```
AlanGreene, line 74:
```suggestion
    top: calc(var(--scroll-button-top) + 3.125rem); //3.125 rem is the maximum between padding-top of pre.tkn--log and the page header height
```

packages/components/src/components/Log/Log.scss
```diff
@@ -53,6 +53,31 @@ pre.tkn--log {
     right: 0;
   }
 
+  .bx--copy-btn.scroll-btn {
+    position: fixed;
+    width: 1.6rem;
+    height: 1.6rem;
+    min-height: 1.6rem; //to override Carbon's .bx--btn
```
AlanGreene, line 60: Let's set `size="sm"` on the buttons instead of overriding the styles. It comes out at 2rem (32px).

If the hit box gets too small it becomes a usability issue so let's keep them a little bigger.

packages/components/src/components/Log/Log.scss
```diff
@@ -53,6 +53,31 @@ pre.tkn--log {
...
+    right: var(--log-element-right);
+  }
+
+  .button-container {
+    position: absolute;
+    clip-path: inset(0); // ensure the children with position:fixed are not shown outside this element.
```
AlanGreene, line 66: We need to resolve the issue with the Carbon tooltip fin still showing:

image

AlanGreene (changes requested):

packages/components/src/components/Log/Log.scss
MA9521: changed as requested

packages/components/src/components/Log/Log.scss
MA9521: changed as requested

packages/components/src/components/Log/Log.scss
MA9521: changed as requested

packages/components/src/components/Log/Log.js
MA9521: removed `Object.keys` and passed a literal string as requested... however ass css custom properties have to start by `--` and js variables cannot start by these 2 characters so I kept the variable in the js file in camelCase and kept the transformation to kebab-case.

packages/components/src/components/Log/Log.scss
MA9521: this fin is defined in `button::before`. I added in css `content: none;` as a pseudo-element to remove it but to do it needed to add an id to the `Button` component as I couldn't define `content: none;` with classes ( Carbon's css has many classes in the selector so it's practically impossible to  have a higher specificity specificity without an id and with classes/elements). The alternative would be to have in-line styling.

To have a truly unique id ( to account for the cases of several `Log` components simultaneously in the dom), we would need to pass the pipelineRunId, stepName and taskRun as props to the `Log` component to generate a unique id. But already the way I have defined the css custom properties as global properties assumes there is only one `Log` component at given moment in the dom.

So, if we want to account for multiple `Log` components simultaneously in the dom, I need to make these custom properties as element properties instead of global ( or use in-line styling) and pass the pipelineRunId, stepName and taskRun as props to the `Log` component ( to be able to remove the fin).

I believe the assumption of one `Log` component at given moment in dom is a safe assumption given that this feature is already flag-enabled.

AlanGreene: > I believe the assumption of one Log component at given moment in dom is a safe assumption given that this feature is already flag-enabled.

Yes this is safe. The flag defaults to false, and it's not something we'll be documenting as part of the public API of the component due to the specific requirements that must be met to have this (and the auto-scroll) work correctly. 👍

packages/components/src/components/Log/Log.js
AlanGreene: While it's true that variable names cannot start with `-`, field names can. There should be no issue with something like `--log-element-right` as a field of the parent object as you're currently setting it.

image

MA9521: changed as requested

AlanGreene (approved):
Looks great. Work for both windowed and non-windowed logs, regardless of the size of the log container (full screen, normal, taller/shorter than viewport etc.). Thanks for sticking with this @MA9521 and making it work for all the variations needed by the Dashboard application. Thanks also for your previous change adding the auto-scroll feature. Both of these are very useful and welcome additions.

/lgtm
/meow

PR #2137, opened 2021-07-19
Remove duplicate requests for Extensions

AlanGreene:
# Changes

Ensure all calls to `useExtensions` are using a consistent query key
by passing the correct value for `namespace` in the params.

Without this change the keys are as follows:
- `['Extensions', null]`
- `['Extensions', { namespace: 'some_namespace' }]`

meaning we have 2 separate requests for Extensions being made.

Setting the namespace param eliminates the additional request with
`null` in the key and ensure that if the tenant namespace is changed
we will invalidate the cache and refetch with the correct namespace.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/Extensions/Extensions.js
+++ b/src/containers/Extensions/Extensions.js
@@ -18,10 +18,15 @@ import {
   Link as CarbonLink,
   InlineNotification
 } from 'carbon-components-react';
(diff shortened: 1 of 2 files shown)


briangleeson (approved):
/lgtm

PR #2140, opened 2021-07-22
Update log trailer when forcePolling is true but step is complete

AlanGreene:
# Changes

Update the message displayed in the log trailer when the step is complete
but the consumer of the log component has set the `forcePolling` prop. This
is to avoid confusion with the existing 'Step Completed' or 'Step Failed'
trailers as they suggest to the user that the logs are displayed in full
which may not be the case while `forcePolling` is still set.

Add a new string to make it clear that additional logs may yet be displayed
even though the step is complete.

image

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/Log/Log.js
+++ b/packages/components/src/components/Log/Log.js
@@ -156,7 +156,14 @@ export class LogContainer extends Component {
   };
 
   getTrailerMessage = trailer => {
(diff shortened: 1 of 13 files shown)


briangleeson (approved):
/lgtm

Issue #2114, opened 2021-07-07
Allow creation of Pending PipelineRuns


AlanGreene:
# Is your feature request related to a problem ?

Allow creation of a PipelineRun in Pending state via the Dashboard UI

https://tekton.dev/docs/pipelines/pipelineruns/#pending-pipelineruns

# Describe the solution you'd like

In the 'Optional values' fieldset of the Create PipelineRun page we should provide an option for the user to create the PipelineRun in Pending state.

It should be presented as a toggle with appropriate labels, and default to disabled so we maintain existing behaviour and create a PipelineRun in normal state so it will begin executing without requiring additional intervention.

# Describe alternatives you've considered

# Additional context

taechonkim:
@AlanGreene Hello 🙂 I'm interested in work on this feature. 
I understood the requirement of the feature from https://tekton.dev/docs/pipelines/pipelineruns/#pending-pipelineruns
I think that it should added checkbox in the 'Optional values' fieldset to create Pending PipelineRuns for satisfying the feature.
And I think it should provides the feature through UI like button to start Pending PipelineRuns in PipelineRun.js as far as I understood.
I'll appreciate let me know I'm on the right track. Thank you :)

AlanGreene:
Thanks for your interest @taechonkim, however @LyndseyBu is already working on this feature.

taechonkim:
@AlanGreene Oh, I just checked https://github.com/tektoncd/dashboard/issues/2069 :( Sorry. 
I think I only need to implement the ability to create Pending PipelineRuns, right?

Thanks for your quick reply :)

AlanGreene:
As I mentioned, @LyndseyBu is working on this feature to add the option to Create PipelineRun in pending state. #2069 is a different feature and has already been completed.

taechonkim:
@AlanGreene Sorry to bother you 😭 I'll find another issue to work on. Thank you!

AlanGreene:
No problem at all, feel free to comment on another issue / feature that you're interested in 😸

PR #2149, opened 2021-07-27
create pending pipeline

LyndseyBu:
# Changes
as described in : https://github.com/tektoncd/dashboard/issues/2114

Allow for the creation of a pending PipelineRun
image
image

# Submitter Checklist

review them:

for more details._


LyndseyBu:
@AlanGreene ,  unsure about some of the strings here, and do we need to translate labels and aria-labels?

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -57,7 +58,8 @@ const initialState = {
   submitError: '',
   timeout: '60',
   validationError: false,
-  validTimeout: true
+  validTimeout: true,
+  pendingPipelineStatus: ''
```
AlanGreene, line 62: nit: put this above `pipelineError` so they're in alphabetical order, makes it easier to find things in the list and avoid duplicates etc.

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -105,7 +107,8 @@ function CreatePipelineRun(props) {
       submitError,
       timeout,
       validationError,
-      validTimeout
+      validTimeout,
+      pipelinePendingStatus
```
AlanGreene, line 111: nit: same as above, list these alphabetically

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -667,6 +682,18 @@ function CreatePipelineRun(props) {
...
             }
           />
+          <Toggle
+            defaultToggled={false}
+            id="pending-pipeline-toggle"
+            labelText={intl.formatMessage({
+              id: 'dashboard.createPipelineRun.pendingPipilinRun',
+              defaultMessage: 'Create PipelineRun in pending state'
+            })}
+            onChange={handleToggleChange}
```
AlanGreene, line 692: `onToggle` seems to give a cleaner interface here, providing the value of `checked` as the first param so we don't have to deal with the event object at all.

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -667,6 +682,18 @@ function CreatePipelineRun(props) {
               setState(state => ({ ...state, timeout: value }))
             }
           />
+          <Toggle
+            defaultToggled={false}
+            id="pending-pipeline-toggle"
+            labelText={intl.formatMessage({
+              id: 'dashboard.createPipelineRun.pendingPipilinRun',
```
AlanGreene, line 689: typo: should be `.pendingPipelineRun`, we could actually shorten this to just `.pending` since it's already scoped to the Create PipelineRun page, or maybe `.status.pending` for future expansion.

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -667,6 +682,18 @@ function CreatePipelineRun(props) {
...
+              defaultMessage: 'Create PipelineRun in pending state'
+            })}
+            onChange={handleToggleChange}
+            labelA="disabled"
+            labelB="enabled"
+            aria-label="Pending PipelineRun state"
```
AlanGreene, line 695: These should all be translated, and let's go with 'Enabled' and 'Disabled' for consistency with other labels in the UI. I spotted a few places where our aria-labels aren't currently translated but we can do a separate scrub of those.

Do we actually need to provide an `aria-label` here or is `labelText` sufficient?

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -134,6 +137,17 @@ function CreatePipelineRun(props) {
...
+  function handleToggleChange(event) {
+    const isPending = event.target.checked;
+    if (isPending) {
+      setState(state => ({
+        ...state,
+        pipelinePendingStatus: 'PipelineRunPending'
+      }));
+    } else {
+      setState(state => ({ ...state, pipelinePendingStatus: '' }));
+    }
```
AlanGreene, line 149: we can flip this inside out since the only piece changing is the status value:

```js
setState(state => ({
  ...state,
  pipelinePendingStatus: isPending ? 'PipelineRunPending' : ''
}));
```

AlanGreene (changes requested):
Looks good, thanks @LyndseyBu. Just a few minor nits and suggestions. Also answered your question about translations.

src/containers/CreatePipelineRun/CreatePipelineRun.js
LyndseyBu: I removed the `aria-label `, I was unsure if both we needed, some of the carbon examples have both and others just have labeText, so we I guess we're fine to go to with that

AlanGreene (approved):
Looks good @LyndseyBu, thanks. Just needs a rebase then it's good to go.

PR #2153, opened 2021-07-29
Store namespace resources for consistency with other resource types

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/1842

Instead of using react-query's `select` feature in the `useNamespace` hook
to return just the names of the namespaces, return the full list of resources
as we do with other types. This ensures the same logic for handling websocket
events will apply to namespaces and avoids needing to add a special case to
handle this one resource type.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/main/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/main/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/main/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/dashboard/blob/main/CONTRIBUTING.md)
for more details._

PR #2154, opened 2021-07-30
Remove remaining use of redux

AlanGreene:
# Changes

~~Depends on https://github.com/tektoncd/dashboard/pull/2153~~ merged
Closes https://github.com/tektoncd/dashboard/issues/1842

Remove the `notifications` reducer which was used to track the websocket
connection state, and update the code to allow for detecting websocket
reconnections on a per-query basis.

Remove the use of `useWebSocketReconnected` from all containers except
`App` so instead of individually refetching each query we invalidate them
all centrally and let react-query take care of refetching any queries that
are currently active (i.e. in use on the current page). Any inactive queries
will only be refetched on navigation to pages that require them.

A future update will make additional changes to the websocket connection,
removing the need for the central websocket with events for all Tekton
resources, instead moving to a per-kind connection so we can significantly
reduce the processing overhead required by the client, especially on pages
that only deal with a small subset of resource kinds (i.e. most pages).

# Submitter Checklist

review them:

for more details._


AlanGreene:
npm install didn't complete successfully, re-running unit tests

briangleeson (approved):
/lgtm

PR #2167, opened 2021-08-11
Further improvements to websocket connections

AlanGreene:
# Changes

When making a websocket connection pass the `resourceVersion` of
the previously retrieved list so we only start watching for updates
from that point rather than receiving duplicate websocket events
for all of the resources retrieved via the initial GET request.

To achieve this, the API functions must return the API response
fully intact, including the `metadata` object containing the
`resourceVersion` for the returned list result. The `useCollection`
hook will then transform the returned data for the query to match
the structure expected by consumers (i.e. it will return `data = response.items`)

Also ensure we don't have duplicate websocket connections for
Namespaces and Extensions. The App component will be responsible
for subscribing to these, and any individual page that requires
them will receive updates via the query without needing their
own websocket connection.

These changes further reduce the number of requests made by the
Dashboard client as well as reducing the processing overhead in browser.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/api/clusterInterceptors.js
+++ b/src/api/clusterInterceptors.js
@@ -13,7 +13,6 @@ limitations under the License.
 
 import { get } from './comms';
 import {
-  checkData,
   getQueryParams,
   getTektonAPI,
   triggersAPIGroup,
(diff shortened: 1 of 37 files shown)


briangleeson:
Cheers for the walkthrough

/lgtm

PR #2174, opened 2021-08-18
Update dependencies and remove remaining use of webpack 4

AlanGreene:
# Changes

- Switch Storybook to webpack 5 for both manager and preview (user code) builds
- Switch visx packages used for experimental graph components to the new `@visx` scope
- Fix issue with table batch selection string under new intl versions
- Clean up dependencies to remove duplicates and unnecessary packages (mostly facilitated by the Storybook and visx changes)
- Run `lint:fix` with the updated rules

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/main/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/main/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/main/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/dashboard/blob/main/CONTRIBUTING.md)
for more details._

PR #2177, opened 2021-08-20
Update strings for table batch selection

AlanGreene:
# Changes

Extracted from https://github.com/tektoncd/dashboard/pull/2174

In a recent update Carbon changed the strings for the number
of items selected on a table. Previously 0 and 1 were handled
by a single string, and all other numbers by a second string.
Now '1 item selected' is a single string, and 0 or >1 are handled
by the other.

'0 items selected' should never really be seen, but could potentially
happen in case of an unexpected error so we still need to account
for it so the string is displayed correctly.

Update the translations to reflect this change.

# Submitter Checklist

review them:

for more details._


briangleeson (approved):
/lgtm

Though it may be worth having the tweaked translations reviewed, especially for the Asian languages

PR #2178, opened 2021-08-20
Make use of ingress in local dev cluster opt-in

AlanGreene:
# Changes

Update the behaviour of the `prepare-kind-cluster create` command
so it no longer installs and configures an ingress. This reduces
the resources required for the default dev environment as we no
longer need to install and configure the nginx ingress controller.

It also avoids issues where ports 80 or 443 are already bound to
other services on the host machine.

Add a new command `prepare-kind-cluster create-with-ingress` which
will behave in the same way as the previous `create` command,
allowing devs to opt-in to this behaviour if they wish.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -136,13 +136,19 @@ e.g. to connect your local dev frontend to the Dashboard deployed on the robocat
 
 The Dashboard repository contains a script that will provision a [`kind` cluster](https://kind.sigs.k8s.io/) named 'tekton-dashboard' with the latest releases of Tekton Pipelines and Tekton Triggers installed, as well as a version of the Tekton Dashboard which can be customised by providing additional parameters matching those expected by the installer script described earlier.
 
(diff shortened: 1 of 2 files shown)


docs/dev/README.md
```diff
@@ -136,14 +136,20 @@ e.g. to connect your local dev frontend to the Dashboard deployed on the robocat
...
 
 To delete the cluster:
 
 `./script/prepare-kind-cluster delete`
 
+To create a cluster with the Tekton Dashboard exposed via ingress at `tekton-dashboard.127.0.0.1.nip.io`:
+
+`./scripts/prepare-kind-cluster create-with-ingress`
+
+This ensures the cluster is create with the required node labels and port mappings (ports 80 and 443).
```
briangleeson, line 151: Typo: "... the cluster is create**d** ..."

briangleeson (changes requested):
One nit

docs/dev/README.md
AlanGreene: 🙈 fixed, thanks

briangleeson (approved):
/lgtm

PR #2192, opened 2021-09-03
Reduce front end bundle size

AlanGreene:
# Changes

Reduce the amount of code shipped to users' browsers by ensuring we're
only loading code that's actually required and making smart use of
our dependencies and support for tree shaking where available.

This change reduces the size of the vendor JS bundle by ~1MB (~50%)
by making two relatively small changes:
- disable the import of carbon-components-react for the custom UI
  extensions which are not currently supported (~500KB savings)
- switch the `react-syntax-highlighter` from using Prism to highlight.js,
  also ensure we're only loading the language definition required

As a result of the switch to loading the highlight.js language definition
from an esmodule, we also need to update the jest config for unit tests
to exclude the react-syntax-highlighter package from some of its processing
to avoid errors related to imports used outside of modules.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2019-2020 The Tekton Authors
+Copyright 2019-2021 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


briangleeson (approved):
/lgtm

PR #2209, opened 2021-09-15
Update route definitions to prepare for react-router update

AlanGreene:
# Changes

The next release of react-router (v6) makes some changes to the way
routes are defined and behave. The `render` and `component` props
should no longer be used in v5.1+ to prepare for this change, instead
favouring `children`.

Instead of passing through the `location` and `match` props from the
`Route`, we should instead retrieve them when / where needed via the
hooks provided by `react-router`, i.e. `useLocation`, `useParams`, etc.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/containers/Triggers/Triggers.js
+++ b/src/containers/Triggers/Triggers.js
@@ -12,7 +12,7 @@ limitations under the License.
 */
 
 import React from 'react';
-import { Link } from 'react-router-dom';
+import { Link, useLocation, useParams } from 'react-router-dom';
 import { injectIntl } from 'react-intl';
 import { getFilters, urls, useTitleSync } from '@tektoncd/dashboard-utils';
 import { FormattedDate, Table } from '@tektoncd/dashboard-components';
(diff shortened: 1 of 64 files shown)


src/containers/Triggers/Triggers.js
```diff
@@ -93,12 +94,7 @@ function Triggers(props) {
   }));
 
   return (
-    <ListPageLayout
-      {...props}
-      error={getError()}
-      filters={filters}
-      title="Triggers"
-    >
+    <ListPageLayout error={getError()} filters={filters} title="Triggers">
```
briangleeson, line 97: Nit: Shouldnt this title be translated?

briangleeson: I see similar titles for ListPageLayout elsewhere, so I'm guessing these are being treated as special keywords and thus not translated?

AlanGreene: exactly, we don't translate the CRD names, e.g. `PipelineRun`, `Trigger`, etc.

PR #2217, opened 2021-10-07
 Add margin to prevent "Clear all btn" from moving fix #2062

### LINKED ISSUE #2062: Clear All label filter incorrect layout when it overflow to the new line
# Describe the bug

I found this bug during investigate #2050. The problems found when we add labels until clear all button overflow to the next line. It will be shown like this:

Screen Shot 2564-05-25 at 21 13 59

But if we add more label, it will looks like:

Screen Shot 2564-05-25 at 21 18 15

We can see that the clear all button in the first picture is less margin than in the second picture.

# Expected behaviour

Expected the clear all button should be the same layout in the second picture.

# Steps to reproduce the bug

1. Open the PipelineRuns page.
2. Add label filter until the clear all button overflow the new line

# Environment details

Found on Tekton Dashboard branch main at commit `5b1372d4cb634e84142ed3abbe60a229a379964d`.

# Additional Info

yamaken1343:
# Changes

fix #2062 

The "filter" has  `.bx--tag` class and the "Clear All button" has `.bx--btn` class.
`.bx--tag` class has margin but `.bx--btn` class has not margin. So, there was a gap between when there was a "filter" on the left and when there was no "filter".

I add margin to `.bx--btn` class to prevent "Clear all button" from moving.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/LabelFilter/LabelFilter.scss
+++ b/packages/components/src/components/LabelFilter/LabelFilter.scss
@@ -29,6 +29,10 @@ limitations under the License.
       flex-shrink: 0;
       margin: $spacing-05 $spacing-03 0 0;
     }
+
+    .bx--btn {
+      margin-top: calc(#{$spacing-04} - 1px );
+    }
   }
 }
 


yamaken1343:
Tag height is 1.5rem and tag margin is 1rem. Btn height is 2rem. So btn margin should be 0.75rem correct.

But 0.75rem is 12px, not 11px.
So I'm subtracting 1px.

PR #2227, opened 2021-10-29
Fix Task component tests

AlanGreene:
# Changes

Some tests were failing silently due to bad structure, e.g. tests
not waiting for a promise to be resolved or not making any assertions.

Split the tests for the 'select default step' behaviour to assert
that the `onSelect` callback is called with the expected values for
task id and step id, and separately test that the `selectedStepId`
prop is processed correctly. This approach avoids having to use additional
test-specific wrapper components to handle the callback and provide
the updated prop value and keeps the purpose of the tests more clear.

Thanks to @LyndseyBu for catching this.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/packages/components/src/components/Task/Task.test.js
+++ b/packages/components/src/components/Task/Task.test.js
@@ -12,13 +12,14 @@ limitations under the License.
 */
 
 import React from 'react';
-import { fireEvent, waitFor } from '@testing-library/react';
+import { fireEvent } from '@testing-library/react';
 import Task from './Task';
 import { render } from '../../utils/test';
 
 const props = {
-  onSelect: () => {},
-  displayName: 'A Task'
+  displayName: 'A Task',
+  id: 'fake_task_id',
+  onSelect: () => {}
 };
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Task/Task.test.js
```diff
@@ -129,31 +145,40 @@ describe('Task', () => {
...
+    const stepName = 'a step';
+    const steps = [{ name: stepName }];
+    const { getByText } = render(
+      <Task {...props} expanded selectedStepId={stepName} steps={steps} />
     );
+
+    expect(
+      getByText(
+        (content, element) =>
+          content === 'a step' &&
```
briangleeson, line 157: Can use `content === stepName &&` here

briangleeson (commented):
/lgtm

One nit but looks good

packages/components/src/components/Task/Task.test.js
AlanGreene: good catch, updated

PR #2230, opened 2021-11-04
Update backend code for getting project version info

AlanGreene:
# Changes

Switch to using the new ConfigMaps introduced in Pipelines v0.25
and Triggers v0.15 that contain the version info. These ConfigMaps
are readable by all authenticated users. This means the Dashboard
no longer requires permissions to list or get deployments and can
instead use the ConfigMaps to retrieve the version information for
the About page and other uses.

Remove the project-specific RBAC for the deployments since it's
no longer required, and update the backend code to read the version
from the ConfigMaps instead.

This also means that we no longer support older releases of the
Pipelines and Triggers projects which did not provide these ConfigMaps.
Most functionality will still work, but there may be some errors in
the logs / UI.

Also add the Dashboard ConfigMap and related RBAC to the release
manifests so it's actually available on the cluster. These were added
some time ago but not actually included in the base kustomize.yaml.

# Submitter Checklist

review them:

for more details._


briangleeson (approved):
/lgtm

PR #2239, opened 2021-11-18
Surface details of step onError continue in lists and details

AlanGreene:
# Changes

Update the step and Task display in the UI to show indicators when
a non-zero exit code was ignored. Propagate the indicator all the
way up to the PipelineRuns and TaskRuns list pages, using the base
'success' icon.

When a step is selected, display the non-zero exit code in the
step details header so the user can easily identify why the status
is displayed differently. We have a similar message in the log trailer
but this may be below the fold depending on log length, window size, etc.

# Submitter Checklist

review them:

for more details._


briangleeson (approved):
/lgtm

PR #2241, opened 2021-11-23
Set content-type header only on valid request methods

triantium:
Avoid sending an empty Header on requests, that should not have a body.
Resolves https://github.com/tektoncd/dashboard/issues/2242

compare with https://developer.mozilla.org/de/docs/Web/HTTP/Methods
which method should and may have a body.

# Changes

Avoid sending an empty Header on requests, that should not have a body.
https://github.com/tektoncd/dashboard/issues/2242

compare with https://developer.mozilla.org/de/docs/Web/HTTP/Methods
which method should and may have a body.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/utils/proxy.go
+++ b/pkg/utils/proxy.go
@@ -17,7 +17,9 @@ func Proxy(request *http.Request, response http.ResponseWriter, url string, clie
 
 	req = req.WithContext(request.Context())
 
-	req.Header.Set("Content-Type", request.Header.Get("Content-Type"))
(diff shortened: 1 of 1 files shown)


AlanGreene (changes requested):
Thanks for the PR @triantium. This must also support setting the Content-Type header on PATCH requests which are used for cancelling a TaskRun/PipelineRun or starting a PipelineRun that was created in pending state.

triantium:
Missed the Patch method.

AlanGreene (approved):
/lgtm
/meow

Issue #2244, opened 2021-11-24
Update error page


AlanGreene:
Update and unify our current error handling to display more useful content to the user in case they hit a bad link, or try to access some feature or resource to which they do not have access.

This will be particularly useful when we make more progress on https://github.com/tektoncd/dashboard/issues/1420 and related issues.

kamilzal:
### **Final outcome**
Assets: **[robocat_404.svg.zip](https://github.com/tektoncd/dashboard/files/7660707/robocat_404.svg.zip)**


<img width="1312" alt="404_1" src="https://user-images.githubusercontent.com/35368089/144846077-f2f614b3-92f2-4187-adbe-79213e613361.png">
---

### **Illustration size @ 1312px width**

<img width="1312" alt="404_2" src="https://user-images.githubusercontent.com/35368089/144846092-d67d7798-4a88-4a63-9c01-110056eb6343.png">
---

### **Text and Spacing**

<img width="1312" alt="404_3" src="https://user-images.githubusercontent.com/35368089/144846095-5d35f77f-4e9f-4bb1-8134-ba3c9f1efcaf.png">
---

### **Alignment**

1. Align text to columns.
2. Align illustration to start at the middle of gutter between 2 columns.
3. Grouping of illustration + message should be centered inside the content area.

<img width="1312" alt="404_4" src="https://user-images.githubusercontent.com/35368089/144846099-4271e0a8-f40f-42a6-8d00-4102e3bac92f.png">

PR #2262, opened 2021-12-06
Update design of error page for page / resource not found

AlanGreene:
# Changes

Resolves https://github.com/tektoncd/dashboard/issues/2244

Add a new component which can be used as an error page in case the user
attempts to access a page / resource to which they do not have access or
which does not exist.

Use the robocat image instead of the Carbon error notification component,
and provide links to commonly accessed pages as suggestions for next steps
instead of leaving the user on an error page with no actions.

Enable this new design for users attempting to access 'Import Resources'
and 'Create' pages when the Dashboard is deployed in read-only mode, and
as the fallback URL when an unknown path is found.

This will be used more broadly in future with user permissions etc.

# Submitter Checklist

review them:

for more details._


src/scss/App.scss
```diff
@@ -52,6 +53,20 @@ limitations under the License.
...
+  height: 100%;
+}
+
+body {
+  // to prevent unwanted scroll when content doesn't fill full height
+  height: calc(100% - 3rem);
+}
+
+#main-content {
+  //height of the main header, added for accessibility of skip to main content link
```
briangleeson, line 66: super-nit: add a space between slashes and height

briangleeson (commented):
/lgtm

Just one microscopic nit

src/scss/App.scss
AlanGreene: updated, thanks

briangleeson (approved):
/lgtm

PR #2271, opened 2022-01-05
Rename string id

AlanGreene:
# Changes

The string is used for logs of steps which exited with non-zero exit code
that also have the `onError: continue` flag set in their definition. This
means that the non-zero exit code will not fail the step and should not be
considered an error. Update the id to reflect this. It may also be useful
in other scenarios in future, so keep the id generic.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/src/nls/messages_zh-Hans.json
+++ b/src/nls/messages_zh-Hans.json
@@ -183,7 +183,7 @@
     "dashboard.pipelineRun.notFound": "未找到 PipelineRun",
     "dashboard.pipelineRun.pipelineTaskName.retry": "{pipelineTaskName}（重试 {retryNumber, number}）",
     "dashboard.pipelineRun.stepCompleted": "步骤已完成",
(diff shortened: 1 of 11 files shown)


src/nls/messages_zh-Hans.json
```diff
@@ -183,7 +183,7 @@
     "dashboard.pipelineRun.notFound": "未找到 PipelineRun",
     "dashboard.pipelineRun.pipelineTaskName.retry": "{pipelineTaskName}（重试 {retryNumber, number}）",
     "dashboard.pipelineRun.stepCompleted": "步骤已完成",
-    "dashboard.pipelineRun.stepCompleted.withError": "步骤已完成，退出代码为 {exitCode}",
+    "dashboard.pipelineRun.stepCompleted.exitCode": "",
```
briangleeson, line 186: Should we keep this translation since the string itself hasnt changed?

AlanGreene: absolutely, forgot to restore it after the second rename. Thanks

briangleeson (approved):
/lgtm

PR #2282, opened 2022-01-13
Fix SideNav links when navigation from a cluster-scoped page

AlanGreene:
# Changes

When navigating from a cluster-scoped resource / list, e.g. ClusterTasks,
to any namespaced resource / list, the URL should include the namespace
of the currently selected global namespace filter. This will become
even more important once the namespace dropdown is moved to the header
in the very near future.

Use the selected namespace instead of the URL as the source of truth
when determining whether the target URL should include the namespace
as it will not be present in the current URL when viewing cluster scoped
resources.

# Submitter Checklist

review them:

for more details._


briangleeson (approved):
/lgtm

PR #2312, opened 2022-02-25
Avoid use of concatenation when producing CSS classnames

AlanGreene:
# Changes

Dynamically concatenating values to produce CSS classnames makes it
difficult to identify which values are / may be used, often leading
to stale styles being left behind as developers are unsure whether
they're safe to remove.

Where possible we should ensure the entire CSS classname appears
as a single string in the codebase so it's easily searchable. This
also becomes increasingly important when using tooling to identify
unused styles.

Some components such as `LogFormat` are exceptions to this rule
due to the potentially large number of values involved, and the
affected classnames already have an easily searchable prefix. These
prefixes can be excluded from the tooling to avoid them being pruned
as they depend on user content only available at runtime.

# Submitter Checklist

  functionality, content, code)

for more details._


briangleeson (approved):
/lgtm

Issue #2094, opened 2021-06-20
Add support for new graceful termination options from TEP-0058


AlanGreene:
Add support for the new [graceful termination](https://github.com/tektoncd/community/blob/0ccd5f907bde2452f76330bf5f7bf8a0a14e8bf2/teps/0058-graceful-pipeline-run-termination.md) feature added in Pipelines v0.25 (added as alpha) which provides three different ways to terminate a PipelineRun, differentiating between 'Stop' and 'Cancel'.

> - "StoppedRunFinally" - To stop (i.e. let the tasks complete, then execute finally tasks) a PipelineRun
> - "CancelledRunFinally" - To cancel (i.e. interrupt any executing non finally tasks, then execute finally tasks)
> - "Cancelled" - replaces today's "PipelineRunCancelled" - i.e. interrupt any executing tasks without running finally tasks
>
> Support for existing statuses has been left unchanged. The status "PipelineRunCancelled" is deprecated and replaced by "Cancelled" (it would be removed in v1). The new states are released as alpha API features.

There are some additional considerations here:
- feature detection - can we tell whether the feature is enabled or not? If not, should we wait for it to be promoted in a future Pipeline release?
- action name - the TEP differentiates between 'Stop' and 'Cancel'
  Using 'Cancel' as the action name was problematic in the past as it's the same as the option to dismiss the confirmation dialog, so that's why we originally went with 'Stop', otherwise we end up with something like this:

  ![image](https://user-images.githubusercontent.com/2829095/122673618-1ff95e00-d1c9-11eb-8ca0-5b887ace329b.png)
  
  We could use something like 'Stop PipelineRun' and 'Cancel PipelineRun' as the action labels to reduce the ambiguity but this could still cause some confusion.
- should we have a single action or separate them out? If we go with a single action in the overflow menu how do we allow the user to select between the different types of termination? Radio button selection in the confirmation dialog for example?

_Originally posted by @AlanGreene in https://github.com/tektoncd/dashboard/issues/2069#issuecomment-864545623_

AlanGreene:
Promotion of TEP-0058 fields to beta and deprecation of the `PipelineRunCancelled` status are being tracked in https://github.com/tektoncd/pipeline/issues/4611

Similar issue to this one tracking CLI plans: `https://github.com/tektoncd/cli/issues/1553`

williamlfish:
[...]

PR #2316, opened 2022-02-28
Update components to support displaying new `Cancelled` status

AlanGreene:
# Changes

TEP-0058 Graceful Termination deprecated the existing `PipelineRunCancelled`
status. Instead, runs have the status reason `Cancelled`.

The feature is still in alpha, so to test it out you can edit the `feature-flags`
ConfigMap to set `enable-api-fields: alpha`, then use one of the following values
when cancelling the PipelineRun:
- `Cancelled` replaces the existing `PipelineRunCancelled`, interrupts any
  executing tasks and does not run finally tasks
- `StoppedRunFinally` lets any executing tasks complete normally but does not
  start any new non-finally tasks, then runs finally tasks
- `CancelledRunFinally` interrupts any executing non-finally tasks, then executes
  finally tasks

In all 3 cases above, the PipelineRun status reason is `Cancelled`

Update tests and storybook to include 'Cancelled' variants.

The Dashboard API will still use `PipelineRunCancelled` to stop a PipelineRun.
A future change once the feature has been promoted to beta will switch to the
replacement `Cancelled` value. Some UI change may also be made to support the
alternative behaviours, tracking this in https://github.com/tektoncd/dashboard/issues/2094

# Submitter Checklist

  functionality, content, code)

for more details._

### DIFF
--- a/packages/components/src/components/StatusIcon/StatusIcon.stories.js
+++ b/packages/components/src/components/StatusIcon/StatusIcon.stories.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2020-2021 The Tekton Authors
+Copyright 2020-2022 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


packages/components/src/components/StatusIcon/StatusIcon.stories.js
```diff
@@ -34,12 +34,18 @@ export default {
   title: 'Components/StatusIcon'
 };
 
-export const Queued = args => <StatusIcon {...args} />;
+export const Cancelled = args => (
+  <StatusIcon reason="TaskRunCancelled" status="False" {...args} />
```
briangleeson, line 38: Should we add a `<StatusIcon reason="Cancelled" status="False" {...args} />` story as well?

AlanGreene: Yep, good point. I've added both the legacy `PipelineRunCancelled` and the upcoming `Cancelled`

briangleeson (approved):
/lgtm

PR #2322, opened 2022-03-09
Fix component imports within components package

AlanGreene:
# Changes

Instead of referencing `@tektoncd/dashboard-components`, imports
within the `components` package that reference other components
should load them locally from the index. This will be more important
in future when we provide the ability to override / customise parts
of the UI via extensions.

# Submitter Checklist

  functionality, content, code)

for more details._

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@tektoncd/dashboard-utils": "file:../utils",
+    "classnames": "^2.3.1",
     "js-yaml": "^3.13.0",
     "linkify-it": "^3.0.2",
     "prop-types": "^15.7.2",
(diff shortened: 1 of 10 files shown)


packages/components/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@tektoncd/dashboard-utils": "file:../utils",
+    "classnames": "^2.3.1",
```
briangleeson, line 22: This is only used within components and graphs packages. So we could remove it from the main package.json? https://github.com/tektoncd/dashboard/blob/main/package.json#L31

AlanGreene: Absolutely, I thought I had done that with this change but clearly not. Good catch!

AlanGreene: Updated

briangleeson (approved):
/lgtm

Issue #2324, opened 2022-03-11
UI becomes unresponsive when there are a large number of completed/running taskruns


rouke-broersma:
# Describe the bug

The browser crashes due to the amount of TaskRuns loaded at the same time

# Expected behaviour

The browser does not crash simply from opening the page

# Steps to reproduce the bug

Have ~1500-2000 completed and/or running taskruns, might probably also happen for other types

# Environment details

- Kubernetes Platform:
Kubernetes
- Kubernetes or OpenShift version:
AKS 1.22.6
- Install mode (if on OpenShift):
Helm chart
- Cloud-provider/provisioner:
AKS
- Versions:
   - Tekton Dashboard:
   v0.24.1
   - Tekton Pipelines:
   v0.33.2
- Install namespaces:
   - Tekton Dashboard:
   tekton-pipelines
   - Tekton Pipelines:
   tekton-pipelines

# Additional Info

Some pagination would probably be helpful

AlanGreene:
Thanks for reporting this @rouke-broersma. It's a known problem when there are a large number of resources on the cluster and has been discussed many times in the past both in GitHub issues and on Slack. We're currently facing some limitations of the Kubernetes API although we do have some plans to address this.

See https://github.com/tektoncd/dashboard/issues/1978#issuecomment-796636018 for some information on how we currently manage this in our own dogfooding cluster by removing older runs as well as plans for integration with Tekton Results which we hope will allow us to address some of the concerns about pagination, large numbers of resources, etc.

rouke-broersma:
Sorry, I looked through the open issues and could not find anything.

AlanGreene:
No problem at all. It took me a while to find them and I already knew what I was looking for 😅

rouke-broersma:
On my end it seems like the issue is not necessarily that we have to wait a long time to get the response from the k8s api but rather it seems that the rendering of all the items is the problem. Could the pagination not be performed client side instead, so that at least the items are not all rendered?

AlanGreene:
We tried that in the past but it didn't make a difference in terms of the perceived performance, but may be worth investigating again.

Can you share some more details about the number/size of the resources you're working with, the total response size for the list, or some examples? This could be helpful when we try to reproduce the issue and test any potential improvements.

rouke-broersma:
[...]

PR #2327, opened 2022-03-18
Add client-side pagination

AlanGreene:
# Changes

https://github.com/tektoncd/dashboard/issues/2324

Add client-side pagination to all list pages via the `ListPageLayout`
container. Update the `children` prop to expect a function so we can
use the render props pattern to allow the `ListPageLayout` component
to handle all of the pagination logic required.

Consuming pages pass their raw resources to `ListPageLayout` and it
passes back the correct slice for the current pagination settings
which the consumer then formats / renders as needed.

Although we're still loading and storing the full list of resources
from the API this does have significant performance benefits in the
client, especially with larger number of resources. Notably, this
addresses freezing reported by a number of users on the PipelineRuns
and TaskRuns pages with busy clusters.

# Submitter Checklist

  functionality, content, code)

for more details._


src/containers/ListPageLayout/ListPageLayout.js
```diff
@@ -72,6 +88,11 @@ export const ListPageLayout = ({
     setPath(newURL);
   }
 
+  const resourcesForCurrentPage = resources.slice(
+    (page - 1) * pageSize,
+    (page - 1) * pageSize + pageSize
```
briangleeson, line 93: Could simplify to `page * pageSize` for the second value?

AlanGreene: absolutely 🥇

AlanGreene: Done

briangleeson (approved):
/lgtm

PR #2356, opened 2022-05-10
Update design of runs pages

AlanGreene:
# Changes

Update the table layout used for the PipelineRuns and TaskRuns pages
to include more useful information such as trigger info, status details,
and combine fields where this makes sense (e.g. created time + duration).

# Submitter Checklist

  functionality, content, code)

for more details._


packages/components/src/components/PipelineRuns/PipelineRuns.js
```diff
@@ -91,27 +110,16 @@ const PipelineRuns = ({
...
-    }),
     pipeline: intl.formatMessage({
       id: 'dashboard.tableHeader.pipeline',
       defaultMessage: 'Pipeline'
     }),
-    namespace: 'Namespace',
-    createdTime: intl.formatMessage({
-      id: 'dashboard.tableHeader.createdTime',
-      defaultMessage: 'Created'
+    run: 'Run',
```
briangleeson, line 117: Just to confirm, should this be extracted or treating as a special word that is never translated?

packages/components/src/components/TaskRuns/TaskRuns.js
```diff
@@ -71,20 +102,17 @@ const TaskRuns = ({
 }) => {
   let hasRunActions = false;
   const headers = [
+    {
+      key: 'run',
+      header: 'Run'
```
briangleeson, line 107: Same question again about extracting the string

packages/components/src/components/Table/Table.js
```diff
@@ -117,19 +117,19 @@ const Table = props => {
     emptyTextAllNamespaces,
     emptyTextSelectedNamespace,
     filters,
+    hasDetailRows,
```
briangleeson, line 120: You had asked about the naming convetion here. Only other thought was `hasDetails`

Alternatively scrub the extra `hasDetailRows` prop and just set the `size` prop to `xl` for the Table component in PipelineRuns and TaskRuns? I dont recall from the walkthrough if you had a case in mind that meant the extra prop was needed

briangleeson (commented):
Looks good, and cheers for the walkthrough

Couple of minor questions, but otherwise good to go

packages/components/src/components/PipelineRuns/PipelineRuns.js
AlanGreene: special word, similar to 'PipelineRun' 👍

packages/components/src/components/Table/Table.js
AlanGreene: I prefer `hasDetails`, I'll update it. Thanks.

Yes it's needed to control positioning of the batch action checkboxes + overflow menu icons as they're not vertically centered in this case as they would be by just setting `size="xl"`, as well as some other tweaks to spacing of the content (see the `tkn--table-with-detail-rows` class which I'll also rename)

AlanGreene: Updated 👍

briangleeson (approved):
/lgtm

PR #2378, opened 2022-05-24
[WIP] Add new cancel states for "graceful" cancel 

williamlfish:
# Changes
Updates to use the new cancel statuses, with backwards compatibility for pipeline controllers that have not updated to consume them. There are still some changes needed for the wording around the new statuses from a ux perspective. A lot has been discussed in the issue [here](https://github.com/tektoncd/dashboard/issues/2094). 

One of the ideas was to add an info icon with a tool tip for and explanation, and from what I can tell the carbon radio buttons break when trying to add the icons ( I don't do a lot of front end, so could be a me problem 😅  ). So there is still the need to explain them, and a way to present. So I admit, this pr may not be ready, but perhaps a conversation starter? 

Juggling the version, and the status passed in is a little cumbersome and hopefully something that can be removed relatively soon. 

# Submitter Checklist

  functionality, content, code)


williamlfish:
@AlanGreene The pr we discussed a while back. Sorry it took me a bit to get to it 😅  but hopefully we can start from here? At the end of the day, just wanted to have it on your radar. thanks!

AlanGreene:
Thanks for the PR @williamlfish. I've just taken a very quick look and it seems like a solid start. It'll likely be a few days before I can do a proper review but I'll try to give some early feedback soon.

/ok-to-test

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -207,15 +265,30 @@ export function PipelineRuns({ intl }) {
             id: 'dashboard.cancelPipelineRun.primaryText',
             defaultMessage: 'Stop PipelineRun'
           }),
-          body: resource =>
-            intl.formatMessage(
+          secondaryButtonText: intl.formatMessage({
+            id: 'dashboard.modal.cancelButton',
+            defaultMessage: 'Cancel'
+          }),
```
AlanGreene, line 271: no need for this, it's the default text on the secondary button already

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -124,11 +140,53 @@ export function PipelineRuns({ intl }) {
     });
   }
 
+  function persistCancelState(state) {
+    setCancelState(state);
+    setDefaultCancelState(state);
```
AlanGreene, line 145: ideally we would only be setting this in one place and reading from that single place consistently.

We could use a similar to pattern to the one in [ListPageLayout](https://github.com/tektoncd/dashboard/blob/136c2d99483313b756321ea107b5ba024d2c862d/src/containers/ListPageLayout/ListPageLayout.js#L61-L63) for the pagination settings. Set the value in `state`, listen to `state` updates and write to `localStorage`. This way they should always be in sync and there's one trigger for the update.

We would read from `localStorage` on mount and use it to populate the initial value in `state`. From that point on `state` is our source of truth, until the page is reloaded.

src/utils/index.js
```diff
@@ -214,3 +214,31 @@ export function setTheme(selectedTheme = getTheme()) {
...
+function sanatizeCancelState(state) {
+  const valid = [
+    'Cancelled',
+    'CancelledRunFinally',
+    'StoppedRunFinally'
+  ].includes(state);
+  if (valid) {
+    return state;
+  }
+  return 'PipelineRunCancelled';
```
AlanGreene, line 227: We have a different default here vs.  the end of `getDefaultCancelState` below which calls this function. Should they be consistent / combined?

AlanGreene (changes requested):
I'm seeing some odd behaviour where the state sent in the API call doesn't always match the selected option in the UI. Possibly in part related to mismatch between the value in the local component state and the value in localStorage, commented below with a suggestion to address that. There may be some other issues causing the mismatch too.

That said the overall structure looks good. Just some styling / content issues to take care of as previously discussed but we're good to proceed with the approach 👍 

Thanks for tackling this @williamlfish and apologies for the delay getting around to reviewing 😸

williamlfish:
@AlanGreene thanks for the review, and no worries about the delay, we are all busy 😅! Speaking of busy, I should have time to get to these changes before next week. thanks again!

src/containers/PipelineRuns/PipelineRuns.js
```diff
@@ -38,22 +39,34 @@ import {
...
 
 export function PipelineRuns({ intl }) {
   const history = useHistory();
   const location = useLocation();
   const params = useParams();
+  const { data } = useProperties();
 
+  const { pipelinesVersion } = data;
+  const versionCheck = parseInt(pipelinesVersion.split('.')[1], 10);
+  const allowCancelOptions = !Number.isNaN(versionCheck) && versionCheck > 34;
```
AlanGreene, line 66: note to self: another option would be to use something like [`semver`](https://www.npmjs.com/package/semver) to perform the comparison, especially if we're going to end up with similar checks for other features. This could help with compatibility checks especially as Tekton Pipelines starts removing some of the v1alpha1 APIs (e.g. Conditions)

e.g. 
```js
const minVersion = '>=0.35.0';
const allowCancelOptions = semver.satisfies(pipelinesVersion, minVersion);
```

Depending on the size the package contributes to the build output we may want to import only the specific function used, there are some docs included for how to achieve that.

Issue #1970, opened 2021-03-05
Add support for Custom Tasks


AlanGreene:
# Is your feature request related to a problem ?

Dashboard should support displaying (at least) all resources provided by the core Tekton projects.

It does not currently display Custom Tasks (Run CRD) in any useful manner.

# Describe the solution you'd like

Add support for displaying / interacting with Custom Tasks (Run CRD) https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md
https://github.com/tektoncd/pipeline/blob/master/docs/runs.md

Details TBD...

# Describe alternatives you've considered

# Additional context

afrittoli:
I wanted to create this feature request... but it's already there :)
A first step here would be to able to display a `Run`. 
I'm not sure we want to be display the specific custom tasks, as they're experimental.

If there's a way display them based on the CRD alone, we could do that, or support some kind of plugin mechanism to support them? 

For things like task loops or pipelines in pipelines, we could stop relying on the status of a pipelinerun and instead search for all resources based on ownership information and/or labels, and that would give us a way to display all related resources.
It would require though being able to combine heterogeneous resources (taskrun, pipelinerun, run) in a single view.

AlanGreene:
@afrittoli you mentioned on the Tekton working group this week that you were working on updating some of our existing CI jobs to use custom tasks. Can you point me at an example of some of those so I can get a feel for how they're being used in practice and how we might display something useful to the user?

afrittoli:
This is an example of a task loop: https://dashboard.dogfooding.tekton.dev/#/namespaces/tekton-ci/pipelineruns/plumbing-unit-tests-flfmk?pipelineTask=clone-repo&step=clone

AlanGreene:
Thanks for the link. Looking at another [example](https://dashboard.dogfooding.tekton.dev/#/namespaces/tekton-ci/pipelineruns/plumbing-unit-tests-2ps97?pipelineTask=unit-tests&step=unit-test&view=logs) where the task loop actually ran I see that we're actually already retrieving the relevant `TaskRuns` from the cluster but because the Dashboard client code isn't aware of custom tasks it just uses the first `TaskRun` matching the pipelineTask name to render in the UI.

Some quick notes (mostly to myself so I can pick this up again later):
- `TaskRuns` generated by the task loop custom task have the following labels:
  - `"custom.tekton.dev/taskLoop": "golang-tests"`
[...]

PR #2418, opened 2022-08-19
Add ability to rerun and cancel Runs

AlanGreene:
# Changes

Related to https://github.com/tektoncd/dashboard/issues/1970

- Standalone Run resources can be rerun from the list page
  (i.e. not child of a Pipeline)
- Any Run that is currently running (i.e. status == Unknown)
can be stopped
- Update StatusIcon to render running state to match behaviour
  described in the Runs documentation

# Submitter Checklist

  functionality, content, code)

for more details._


packages/components/src/components/StatusIcon/StatusIcon.test.js
```diff
@@ -43,4 +43,8 @@ describe('StatusIcon', () => {
   it('gracefully handles unsupported state', () => {
     render(<StatusIcon reason="???" status="???" />);
   });
+
+  // it('renders completed step', () => {
```
briangleeson, line 47: Leaving this commented out?

src/containers/Runs/Runs.js
```diff
@@ -196,7 +214,44 @@ function Runs({ intl }) {
...
+      },
+      {
+        actionText: intl.formatMessage({
+          id: 'dashboard.cancelTaskRun.actionText',
+          defaultMessage: 'Stop'
+        }),
+        action: cancel,
+        disable: resource => {
+          const { status } = getStatus(resource);
+          return status && status !== 'Unknown';
```
briangleeson, line 233: Should this option be disabled if `!isRunning(...)`? or is this check equivalent?

packages/components/src/components/StatusIcon/StatusIcon.test.js
AlanGreene: not needed anymore, removed it 👍

src/containers/Runs/Runs.js
AlanGreene: `isRunning` is a little different. It only works for `TaskRun` and `PipelineRun` resources as we determine they're running by looking for the combination of `status === 'Unknown'` AND a known reason (e.g. `Running`, `Stopping`, etc.)

For `Run` resources, any `status === 'Unknown'` is considered running. If status is unset it means the controller hasn't started processing it yet, in which case we should also be allowed to prevent it from running.

I realise I mistakenly used `isRunning` for the delete action so I've updated to remove that and correctly check for `status === 'Unknown'` instead.

briangleeson (approved):
/lgtm

Issue #1283, opened 2020-04-11
Feature Request - Ability to bind Workspaces


davissp14:
It would be really nice to be able to configure a workspace prior to creating a PipelineRun within the dashboard.

I understand that triggers can be used to do this automatically through events, but for folks who want to manually kick something off through the dashboard, this would be pretty useful.

AlanGreene:
I agree it would be nice to be able to manually create runs using workspaces.

We're currently working on introducing the ability to manually create a TaskRun, similar to the existing functionality for PipelineRuns. As part of this we're likely to revisit some of the PipelineRun pieces.

Some initial thoughts for anyone interested in picking this up...

Adding support for workspaces would require:
- knowledge of available PVCs, ConfigMaps, changes to how we handle Secrets
  - would it be sufficient to list existing resources of these types without providing the ability to create / edit them?
- similar logic as we have for resources / params to detect the required resources and present them in the create UI
  - given the potential for many more items in this UI we really should look at moving to the full page experience instead of the current modal dialog, see item 5 of https://github.com/tektoncd/dashboard/issues/966
- possibly more...

Reference:
- [Workspaces docs](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md). More info in TaskRun and PipelineRun docs that may be of use.
- [TaskRun example](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/taskruns/workspace.yaml)
- [PipelineRun example](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/workspaces.yaml)

davissp14:
It seems a little silly that workspace definitions happen on the PipelineRun/TaskRun resource given how they are used.  Basically, you'd have to specify one-to-many workspaces, per run, and hope that the end-user is able to map the bindings correctly on each go.  It doesn't look like there' a reliable way to determine the workspace specifications/requirements by reading the pipeline.

Reference: https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md

> Creation of a PipelineRun will trigger the creation of TaskRuns for each Task in your pipeline.

It seems like this kind of configuration should belong on the Pipeline, or there needs to be another level of abstraction "PipelineTemplate" that can hold re-usable configuration data that PipelineRun's can inherit.

AlanGreene:
[...]

PR #2459, opened 2022-09-12
Bind Workspaces for PipelineRun

marniks7:
# Changes

Implements https://github.com/tektoncd/dashboard/issues/1283

- [x] UI: Workspaces are displayed as Dropdown with all possible kinds together: secrets, configmaps, pvc, emptyDir

Workspaces for:
- [x] PipelineRun
- [ ] TaskRun 
- [ ] Run

`VolumeSource`s:
- [x] ConfigMaps both UI and backend. 
- [x] Secrets both UI and backend
- [x] PVC both UI and backend
- [x] EmptyDir UI (no changes on backend required)
- [ ] PVC Template
- [ ] Projected (alpha)
https://tekton.dev/docs/pipelines/workspaces/#projected
- [ ] CSI (alpha)
https://tekton.dev/docs/pipelines/workspaces/#csi

Features:
- [x] Optional Workspaces
https://tekton.dev/docs/pipelines/workspaces/#optional-workspaces
- [x] Workspace Description: is displayed
- [ ] Sub-path - UI part was implemented, but later I dropped it

Notes:
- [x] K8s proxy & Security: tekton-backend will not allow to retrieve internals (sensitive data) of `configmaps`,  `secrets` or `persistentvolumeclaim` data by modifying response from k8s: only minimal required amount of data is preserved. Such filtration is applied to all urls with `secret` or `configmap` or `persistentvolumeclaim` in the path. 
> ANY request (not only to secrets or configmaps or `persistentvolumeclaim`) to namespace that contains such strings are NOT allowed as well. (this is false positive...)
- [x] RBAC: `configmaps`, `secrets`, `persistentvolumeclaims` (`list` and `watch` verbs) were added to the ClusterRole `tekton-dashboard-backend`.
> Note, this allows to read secret data, such issue is handled by limiting what backend `tekton-dashboard` returns to UI. (see above)
- [x] E2E tests with Cypress

# Submitter Checklist

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing. 
**Do we need any other docs?**
  functionality, content, code)

for more details._


AlanGreene (changes requested):
Thanks for the PR @marniks7. Overall it looks like a pretty good start but as discussed on Slack could do with being broken into a series of smaller changes so we can iterate on it more easily and do a more complete review.

I have some general comments:
- we only need to filter Secrets, the other resources do not contain sensitive information and we will actually be using ConfigMaps for other functionality in the near future and will require their full content
- when filtering the data it should only affect resources of the correct kind, it shouldn't affect other resources that happen to have 'secret' elsewhere in the API path (e.g. namespace, resource name, etc.)
- need to confirm that any websocket messages are also filtered, or block websocket connections for Secrets
- the user should be able to differentiate between types of resource when configuring the workspace, e.g. identify which resource named 'foo' is a Secret vs. ConfigMap vs. PVC. We could include the type in the option text as you suggested on the issue but I think it's worth exploring other approaches here to see what works best
- test resources should be no larger / more complex than needed to verify functionality, e.g. no need to include large cert values in secrets
- it should be possible to run the Cypress E2E tests in isolation and against any environment (e.g. local dev environment), shouldn't rely on setup from e2e-tests.sh creating specific resources etc.
- new files should just have current year in copyright header

I think the next step would be to open a PR with just the back end changes providing the new resources with filtering etc.

Let me know if you have any questions.

marniks7:
Hi @AlanGreene 
* ConfigMaps, PVC are internals, it is not a good idea to provide ability to see what is inside. for example, even here author suggests to create ConfigMap https://hub.tekton.dev/tekton/task/kubernetes-actions
* based on kind - ok
* websockets - ok
* Alternative is https://github.com/redhat-developer/vscode-tekton see "Pipeline with Workspace", they provide separate field with "Type = ConfigMap \ PVC \ Secret", but i don't like this idea - making it more complex.
** use can filter by text 
** there should be some separate configuration to be able to show only target type and target list of values (e.g. filtered by label) for pipeline or group of pipelines
* large test resources - ok
* We need to upload resources to cluster, and in case of PVC we need to wait for them to become "bound".
So, where do I put them and how to upload to cluster?
Alternative is to use "Import resources from repository", but not sure that it will work.
* current year - ok

AlanGreene:
> author suggests to create ConfigMap https://hub.tekton.dev/tekton/task/kubernetes-actions

ConfigMaps should not be used for sensitive info, I don't know why the author suggests that

> We need to upload resources to cluster, and in case of PVC we need to wait for them to become "bound".
So, where do I put them and how to upload to cluster?

The e2e tests can execute commands on the host if needed, e.g. run `kubectl` to apply resources etc. See https://docs.cypress.io/api/commands/exec

If the Dashboard already has permissions to access / modify the resources in question you can also use [`cy.request`](https://docs.cypress.io/api/commands/request) to access the APIs directly.

triThirty:
Hi @AlanGreene , is there any plan to merge this pr?

AlanGreene:
Thanks for your interest @triThirty. There are a number of outstanding issues that need to be addressed (some outlined in the description and comments above) and we're working through them at the moment. This PR will not be merged in its current form but we do plan to enable this functionality in a future release, keep an eye on #1283 for updates.

Issue #2465, opened 2022-09-15
v0.29.x has a regex in the `vendors` chunk which causes a `SyntaxError` in Safari


jbg:
# Describe the bug

After upgrading from `v0.28.0` to `v0.29.1`, the Tekton Dashboard fails to load in Safari 16 (a blank page is shown). Disabling cache doesn't help. I haven't tried other browsers.

The following error appears in the JS console:

```
[Error] SyntaxError: Invalid regular expression: invalid group specifier name
	(anonymous function) (vendors.4cad84fe7cb5fd068c1a.js:2)
```

Downgrading to `v0.29.0` does not help, but downgrading to `v0.28.0` resolves the problem.

`vendors.4cad84fe7cb5fd068c1a.js` is minified and there doesn't seem to be a source map, but downloading it and running it through prettier and then using a response override to get the browser to run the formatted version reveals the location of the syntax error, which is in this regex:

```
/(?<!\b(?:[a-z][a-z\d+\-.]{1,50}:))\/{2,}/g
```

Pasting this regex into Safari's JS console gives the same `SyntaxError`, but Chrome's JS console accepts it just fine.

# Expected behaviour

The dashboard loads and no `SyntaxError` is found in the JS console.

# Steps to reproduce the bug

1. Upgrade to `v0.29.0` or `v0.29.1`
2. Open the dashboard in Safari

# Environment details

- Kubernetes Platform: Amazon EKS
- Kubernetes or OpenShift version: `v1.23.7-eks-4721010`
- Versions:
   - Tekton Dashboard: `v0.29.1`
   - Tekton Pipelines: `v0.39.0`
   - Tekton Triggers: `v0.21.0`
- Install namespaces:
   - Tekton Dashboard: `tekton-pipelines`
   - Tekton Pipelines: `tekton-pipelines`
   - Tekton Triggers: `tekton-pipelines`

jbg:
The issue is probably the lookbehind, which is [not supported by any version of Safari yet](https://caniuse.com/js-regexp-lookbehind).

AlanGreene:
Thanks for reporting this @jbg. We had a similar issue earlier in the year and put a fix in place to pin the affected dependency to a supported version: https://github.com/tektoncd/dashboard/pull/2300

That fix is still in place but either doesn't appear to be working any more or this is now coming from a different package. I'll take a look later today, hopefully an easy fix and we'll publish a patch release 🤞

AlanGreene:
[...]

PR #2466, opened 2022-09-15
Fix for Safari

AlanGreene:
# Changes

Resolves https://github.com/tektoncd/dashboard/issues/2465

Remove `git-up` override to avoid conflicts in package resolution that resulted in the wrong version of the package being used in production builds.

This lead to some regex code that's unsupported in Safari being included in the production bundle, but not being surfaced in development builds.

The version was originally pinned to fix this same problem Due to changes in other dependencies since the original fix was put in place, this approach does not work for this particular package (we have a transitive dependency on an older version via one of our dev dependencies, and for some reason this was being picked up in production builds).

Removing the override allows for newer (fixed) versions of the affected package to be picked up.

# Submitter Checklist

  functionality, content, code)

for more details._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41105,7 +41105,8 @@
           }
         },
         "git-up": {
-          "version": "https://registry.npmjs.org/git-up/-/git-up-4.0.5.tgz",
+          "version": "4.0.5",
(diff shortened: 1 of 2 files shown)


briangleeson (approved):
/lgtm

PR #2468, opened 2022-09-15
Move global namespaces filter dropdown to the application header

AlanGreene:
# Changes

Users have been confused by the namespaces dropdown, not realising it affects filtering globally across the application. This is in no small part due to the fact that it doesn't appear on all pages, and behaviour is not always consistent.

Move the dropdown to the application header so that it appears on all pages, making it more clear that this affects the entire application, not just the current page.

Add some additional content to the existing NamespaceContext which is used to return to the correct list view when viewing a details page and selecting 'All Namespaces', and to ensure the URL can be updated correctly when selecting a different namespace (both for list and details pages).

Move and update the relevant tests from ListPageLayout and SideNav to the new HeaderBarContent container which now manages all logic related to namespace changes, both for the global filter and in the URL.

# Submitter Checklist

  functionality, content, code)

for more details._


briangleeson (approved):
/lgtm

thanks for the walkthrough

PR #2514, opened 2022-10-07
Update React Query

AlanGreene:
# Changes

React Query v4 switches to a new package name, so `react-query` is now `@tanstack/react-query`.

Breaking changes:
- all query keys must be arrays, so for example `'properties'` is now `['properties']`
- query data is read-only in the response from `useQuery`
- disabled queries are 'loading' by default, need to switch to `isInitialLoading` flag to prevent blocking page in loading state in case of disabled queries, e.g. Pipeline resource on PipelineRun page when using an inline Pipeline spec as we never need to request a Pipeline resource in that case
- `setLogger` removed, logger is now set as a property when defining the QueryClient
- `undefined` is no longer a valid value for the data returned for a successful query

Package imports and most of the key updates were achieved using the provided codemods. Some manual changes required to address the rest.

Tests updated to match including fix for infrequent flake in App.test

# Submitter Checklist

  functionality, content, code)

for more details._

### DIFF
--- a/config_frontend/setupTests.js
+++ b/config_frontend/setupTests.js
@@ -11,17 +11,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { setLogger } from 'react-query';
 import fetch from 'node-fetch';
 import { TextDecoder, TextEncoder } from 'util';
 
 import { server } from './msw';
-
-setLogger({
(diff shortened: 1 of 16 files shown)


briangleeson (approved):
/lgtm

briangleeson (approved):
/lgtm

PR #2565, opened 2022-11-03
Add E2E test to create pipeline run

marniks7:
# Changes

Add e2e test to create pipeline run that can be executed from any environment.
This is preparation for tests for workspaces and for yaml editor

# Submitter Checklist

  functionality, content, code)

### DIFF
--- a/packages/e2e/cypress/e2e/run/create-pipelinerun.cy.js
+++ b/packages/e2e/cypress/e2e/run/create-pipelinerun.cy.js
@@ -1,0 +1,62 @@
+/*
+Copyright 2022 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+const namespace = 'e2e';
+describe('Create Simple Pipeline Run', () => {
+  before(() => {
+    cy.exec('kubectl version --client');
+    cy.exec(`kubectl create namespace ${namespace} || true`);
+  });
+  it('should create pipelinerun', function () {
+    const uniqueNumber = Date.now();
+
+    const pipelineName = `simple-pipeline-${uniqueNumber}`;
+    const pipeline = `
+      apiVersion: tekton.dev/v1beta1
+      kind: Pipeline
+      metadata:
+        name: ${pipelineName}
+        namespace: ${namespace}
+      spec:
+        tasks:
+          - name: hello
+            taskSpec:
+              steps:
+                - name: echo
+                  image: busybox
+                  script: |
+                    #!/bin/ash
+                    echo "Hello World!"
+    `;
+    cy.exec(`echo "${pipeline}" | kubectl apply -f -`);
+    cy.visit(
+      `/#/pipelineruns/create?namespace=${namespace}&pipelineName=${pipelineName}`
+    );
+    cy.get('[id=create-pipelinerun--namespaces-dropdown]').should(
+      'have.value',
+      namespace
+    );
+    cy.get('[id=create-pipelinerun--pipelines-dropdown]').should(
+      'have.value',
+      pipelineName
+    );
+
+    cy.contains('button', 'Create').click();
+
+    cy.contains(`${pipelineName}-run`).parent().click();
+
+    cy.get('header[class="tkn--pipeline-run-header"]')
+      .find('span[class="tkn--status-label"]', { timeout: 15000 })
+      .should('have.text', 'Succeeded');
+  });
+});
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -48,6 +48,10 @@ test_dashboard() {
   # kubectl or proxy (to create the necessary resources)
   local creationMethod=$1
 
+  local readonly=false
+  if [[ "${@:2}" =~ "--read-only" ]]; then
+    readonly=true
+  fi
   header "Setting up environment (${@:2})"
   $tekton_repo_dir/scripts/installer install ${@:2}
   wait_dashboard_backend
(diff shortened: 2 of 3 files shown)


packages/e2e/cypress/e2e/run/create-pipelinerun.cy.js
```diff
@@ -0,0 +1,48 @@
...
+    cy.exec('kubectl version --client');
+    cy.exec('kubectl create namespace e2e || true');
+    cy.exec(`echo "${pipeline}" | kubectl apply -f -`);
+  });
+  it('should create pipelinerun', function () {
+    cy.visit(
+      `/#/pipelineruns/create?namespace=e2e&pipelineName=simple-pipeline`
+    );
+    cy.contains('button', 'Create').click();
+    cy.contains('simple-pipeline').first().click();
```
AlanGreene, line 42: Ideally the created run would have a unique identifiable name to ensure we're looking at the newly created one and not another run that happens to have the same name, possibly left over from a previous test.

This can be achieved by generating a unique value (e.g. adding a timestamp suffix) and typing it in the 'PipelineRun name' field of the create form.

test/e2e-tests.sh
```diff
@@ -48,6 +48,10 @@ test_dashboard() {
   # kubectl or proxy (to create the necessary resources)
   local creationMethod=$1
 
+  readonly=false
```
AlanGreene, line 51:
```suggestion
  local readonly=false
```

test/e2e-tests.sh
```diff
@@ -178,7 +182,11 @@ test_dashboard() {
 
   # In case of failure we'll upload videos of the failing tests
   # Our Cypress config will delete videos of passing tests before exiting
-  docker run --rm --network=host $CYPRESS_ENV -v $VIDEO_PATH:/home/node/cypress/videos dashboard-e2e || fail_test "Browser E2E tests failed"
+  CYPRESS_SPEC=""
+  if [[ "$readonly" == true ]]; then
```
AlanGreene, line 186:
```suggestion
  if $readonly; then
```

test/e2e-tests.sh
```diff
@@ -178,7 +182,11 @@ test_dashboard() {
...
+    CYPRESS_SPEC="-- --spec ""cypress/e2e/common/*.js"""
```
AlanGreene, line 187:
```suggestion
    CYPRESS_SPEC='-- --spec "cypress/e2e/common/**/*"'
```

gives us a bit more flexibility in how we organise tests in future, and Cypress is already configured to look for the `.cy.js` extension so we don't need to repeat it here.

packages/e2e/cypress/e2e/run/create-pipelinerun.cy.js
```diff
@@ -0,0 +1,48 @@
...
+
+    cy.get('header[class="tkn--pipeline-run-header"]', { timeout: 10000 })
```
AlanGreene, line 44: I'm curious where the timeout values here came from, was it from experimentation?
We'll need to keep an eye on these to ensure we don't end up with a flaky test.

AlanGreene (changes requested):
Thanks for the PR @marniks7. Overall it looks good, we just need to avoid doing anything platform specific where possible. Let me know if you have any questions.

packages/e2e/cypress/e2e/run/create-pipelinerun.cy.js
marniks7: It was experimentation over many runs (but I have done that for pipeline with workspaces)

reworked
* remove timeout for header (let's assume I my system has been hanging)
* decrease timeout for status from 15s to 10s
* add retry = 3
https://cypress.io/blog/2020/12/03/retry-rerun-repeat/

AlanGreene: Sorry I should have been clearer, my comment was just pointing out that we'll need to be aware of this and account for CI potentially being slower than a local cluster, not requesting a change in strategy.

I would prefer to increase the timeout for the status and remove the retries. Especially in CI where we create a new cluster and registry it's likely to take longer to run the first time due to pulling the image used in the pipeline. Subsequent attempts will most likely be faster as the image has already been pulled which is why the retries succeed. Using retries to accommodate this also has the side effect of creating multiple runs which we don't really need.

We should also consider cleanup after a successful run and delete it but that can be done later.

packages/e2e/cypress/e2e/run/create-pipelinerun.cy.js
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+const namespace = 'e2e';
+describe('Create Simple Pipeline Run', () => {
+  before(() => {
+    cy.exec('kubectl version --client');
+    cy.exec(`kubectl create namespace ${namespace} || true`);
+  });
+  it('should create pipelinerun', { retries: 3 }, function () {
+    const uniqueNumber = `${Math.floor(Math.random() * 100000)}`;
```
AlanGreene, line 21: We could also use `Date.now()` but this is fine, although no need to wrap it in quotes

packages/e2e/cypress/e2e/run/create-pipelinerun.cy.js
AlanGreene: Please also squash your commits when done so this is ready to merge.

marniks7: Ok,
* timeout increased to `15s` for status wait
   * on my pc and CI it takes 6s for everything (but image is pulled before that because `image: busybox` is used in static resources and those tests)
image
* retries removed

@AlanGreene, note, I have changed permission for `~/.kube/config` file to `644` otherwise permission denied on CI. I don't like that much for local runs, because it change permissions for common user kubeconfig file
Other options:
* copy `~/.kube/config` file to `~/.kube/tekton`, export KUBECONFIG=~/.kube/tekton`, change permissions and cleanup somewhere (don't know where for CI)
* `export KUBECONFIG=~/.kube/tekton`, then `kind export kubeconfig --name xxxx`, then `chmod 644 $KUBECONFIG` and finally mount `-v $KUBECONFIG:/home/node/.kube/config` (but it is not clear for me where cluster is created for CI). But it may have other problems for user, e.g. it will require to specify kubeconfig for each terminal session. For example, I use this kubecluster for development purposes and I don't tear down it after runs.

AlanGreene (approved):

AlanGreene:
Thanks @marniks7 
/lgtm

Issue #2437, opened 2022-09-05
Edit and Run


marniks7:
# Is your feature request related to a problem ?

CI can be used not only for automated operations, but to run any task that user might need.

E.g. I would like to run E2E tests (pipeline) manually with some parameters. They can fail.
It is nice to be able to change one (few) parameters and rerun it once again.
Or another operation, e.g. download backup, upload backup or run deploy.

Tekton can be used as common purpose pipeline \ workflow engine, e.g. for Chaos Engineering, and to experiment we need to rerun with different parameters like "duration of chaos", "wait"

# Describe the solution you'd like
I would like to see 'Rerun with Parameters' near 'Rerun' button for TaskRun\PipelineRun\Run

Jenkins analog: https://plugins.jenkins.io/rebuild/

# Describe alternatives you've considered
`kubectl apply -f` with different parameters, no alternatives in UI

# Additional context

- The idea is to use the same form as for `CreatePipelineRun` and add parameter `pipelinerun` to url, this parameter can be used as `template`
`http://localhost:9097/#/pipelineruns/create?namespace=tekton-namespace&pipelinerun=my-previous-pipeline-run`
- New button
![image](https://user-images.githubusercontent.com/9289172/188487812-1a361a03-1979-45e3-b9e6-5998f939339f.png)
- One more problem I see that `Rerun with Parameters` for this approach will support only those parameters that are specified in `CreatePipelineRun.js`, in particular I am worried about absence of workspaces. This can be done in separate PR
- I am willing to implement this functionality if no concerns

AlanGreene:
Thanks for the feature request @marniks7. I think this makes a lot of sense and would be great to add. However, as you mentioned I also have some concerns about missing support in the existing Create PipelineRun UI, particularly for workspaces, but also array params.

Adding a 'Rerun with parameters' (naming might need some work, 'parameters' is an overloaded term) makes it more likely that users will run into issues and could potentially end up with unexpected behaviour depending on how their pipelines are written. I think it's important that the user has full visibility into what they're creating, so not showing unsupported fields but having them passed in the background would not be good. Likewise, stripping unsupported fields wouldn't be great either as it's unlikely this would yield the desired outcome.

We're tracking the missing functionality for the Create UI in https://github.com/tektoncd/dashboard/issues/1283 and https://github.com/tektoncd/dashboard/issues/2171. Contributions on either of those would be very welcome and get us closer to being able to add this new feature safely.

marniks7:
[...]

PR #2633, opened 2022-12-12
Add new action for PipelineRun - Edit and Run

marniks7:
# Changes

Add new action for PipelineRun - Edit and Run.
Solves: https://github.com/tektoncd/dashboard/issues/2437
* Action from list of PRs
image
* Action from single PR
image
* Result of Edit and Run click:
image
* YAML page loading while PipelineRun is not fetched yet
image

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Add new feature: ability to edit and run new pipeline (completed or failed or running)
```


src/containers/CreatePipelineRun/YAMLEditor.test.js
```diff
@@ -212,3 +239,158 @@ describe('YAMLEditor', () => {
...
+      left: 0,
+      right: 0,
+      top: 0,
+      width: 0
+    });
+    Range.prototype.getClientRects = () => ({
+      item: () => null,
+      length: 0,
+      [Symbol.iterator]: jest.fn()
+    });
```
AlanGreene, line 261: We already have this in at least one other test, can it be extracted to an appropriately named function for reuse?

src/containers/PipelineRun/PipelineRun.js
```diff
@@ -328,6 +336,13 @@ export /* istanbul ignore next */ function PipelineRunContainer() {
           return isPending(reason, status);
         }
       },
+      {
+        actionText: intl.formatMessage({
+          id: 'dashboard.editAndRun.actionText',
+          defaultMessage: 'Edit and Run'
```
AlanGreene, line 342: nit: `Edit and run`
We use sentence case for UI elements: https://carbondesignsystem.com/guidelines/content/writing-style#use-sentence-case-capitalization

src/containers/CreatePipelineRun/YAMLEditor.js
```diff
@@ -23,21 +27,97 @@ import React, { useState } from 'react';
...
-import { createPipelineRunRaw, useSelectedNamespace } from '../../api';
+import { useLocation, useNavigate } from 'react-router-dom-v5-compat';
+import deepClone from 'lodash.clonedeep';
+import {
+  createPipelineRunRaw,
+  usePipelineRun,
+  useSelectedNamespace
+} from '../../api';
+
+export function generateNewPipelineRun(pipelineRun) {
```
AlanGreene, line 38: Could we reuse the existing code from the `rerunPipelineRun` API? Similar to what we've done with `createPipelineRun`, having a function to generate the sanitised payload.

AlanGreene (changes requested):
Thanks for the PR @marniks7. Overall it looks pretty good. I've left a few suggestions, let me know if you have any questions.

As we discussed on one of the previous issues / PRs I'd like to avoid adding more PipelineRun specific code into the YAML editor so we can more easily reuse it for other resource types in the near future.

src/containers/CreatePipelineRun/YAMLEditor.js
```diff
@@ -11,7 +11,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import { useIntl } from 'react-intl';
-import { ALL_NAMESPACES, urls } from '@tektoncd/dashboard-utils';
+import {
```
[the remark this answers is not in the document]
marniks7: >  The existing PipelineRun specific pieces can be left as they're beyond 
I was thinking about sending it as a code from CreatePipelineRun, but existence of PipelineRun specific pieces stopped me from doing that.
Thanks for clarifying this.

src/containers/CreatePipelineRun/YAMLEditor.js
marniks7: what about this one? we are not kinda re-running
```
'dashboard.tekton.dev/rerunOf': name
```
why do we need this one?
```
delete payload.metadata.labels['tekton.dev/pipeline'];
```

AlanGreene: `'dashboard.tekton.dev/rerunOf': name`

Yes, we can skip that one 👍 

`delete payload.metadata.labels['tekton.dev/pipeline'];`

We shouldn't set any `tekton.dev/*` labels, they're reserved for use by the Tekton Pipelines controllers, and adding them on the input could potentially cause issues. See https://github.com/tektoncd/pipeline/issues/4366 for some context.

packages/e2e/cypress/e2e/run/actions-pipelinerun.cy.js
```diff
@@ -0,0 +1,80 @@
...
+
+    cy.contains('button', 'Create').click();
+    cy.get(`[title*="${pipelineName}-run"]`)
+      .should('have.length', 1, { timeout: 10000 })
+      .parents('tr')
+      .find('[class="tkn--status"]')
+      .each(item => {
+        cy.wrap(item, { timeout: 15000 }).should('contain.text', 'Succeeded');
+      });
+    cy.contains(`${pipelineName}-run`).parent().click();
```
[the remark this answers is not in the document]
marniks7: This approach is used for the purpose to be similar to the second pipeline run validation.

packages/e2e/cypress/e2e/run/actions-pipelinerun.cy.js
```diff
@@ -0,0 +1,80 @@
...
+
+    cy.contains('button', 'Actions').click();
+    cy.contains('button', 'Edit and Run').click();
+    cy.contains('button', 'Create').click();
```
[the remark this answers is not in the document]
marniks7: So, I have tried todo that but failed.

`getByRole(/textbox/).textContent;` returns content without space `apiVersion: tekton.dev/v1beta1kind` and cannot be converted to yaml. `getByRole(/textbox/).innerHTML` requires even more manipulations. So only string validation can be used.  In `CreatePipelineRun.test.js` I have done the validation by using regexp.
However, for e2e we have another issue. Tekton has some default parameters https://tekton.dev/docs/pipelines/install/#customizing-basic-execution-parameters, e.g. `timeout: 1h0m0s` are populated by tekton engine one pipeline run creation. Do we want to compare those fields as well? Not sure. One of the options is to get first pipeline run, remove those fields (and fields in our logic), and compare. 
Another approach would be to use after `Create` with `cy.intercept`. In this case we will be able to use json\yaml and compare, but fields removal is required.
Instead of removal, since we don't know all the fields ahead, and they are subject to change, we can construct new pipelinerun with only known fields.
One more possible option is to leverage some advanced testing features \ frameworks for such goal as  skipping unknown fields. However, I am not aware of such, except json schema validation.

And finally the last option would be to use only some contains assertions.That is what I did.

packages/e2e/cypress/e2e/run/actions-pipelinerun.cy.js
```diff
@@ -0,0 +1,80 @@
...
+
+    cy.get(`[title*="${pipelineName}-run"]`)
+      .should('have.length', 2, { timeout: 10000 })
+      .parents('tr')
+      .find('[class="tkn--status"]')
+      .each(item => {
+        cy.wrap(item, { timeout: 15000 }).should('contain.text', 'Succeeded');
+      });
```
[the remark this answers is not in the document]
marniks7: I don't see how we can use `name` since `generateName` is generated by us and what for...

changed to what you have suggested, but with `*`, like `[title*="${pipelineName}-run"]`

src/containers/CreatePipelineRun/YAMLEditor.test.js
marniks7: after rework it is not needed

marniks7:
Didn't want to create separate PR for an issue with `pipelinePendingStatus`, in yaml it was `status: !<tag:yaml.org,2002:js/undefined>`.
It is fixed

packages/e2e/cypress/e2e/run/actions-pipelinerun.cy.js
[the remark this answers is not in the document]
AlanGreene: > Used only one contains assertions for pipeline name

Yes that's fine, we just want to ensure the content is present and reasonable. Checking the name is enough for now. No need to do a full comparison.

src/api/pipelineRuns.test.js
```diff
@@ -333,3 +335,113 @@ it('startPipelineRun', () => {
...
+  expect(yaml.dump(payload)).toEqual(expected);
+});
+it('generate new pipeline run maximum', () => {
+  const pr = {
+    apiVersion: 'tekton.dev/v1beta1',
+    kind: 'PipelineRun',
+    metadata: {
+      annotations: {
+        keya: 'valuea',
+        'kubectl.kubernetes.io/last-applied-configuration': {
```
AlanGreene, line 373: nit: the value of an annotation should be a string

src/containers/CreatePipelineRun/YAMLEditor.js
```diff
@@ -26,18 +26,29 @@ import { yaml as codeMirrorYAML } from '@codemirror/legacy-modes/mode/yaml';
...
+  if (prevLoading !== loading) {
+    setState(state => ({
+      ...state,
       code: initialCode,
-      isCreating: false,
-      submitError: '',
-      validationErrorMessage: ''
-    });
+      prevLoading: loading
+    }));
```
AlanGreene, line 50: `useEffect` should be used to run effects when certain values change

src/containers/CreatePipelineRun/YAMLEditor.js
```diff
@@ -183,6 +194,7 @@ export function CreateYAMLEditor({ code: initialCode = '' }) {
             theme="dark"
             extensions={[StreamLanguage.define(codeMirrorYAML)]}
             onChange={onChange}
+            editable={!loading}
```
AlanGreene, line 197: We should display some form of loading state, otherwise the user won't know why the editor is disabled.

AlanGreene (changes requested):
/ok-to-test

Please also squash your commits when done

packages/e2e/cypress/e2e/run/actions-pipelinerun.cy.js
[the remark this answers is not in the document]
AlanGreene: My suggestion was to switch from using `generateName` to `name` so we have full control over the name used for the resource instead of relying on kubernetes to add a randomly generated suffix. That would allow us to query for the specific run we're interested in, instead of using a broader match.

However, as long as `pipelineName` is unique to this test run and we're only creating a single PipelineRun, using `*=` will work, we just need to be careful when modifying this test in future.

src/containers/CreatePipelineRun/YAMLEditor.js
AlanGreene: Yes either some kind of spinner or skeleton until the resource is loaded and ready to display.

Carbon provides a number of elements for this purpose:
- https://v7-react.carbondesignsystem.com/?path=/story/components-loading--default
- https://v7-react.carbondesignsystem.com/?path=/story/components-skeleton-skeletonplaceholder--default&knob-Active%20(active)=true&knob-Classes%20with%20different%20sizes=my--skeleton__placeholder--large&knob-Description%20(description)=Active%20loading%20indicator&knob-Small%20(small)=true

Personally I think the Loading component with some text explaining what it's doing would be best here, but if you have another suggestion I'm willing to consider it.

You can find an example of how we use the Loading component in the LoadingShell while the Dashboard application is loading:
https://github.com/tektoncd/dashboard/blob/896eb37908d12c9f1d2af6d60647c5f9f32e3e83/packages/components/src/components/LoadingShell/LoadingShell.js#L81-L84

packages/e2e/cypress/e2e/run/actions-pipelinerun.cy.js
```diff
@@ -0,0 +1,78 @@
...
+    cy.get('[id=create-pipelinerun--namespaces-dropdown]').should(
+      'have.value',
+      namespace
+    );
+    cy.get('[id=create-pipelinerun--pipelines-dropdown]').should(
+      'have.value',
+      pipelineName
+    );
+    cy.contains('button', 'Create').click();
+
```
AlanGreene, line 59: We had problems with flakiness in the 'create' e2e so made some changes there, we'll need to do the same here. Specifically the following 2 changes:

- after clicking Create, ensure navigation has completed
  ```js
  cy.contains('h1', 'PipelineRuns');
  ```
- update query to specify an anchor tag as it was occasionally clicking on a matching span instead which caused the test to fail, also no need for `parent()` this way
  ```js
  cy.contains('a', `${pipelineName}-run`).click();
  ```

packages/utils/src/utils/index.js
```diff
@@ -356,6 +356,14 @@ export function getGenerateNamePrefixForRerun(name) {
   return `${root}${rerunIdentifier}`;
 }
 
+export function getGenerateNamePrefixForNewRun(name) {
+  let generateName = name;
+  if (!name.endsWith('-')) {
+    generateName = `${name}-`;
+  }
+  return `${generateName}`;
```
AlanGreene, line 364: no need to use template here:
```suggestion
  return generateName;
```

packages/utils/src/utils/index.js
```diff
@@ -356,6 +356,14 @@ export function getGenerateNamePrefixForRerun(name) {
   return `${root}${rerunIdentifier}`;
 }
 
+export function getGenerateNamePrefixForNewRun(name) {
+  let generateName = name;
+  if (!name.endsWith('-')) {
+    generateName = `${name}-`;
+  }
```
AlanGreene, line 363: This function is only called if `generateName` is not provided, and a valid `name` cannot end with `-` so this function could be simplified to just:

```js
export function getGenerateNamePrefixForNewRun(name) {
  return `${name}-`;
}
```

Instead of having a new function for this I'd suggest just doing it inline where it's used since it's only used in one place…

src/api/pipelineRuns.js
```diff
@@ -170,30 +173,46 @@ export function createPipelineRun({
...
+    annotations,
+    generateName: getGenerateName(),
+    labels,
     namespace
   };
+  if (rerun) {
+    payload.metadata.labels = {
+      ...labels,
+      'dashboard.tekton.dev/rerunOf': name
+    };
```
AlanGreene, line 203: Let's keep the `...labels` above to handle potentially missing labels, set defaults, etc.

Then here for the rerun, just add the one specific to reruns:

`payload.metadata.labels['dashboard.tekton.dev/rerunOf'] = name;`

This approach keeps clear separation between the two and makes it easier to refactor in future.

src/api/pipelineRuns.js
```diff
@@ -170,30 +173,46 @@ export function createPipelineRun({
...
+  }
 
-  Object.keys(payload.metadata.labels).forEach(label => {
-    if (label.startsWith('tekton.dev/')) {
-      delete payload.metadata.labels[label];
-    }
-  });
+  if (payload.metadata.labels) {
```
AlanGreene, line 206: with the suggestion above this check wouldn't be required as `labels` would always be defined

similar below for annotations

src/api/pipelineRuns.js
```diff
@@ -170,30 +173,46 @@ export function createPipelineRun({
   return post(uri, payload).then(({ body }) => body);
 }
 
-export function rerunPipelineRun(pipelineRun) {
-  const { annotations, labels, name, namespace } = pipelineRun.metadata;
+export function generateNewPipelineRunPayload(pipelineRun, rerun) {
```
AlanGreene, line 176:
```suggestion
export function generateNewPipelineRunPayload({ pipelineRun, rerun }) {
```

This makes it easier to understand the code when you see a call to this function:

`…generateNewPipelineRunPayload({ pipelineRun, rerun: true })`
instead of
`…generateNewPipelineRunPayload(pipelineRun, true)`

src/api/pipelineRuns.test.js
```diff
@@ -326,3 +328,226 @@ it('startPipelineRun', () => {
     }
   );
 });
+
+it('generate new pipeline run minimum for rerun', () => {
```
AlanGreene, line 332: I think it would be better to replace `minimum` and `maximum` in these tests with descriptions of what they're actually doing, e.g. 'with labels, annotations, and params' or whatever

This gives more context in test reports etc. without having to revert to reading the code to understand what the test is doing.

src/api/pipelineRuns.test.js
```diff
@@ -326,3 +328,226 @@ it('startPipelineRun', () => {
...
+  pipelineRef:
+    name: simple
+  params:
+    - name: param-1
+    - name: param-2
+`;
+  expect(namespace).toEqual('test-namespace');
+  expect(yaml.dump(payload)).toEqual(expected);
+});
+it('generate new pipeline run last applied configuration should be removed', () => {
```
AlanGreene, line 489: This could be combined with one of the tests above to avoid all the setup duplication, same for the labels test below. Logically it's part of the same processing, preserving only the relevant annotations/labels on the generated payload.

src/containers/CreatePipelineRun/CreatePipelineRun.js
```diff
@@ -56,7 +58,7 @@ const initialState = {
   nodeSelector: [],
   params: {},
   paramSpecs: [],
-  pendingPipelineStatus: '',
+  pipelinePendingStatus: '',
```
AlanGreene, line 61: good catch, thanks 👍

src/containers/CreatePipelineRun/YAMLEditor.js
```diff
@@ -16,29 +16,41 @@ import {
...
 
-  const [{ code, isCreating, submitError, validationErrorMessage }, setState] =
+  const [{ isCreating, submitError, validationErrorMessage, code }, setState] =
     useState({
-      code: initialCode,
       isCreating: false,
       submitError: '',
-      validationErrorMessage: ''
+      validationErrorMessage: '',
+      code: initialCode
```
AlanGreene, line 44: nit: please revert this to keep them listed alphabetically, it makes it easier to scan the list

src/containers/CreatePipelineRun/YAMLEditor.js
```diff
@@ -16,29 +16,41 @@ import {
...
     });
 
+  useEffect(() => {
+    setState(state => ({
```
AlanGreene, line 48: we only want to do this `if(!loading)` right?

what happens in case of an error loading the PipelineRun?

src/containers/CreatePipelineRun/YAMLEditor.test.js
```diff
@@ -211,4 +212,17 @@ describe('YAMLEditor', () => {
...
+      <CreateYAMLEditor loading loadingMessage="wait. test is in progress" />
+    );
+    expect(getAllByText(/wait. test is in progress/)).toBeTruthy();
+  });
+
+  it('during loading button create should be disabled and button cancel should not', () => {
+    const { queryAllByText } = renderWithRouter(<CreateYAMLEditor loading />);
+    expect(submitButton(queryAllByText).disabled).toBe(true);
+    expect(cancelButton(queryAllByText).disabled).toBe(false);
+  });
```
AlanGreene, line 227: these two tests could be combined as they're part of the same piece of functionality, one test for the loading state should be sufficient as it's not particularly complex

AlanGreene (changes requested):
Thanks @marniks7, this is looking pretty good now.

src/containers/CreatePipelineRun/YAMLEditor.js
marniks7: yeah, added

- in case of the error code mirror content will be empty, but editable (loading true -> false)
- no information about the error will be displayed to user
image

src/api/pipelineRuns.test.js
marniks7: - `rerun. generate new pipeline run with spec, metadata - name and namespace`
- `rerun. generate new pipeline run with spec, status, metadata - annotations, labels, uid and resourceVersion`

vs

- `rerun. generate new pipeline run. minimum`
- `rerun. generate new pipeline run. maximum`

Here is the rational behind `maximum`
- nobody would bother to read such long description (with listing words)
- nobody would bother to change such description on changes, which makes description incorrect and more harmful
- `maximum` assumed to be used by anybody making changes there and naming helps to convince others about that (sure, not 100% convincible)

Here is the rational behind `Minimum`
- making sure that this is a minimum input and in the absence of other fields function doesn't fail
- by word `minimum` it prevents others from making unrelated modifications in this test (sure, not 100% convincible)

src/api/pipelineRuns.js
marniks7: Modern IDEs are cable of solving such problems without making it more complex (adding 2 `{`, `}` with adding for each function `{`, `}` and `rerun`).
image
Github supports definition display and step into (for some languages)

AlanGreene: There are other benefits beyond the self-documenting nature of the code, e.g. consistency with the rest of the codebase, and easier refactoring if new params added in future (especially with any optional / default values) as order won't affect call-sites.

src/containers/CreatePipelineRun/YAMLEditor.js
AlanGreene: OK that will do for now but we may revisit later to display some error message to the user. They can always refresh the page to retry if needed and it should be rare enough unless they're following an old link that was shared / bookmarked.

src/api/pipelineRuns.test.js
AlanGreene: > only few would change such description on changes, which makes description incorrect and more harmful

I agree this could be a problem, I shouldn't have listed the affected fields in my example, I was just trying to give an example that explains what it's doing, e.g. "… with all processed fields", perhaps that is a better description. Also, it is up to the maintainers to ensure that if the test changes the test and its description still match / make sense.

>  prevents others from making unrelated modifications in this test

Not sure about this, also as above it would be the responsibility of the maintainers to identify this and provide feedback if it happens.

I still think we can be clearer than 'minimum' and 'maximum' here but I won't block the PR on this.

marniks7: Changed to
- `edit. generate new pipeline run with all processed fields`
- `edit. generate new pipeline run with minimum possible fields`

Another option would be to switch to more business terms, e.g.
- generate new pipeline run from completed pipeline run
at the same time I don't provide 100% of fields (however I can copy-paste from any completed pipeline run and that would make it very close to that)
- generate new pipeline run from non-created pipeline run with minimum fields

AlanGreene (approved):
Thanks @marniks7. This will be included in the v0.32 release due in the next few days.
/lgtm

PR #2634, opened 2022-12-12
Update to Cypress 12

AlanGreene:
# Changes

The Session and Origin experiment is now GA which means that the flag in the config is not required to enable it, and also that test isolation is now enabled by default.

Test isolation causes Cypress to navigate to `about:blank` between tests, so update the read permissions test to disable this as we only want to navigate to the app once then iterate over all of the side nav links. Without this change the `before` would need to be changed to a `beforeEach` and the tests would take significantly longer without any real additional benefit.

Also swith the namespace used for the 'create' tests to something less likely to cause conflicts, especially important as we delete the namespace at the end of the tests.

/kind misc
# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Update to Cypress 12
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16984,9 +16984,9 @@
       "dev": true
     },
     "node_modules/cypress": {
-      "version": "11.2.0",
-      "resolved": "https://registry.npmjs.org/cypress/-/cypress-11.2.0.tgz",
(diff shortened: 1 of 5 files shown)


briangleeson (approved):
/lgtm

PR #2649, opened 2023-01-04
Add toggle to settings UI to opt-in to Pipelines v1 resources

AlanGreene:
# Changes

Resolves https://github.com/tektoncd/dashboard/issues/2598

Add a toggle to the settings page allowing users to opt-in to requesting and managing v1 versions of the Pipelines CRDs: `Pipeline`, `PipelineRun`, `Task`, and `TaskRun`. This updates the API endpoints to use the corresponding version.

The toggle is off by default and will be included in the next release. Following that it will be on by default for one release, then removed.

/kind feature

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (new features, significant UI changes, API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Add ability to opt-in to using v1 `Pipeline`, `PipelineRun`, `Task`, and `TaskRun` resources with a compatible Tekton Pipelines release (v0.43.0+). This will be enabled by default in a future release.
```

PR #2651, opened 2023-01-04
Update import resources to specify version v1beta1

AlanGreene:
# Changes

Follow-up to https://github.com/tektoncd/dashboard/pull/2649

Stick with v1beta1 API version for the import resources API call (create PipelineRun) as we currently rely on PipelineResources which are no longer available in v1.

We'll be switching to using the resolvers included since Tekton Pipelines v0.40 instead, but pin the version for now to prevent issues when users enable the v1 toggle on the settings page.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2019-2022 The Tekton Authors
+Copyright 2019-2023 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


briangleeson (commented):
Worth adding a comment here about the hard-coded version?

AlanGreene:
It's the same as the previous change for ClusterTasks and other resources that don't yet provide v1 support, and I expect it'll be gone in a week or two anyway once we switch to using resolvers to pull in the git-clone tasks instead of using the git PipelineResource.

briangleeson (approved):
/lgtm

PR #3001, opened 2023-07-21
Remove use of defaultProps in functional components

AlanGreene:
# Changes

defaultProps for functional components will be deprecated in react 18.3.0 and the recommendation is to use JavaScript native support for default parameters instead.

One additional issue to bear in mind is that the default values must be serially stable, i.e. non-primitive values should not be set inline, otherwise the prop would appear to have a new value on each re-render.

This means things like `[]`, `{}`, and `() => {}` should be defined outside the function params.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


packages/components/src/components/TaskRunDetails/TaskRunDetails.js
```diff
@@ -54,7 +54,16 @@ function getDescriptions(array) {
...
 
-const TaskRunDetails = ({ onViewChange, pod, task, taskRun, view }) => {
+const noop = /* istanbul ignore next */ () => {};
+const emptyObj = {};
+
+const TaskRunDetails = ({
+  onViewChange = noop,
+  pod,
+  task = emptyObj,
+  taskRun = emptyObj,
```
briangleeson, line 64: Any chance that setting `task` and `taskRun` to the same initial object could result in cross-contamination when values are added to either object?

packages/components/src/components/Table/Table.js
```diff
@@ -109,26 +109,27 @@ function getTranslateWithId(intl) {
...
+  headers: dataHeaders,
+  id = null,
+  isSortable = false,
+  loading = false,
+  rows: dataRows,
+  selectedNamespace = null,
+  size = hasDetails ? 'xl' : undefined,
+  skeletonRowCount,
+  title = null,
+  toolbarButtons = emptyArray
```
briangleeson, line 130: Same question but for Arrays, any chance that setting `batchActionButtons` and `toolbarButtons` to the same initial array could result in cross-contamination when values are added to either array?

packages/components/src/components/TaskRunDetails/TaskRunDetails.js
AlanGreene: It shouldn't as the props shouldn't be modified by the component but it's a risk, and would be difficult to debug in future if we ran into a problem caused by it.

I've updated all of these to define a `defaults` object (similar to the old `defaultProps`) assigning the value for each affected prop, and reference that in the params instead.

So for example this one becomes:
```js
task = defaults.task,
taskRun = defaults.taskRun,
```

packages/components/src/components/Table/Table.js
AlanGreene: See comment above

PR #3030, opened 2023-08-08
[DO NOT MERGE] Switch from webpack and babel to vite

AlanGreene:
# Changes

**NOTE**: This PR is not intended to be merged in its current state. It will be used for discussion about potential changes to the Dashboard, its build process, user-facing enhancements, etc.

Any accepted changes will then be separated out and PRs submitted with the usual documentation, tests, etc.

This is still a work in progress, not all features and functionality have been tested.

TODO:
- compare bundle content + sizes (JS chunks seem bigger)
- verify Carbon styling (dark mode, hover states, etc.)
- split chunks (started with YAMLEditor)
- verify impact on consumers of the packages
- … ?

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [ ] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [ ] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [ ] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [ ] Release notes block below has been updated with any user facing changes (new features, significant UI changes, API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

PR #3033, opened 2023-08-12
Lazy-load the YAML editor

AlanGreene:
# Changes

Closes https://github.com/tektoncd/dashboard/pull/3030

- Add new `Loading` component for reuse across the app

  There are a number of full-screen loading states in the app that have very similar design. Extract these to a common component to ensure consistency since we'll be adding more loading states when code-splitting / lazy-loading the YAML editor and its dependencies in the next commit

- Lazy-load the YAML editor

  Use dynamic import with React.lazy to delay loading of the YAML editor and CodeMirror dependencies until the user switches the Create page to YAML mode. This significantly reduces the initial bundle, improving load times and reducing processing overhead for the majority of users who do not use this functionality, including all users running the Dashboard in read-only mode.

  Remove the core-js polyfill as it's no longer required, and vite will warn us if we're using features not supported by the specified build target (currently es2021).

  Enable vendor chunk splitting plugin. The end-result is that the build now produces 3 chunks:
    - one with the bulk of the Dashboard's own code
    - one with the YAMLEditor component and its dependenceis (CodeMirror etc.)
    - vendor chunk with the remaining 3rd party deps from node_modules

  The main chunk and vendor chunk are loaded in parallel so there's no additional delay introduced by splitting these.
  
/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


src/containers/index.js
```diff
@@ -53,4 +53,3 @@ export { default as Trigger } from './Trigger';
 export { default as Triggers } from './Triggers';
 export { default as TriggerTemplate } from './TriggerTemplate';
 export { default as TriggerTemplates } from './TriggerTemplates';
-export { default as YAMLEditor } from './YAMLEditor';
```
AlanGreene, line 56: Had to remove this to let the chunk splitting work properly, vite helpfully pointed this out:

```
(!) dashboard/src/containers/YAMLEditor/index.js is dynamically imported by dashboard/src/containers/CreatePipelineRun/CreatePipelineRun.jsx, dashboard/src/containers/CreateTaskRun/CreateTaskRun.jsx but also statically imported by dashboard/src/containers/index.js, dynamic import will not move module into another chunk.
```

src/containers/YAMLEditor/YAMLEditor.jsx
```diff
@@ -15,13 +15,14 @@ import {
...
-  InlineNotification,
-  Loading
+  InlineNotification
 } from 'carbon-components-react';
 import yaml from 'js-yaml';
 import React, { useEffect, useState } from 'react';
 import CodeMirror from '@uiw/react-codemirror';
-import { langs } from '@uiw/codemirror-extensions-langs';
+import { StreamLanguage } from '@codemirror/language';
+import { yaml as yamlMode } from '@codemirror/legacy-modes/mode/yaml';
```
AlanGreene, line 24: Switched back to the previous method of loading the YAML definition as the 'new' way results in problems with tree-shaking and unwanted additional language definition files being produced in the build output.

See comment below with fix (workaround) for previous problem with vitest.

PR #3092, opened 2023-08-28
Check for the wrong start date

jisoolee:
Add wrongDateChecker to catch the wrong start date which is causing the wrong duration for PipelineRuns and TaskRuns

# Changes

This will prevent to show the wrong duration for the users.
Related issue: https://github.ibm.com/one-pipeline/adoption-issues/issues/1374

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/packages/components/src/components/DetailsHeader/DetailsHeader.jsx
+++ b/packages/components/src/components/DetailsHeader/DetailsHeader.jsx
@@ -28,7 +28,7 @@ class DetailsHeader extends Component {
         stepStatus.terminated || {});
     }
 
-    if (!startTime || !endTime) {
+    if (!startTime || !endTime || new Date(startTime).getTime() == 0) {
       return null;
     }
 


packages/components/src/components/DetailsHeader/DetailsHeader.jsx
```diff
@@ -28,7 +28,9 @@ class DetailsHeader extends Component {
         stepStatus.terminated || {});
     }
 
-    if (!startTime || !endTime) {
+    const wrongDateChecker = new Date(startTime).getTime()
+
+    if (!startTime || !endTime || wrongDateChecker == 0) {
```
AlanGreene, line 33: If the startTime is either an empty string `''` or `undefined` this will not give the expected result.

`new Date('')` and `new Date(undefined)` both result in an 'Invalid Date' object, and the call to `getTime()` will return `NaN` instead of 0.

Since this is a relatively simple expression we can just inline it in the if-statement after we've already checked that `startTime` is truthy so we know it won't hit these cases. 

```suggestion
    if (!startTime || !endTime || new Date(startTime).getTime() == 0) {
```

AlanGreene (changes requested):
Thanks for the PR @jisoolee. I've left a suggestion for a change to make sure handles some other possible bad input.

AlanGreene (approved):
/lgtm

Issue #3297, opened 2024-02-09
Ability to select an initial namespace on GUI


achetronic:
# Feature request

By default, when entering into the GUI, in the upper right selector, "All Namespaces" is selected, but when you have a giant cluster with a lot of PipelineRun resources at the same time, this seems to be slow and cause more problems when a lot of people enters at the same time

# Use case

If a flag like `--default-selected-namespace` is added, you can choose an empty namespace and force the user to select an specific one

# Alternatives

N/A

# Additional context

Latest version of everything nowadays (dashboard, pipelines, kubernetes 😄 )

achetronic:
CC: @jdominguez408 FYI

AlanGreene:
Hi @achetronic, thanks for the feature request. From your description it sounds like you'd benefit from disabling the 'All Namespaces' option too. This is actually possible today by setting the `namespaces` arg in the Dashboard deployment, or via the `--tenant-namespaces` option of the installer script.

This limits the displayed namespaces to only those specified, removing the 'All Namespaces' option. The namespaces are displayed in sorted order and the first one is selected by default.

Would this suit your requirements or do you still see a need for something like the 'default' flag you described?

achetronic:
hello @AlanGreene and thanks for the answer in advance :) 

The point is that this is an entire cluster for the pipelines, so we want the ability to inspect all namespaces, as each namespace has different pipelines runs entering. But the problem is that they are too many, so if you enter in the GUI selecting "All Namespaces", multiplied by the number of users, is slowing down sometimes even the cluster

The point is that namespaces are created dynamically, so we can not select only some of them with the flag you mentioned, as I understand it's used to select only some of them. Honestly, for us the dashboard is quite good, but only lacking of a flag to not select "All Namespaces" unless you explicitly select it on the GUI

WDYT? :)

AlanGreene:
I see, thanks for the additional context. Yes unfortunately that config won't help in the case of dynamically created namespaces.

We'll need to consider what approaches might make sense to address this type of issue and the impact they would have on other functionality. We may end up with a 'default namespace' setting as you suggested, or perhaps an entirely different solution.

Can you also provide some additional data to help understand the load you're seeing? i.e. number of concurrent users, number of namespaces, runs, etc.

achetronic:
[...]

PR #3334, opened 2024-02-28
Add support for configuring default namespace

AlanGreene:
# Changes

Resolves https://github.com/tektoncd/dashboard/issues/3297

When the Dashboard deployment's `--default-namespace` arg is set, this configures teh default namespace selected when loading the application instead of 'All Namespaces' being selected.

This behaviour will only be triggered when the Dashboard is loaded with no path on the URL, i.e. only for the root (which automatically redirects to the About page).

The reason for this limitation is that handling the default namespace configuration as part of the normal namespace processing would break the ability to select 'All Namespaces' from the namespace filter dropdown, as well as causing issues with URL consistency. It would also be impossible to differentiate between a user attempting to visit a namespaced page (e.g. PipelineRuns) with 'All Namespaces' selected, vs. with no selection which would be expected to use the default namespace instead.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Add the ability to configure the default namespace selected in the UI instead of 'All Namespaces' when the user visits the application root (i.e. `/`).
```

### DIFF
--- a/cmd/dashboard/main.go
+++ b/cmd/dashboard/main.go
@@ -34,12 +34,13 @@ var (
 	kubeConfigPath     = flag.String("kube-config", "", "Path to kube config file")
 	portNumber         = flag.Int("port", 8080, "Dashboard port number")
 	readOnly           = flag.Bool("read-only", true, "Enable or disable read-only mode")
-	logoutURL          = flag.String("logout-url", "", "If set, enables logout on the frontend and binds the logout button to this url")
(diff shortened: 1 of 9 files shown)


AlanGreene:
Some additional testing required to ensure compatibility between this new config option and the existing tenant namespaces behaviour, and no other unexpected side effects.

briangleeson (commented):
Looks good, just a minor string suggestion

briangleeson (approved):
/lgtm

Issue #2452, opened 2022-09-09
Update React Router to v6


AlanGreene:
Update react-router / react-router-dom to the v6 releases. This will likely be a breaking change for consumers of the Dashboard's `@tektoncd/dashboard-components` and `@tektoncd/dashboard-utils` packages and will need to be managed carefully.

See the documentation for details of the changes:
https://reactrouter.com/en/main/upgrading/v5

There is a compatibility layer that should make the update process a bit smoother, allowing us to use both APIs in parallel while we make the required changes across the app, although it may not cover all use cases. This is documented at:
https://github.com/remix-run/react-router/discussions/8753

Some changes have already been done in preparation for this update:
- https://github.com/tektoncd/dashboard/pull/2209
- https://github.com/tektoncd/dashboard/pull/2453

Some additional changes identified:
- Link `component` prop is no longer supported in v6, `useHref` hook in custom component instead
- When updating search params, `useSearchParams` instead of current manual `URLSearchParams` and `history.push` approach. Not 'required' but is a nice cleanup that we should consider

Once we agree a timeline for this update the next steps will be to review the latest documentation, enable the compatibility layer, and start by updating a small set of routes to validate the process.

AlanGreene:
Recent React Router v5 releases have some fixes for React 18 compatibility so we should try updating React first and see how that goes.

Our previous attempt at updating React Router using the compatibility layer ran into some issues but with most other dependencies now updated, and assuming the React 18 update works as expected, it's time to try this again.

If the compatibility layer doesn't work out we can just update it all in one go but it would be nice to break it down into a series of smaller changes if possible to facilitate easier review and testing.

AlanGreene:
PR #2542 opened to add the compatibility layer and start migrating to the new API. All routes have been converted to use the compatibility layer, including hooks from the new API.

Remaining:
- convert `<Route children>` to `<Route element>`
  - header already using this approach and working as expected, only minor inconvenience was that the `children`  render prop approach is no longer supported to always render content whether the path matches or not. Instead we duplicate the route including a 'splat' (wildcard) for any unmatched paths to ensure it always renders even when on a non-namespaced path.
- `<Switch>` becomes `<Routes>`
  - remove `exact` prop on routes
- updates for `<Link>` to address the `component` prop issue
- optional: `useSearchParams` instead of current manual approach using `URLSearchParams` etc.

AlanGreene:
/lifecycle frozen
[...]

PR #3465, opened 2024-06-18
Update to use React Router data router APIs

AlanGreene:
# Changes

Related to https://github.com/tektoncd/dashboard/issues/2452

There will be be at least one additional PR to refactor some of the remaining resource list / resource details containers and move the configuration of the resource GVK to the routes instead of using custom API hooks.

This is all part of a greater effort to reduce boilerplate, which reduces the maintenance overhead for each additional resource type added, and in turn reduces the barrier to entry for contributors as the code is much simplified and easier to modify. It will also provide greater consistency across resources in terms of layout, functionality, etc. Some resources such as *Runs are likely to remain quite different due to the additional content associated with them, but will be able to leverage the simplified config to reduce the amount of custom code required for each of their views.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


src/api/utils.js
```diff
@@ -235,19 +251,13 @@ export function useSelectedNamespace() {
...
+export function useCollection({ group, kind, params, queryConfig, version }) {
   const { disableWebSocket, ...reactQueryConfig } = queryConfig || {};
-  const query = useQuery(
-    [kind, params].filter(Boolean),
-    () => api(params),
-    reactQueryConfig
-  );
+  const query = useQuery({
+    queryKey: [group, version, kind, params].filter(Boolean),
+    // () => api(params),
```
briangleeson, line 258: Can remove the commented out line now?

src/routes/dashboard.jsx
```diff
@@ -0,0 +1,84 @@
+/*
+Copyright 2019-2024 The Tekton Authors
```
briangleeson, line 2: New file only needs 2024, like the other new files?
```suggestion
Copyright 2024 The Tekton Authors
```

briangleeson (changes requested):
Thanks for the walkthrough. Couple of nits but otherwise looks good overall

src/routes/dashboard.jsx
AlanGreene: The content is largely copied from the old route definitions in App.jsx which is why I kept the original date, just restructured to follow the new API

AlanGreene: although I realise now that I didn't do the same for the pipelines and triggers routes… 🙈 
I'll update them all for consistency. Thanks

src/api/utils.js
AlanGreene: Done

briangleeson (approved):
/lgtm

Issue #3606, opened 2024-09-14
Configuration of Tekton Dashboard service url content path


Romanecek:
# Feature request

Consider url content path when calling Tekton Dashboard service API

# Use case

Tekton Dashboard service might be exposed on nonroot content path for example `https://my-server/ci/tekton` . 
When falling back for external logs the API call from browser is made to `https://my-server/v1/logs-proxy` instead of `https://my-server/ci/tekton/v1/logs-proxy` .

It would be helpful to have a configuration property to specify the correct path for API calls.

# Alternatives

Here I created some raw solution which works for as at this moment: [https://github.com/tektoncd/dashboard/pull/3607](https://github.com/tektoncd/dashboard/pull/3607)

AlanGreene:
Thanks for raising this @Romanecek. It's actually a bug in the client. For some reason the external logs requests weren't using the correct context root like other APIs. PR opened to address this: https://github.com/tektoncd/dashboard/pull/3614

AlanGreene:
Fix is available in the latest nightly release:
- [release.yaml (read-only)](https://storage.googleapis.com/tekton-releases-nightly/dashboard/previous/v20240916-2efc34c79f/release.yaml)
- [release-full.yaml (read-write)](https://storage.googleapis.com/tekton-releases-nightly/dashboard/previous/v20240916-2efc34c79f/release-full.yaml)

It will be included in v0.51.0 due in the next ~1-2 weeks.

PR #3607, opened 2024-09-14
Configuration of Tekton Dashboard service url content path

Romanecek:
This is a solution proposal for the feature request described in [https://github.com/tektoncd/dashboard/issues/3606](https://github.com/tektoncd/dashboard/issues/3606)

Thank you for your feedback


AlanGreene:
Thanks for the PR @Romanecek. As mentioned in https://github.com/tektoncd/dashboard/issues/3606#issuecomment-2352660640 there was actually a bug in the client affecting the external logs API.

The Dashboard client automatically detects its context root and uses this as a prefix for all API requests. However, the external logs were missed so they would always go to the domain root instead.

https://github.com/tektoncd/dashboard/pull/3614 addresses this and once merged I'll spin a nightly release so you can verify it resolves your issue.

The fix will be included in v0.51.0 which is due in ~1-2 weeks.

Closing this.

PR #3688, opened 2024-10-23
Fix style issues in Storybook production build

AlanGreene:
# Changes

The Storybook production build injects some CSS into the iframe used to render the stories. This uses nesting, which while well supported in browsers, causes issues with some build tooling which is not expecting it.

The `vite-plugin-html` minifies the CSS in any `style` tags by default, but this minification process does not account for nested CSS. The net result of this is that some of the styles which are expected to be scoped to `.sb-errordisplay` end up applying globally instead.

The most obvious manifestation of this was on the PipelineRun details story where there was an additional gap between each Task in the TaskTree caused by an unexpected `li+li` rule.

> [!NOTE]
> This only affected display in the Storybook

Disable CSS minification in the HTML plugin since we don't need this for Storybook, and don't make use of `style` tags as part of the template for the Dashboard application itself.

/kind misc

Before:
image

After:
image

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -47,7 +47,9 @@ export default defineConfig(({ mode }) => ({
     target: 'es2022'
   },
   plugins: [
-    createHtmlPlugin({}),
+    // disable CSS minification since we don't use it ourselves and it causes
+    // problems with use of CSS nesting in Storybook production build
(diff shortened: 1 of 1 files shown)


briangleeson (approved):
/lgtm

PR #3667, opened 2024-10-17
Bump @carbon/react from 1.65.0 to 1.68.0

AlanGreene:
# Changes

Replaces https://github.com/tektoncd/dashboard/pull/3662

This includes a number of fixes for the ComboBox component, including properly clearing state when the user clicks the 'clear' button, so our workaround for this is no longer required.

It also includes a number of a11y improvements for various components.

However, it also introduces a breaking change for the ComboBox related to use of the `selectedItem` prop. Previously this did not behave consistently as a fully controlled component, however with this update it now does, resulting in a number of our existing workarounds for the old behaviour becoming problematic. This will likely require significant refactoring of some of the code, specifically around the `NamespacesDropdown` and the `ALL_NAMESPACES` item, as these appear to cause the Carbon component to lock up / go into a loop. This manifested as OOM issues with unit tests in CI, as well as failure of the E2E in CI. Running locally produced intermittent errors, including hanging tests for `HeaderBarContent` and `CreateTaskRun`, both of which exercise the `NamespacesDropdown`.

Temporarily switch use of `selectedItem` to `initialSelectedItem` to maintain the existing behaviour so we can benefit from the other fixes included in this update.

/kind misc

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [ ] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [ ] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [ ] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [ ] Release notes block below has been updated with any user facing changes (new features, significant UI changes, API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

PR #3749, opened 2024-11-07
Adopt ComboBox selectedItem prop

AlanGreene:
# Changes

Follow-up to https://github.com/tektoncd/dashboard/pull/3667 now that we've [updated to `@carbon/react@1.70`](https://github.com/tektoncd/dashboard/pull/3740)

Remove previous workaround and adopt the `selectedItem` prop to make the `ComboBox` component fully controlled.

To avoid the previous issue where the ComboBox would get into a loop or cause the tests to crash due to OOM, we memoise the value passed to `selectedItem` to ensure the same reference is kept if the value is unchanged, e.g. in case of a selected namespace: `{ id: namespace, text: namespace }`
we memoise on the `id` field so the same object is returned unless the namespace changes. This avoid the ComboBox unnecessarily re-rendering, and also handles the case of translated text for the 'All namespaces' option.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/packages/components/src/components/TooltipDropdown/TooltipDropdown.jsx
+++ b/packages/components/src/components/TooltipDropdown/TooltipDropdown.jsx
@@ -74,12 +74,11 @@ const TooltipDropdown = ({
       className={className}
       disabled={disabled}
       id={id}
-      initialSelectedItem={selectedItem}
       items={options}
       itemToString={itemToString}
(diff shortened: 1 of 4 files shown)


briangleeson (approved):
/lgtm

PR #3841, opened 2024-12-13
Improve release process and structure of release manifests

AlanGreene:
# Changes

Reduce unnecessary differences between the read-only and read-write manifests, both to reduce maintenance overhead and to avoid user confusion. There have been a few instances where users thought the read-only mode was broken as they switched the `read-only` flag in the deployment args but didn't replace the associated roles with the appropriate versions. With this update we now include all roles in the manifest, and create bindings for the relevant ones with updated naming to make it clear which grant view access and which grant edit rather than keeping the same name and swapping out the rules.

Update install script to accept input from a previously generated installer manifest. This removes the need to run the `ko` build 4 times (2x installer modes + 2x normal release modes) as part of our releases. Now it builds once, and uses the resulting manifest as input to the normal releases so they can augment it with the appropriate RBAC resources depending on mode (read-only vs. read-write). This also means there's no need for a dedicated read-write installer manifest.

Update the release pipeline's publish task to take advantage of this new approach, and eliminate other unnecessary config.

Use the aggregate ClusterRoles provided by Tekton Pipelines and Triggers instead of defining them ourselves, ensuring we stay in sync with the versions deployed on the cluster. The one exception to this is for `ClusterTask` resources as they're not included in the aggregate roles provided by Tekton Pipelines. We will remove support for these resources in a future release and remove the corresponding rule from the Dashboard backend role.

With this new approach to building the manifests, we no longer need separate overlays or additional patches to modify the resources to match the desired mode. Instead, the installer augments the base installer manifest with the appropriate ClusterRoleBindings or RoleBindings to grant the read-only or read-write permissions as desired, and replaces other config in-place.

Move the remaining resources into the `config` folder for consistency with other Tekton projects.

All of this should be mostly transparent to consumers. They will still use the `installer` or `release-installer` scripts as before. There is some cleanup that should be performed on clusters if upgrading from a previous Dashboard release, by removing the following resources after upgrade:
- `clusterrole/tekton-dashboard-backend`
- `clusterrole/tekton-dashboard-tenant`
- `clusterrolebinding/tekton-dashboard-backend`
- `clusterrolebinding/tekton-dashboard-tenant`

and if the Dashboard was installed with limited namespace visibility (i.e. using the `--tenant-namespaces` installer flag, or directly via the `--namespaces` deployment arg):
- `rolebinding/tekton-dashboard-tenant` in each of the tenant namespaces

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Renamed and updated a number of ClusterRole and related ClusterRoleBinding or RoleBinding resources for improved clarity and to reduce unnecessary differences between the read-only and read-write installs.

🚨 Action required: If upgrading from a previous Dashboard release, please remove the following resources after upgrade:
- `clusterrole/tekton-dashboard-backend`
- `clusterrole/tekton-dashboard-tenant`
- `clusterrolebinding/tekton-dashboard-backend`
- `clusterrolebinding/tekton-dashboard-tenant`

and if the Dashboard was installed with limited namespace visibility (i.e. using the `--tenant-namespaces` installer flag, or directly via the `--namespaces` deployment arg):
- `rolebinding/tekton-dashboard-tenant` in each of the tenant namespaces

These are no longer included in the install and should be removed to avoid unexpected behaviour or inadvertently granting unexpected permissions to the Dashboard's `ServiceAccount`.
```


briangleeson (approved):
/lgtm

Thx for the walkthrough

PR #4344, opened 2025-08-05
Kustomization Missing 100-namespace.yaml file

KorenP1:
Adding ./100-namespace.yaml file to kustomization resources field


korenperetz:
@skaegi honestly i just wanted to apply the missing namespace file and i faced so many regulations. would be awesome if it can be done. Thanks.

PR #4431, opened 2025-09-08
fix: preserve steps execution order on task run details

Maximilien-R:
This PR adds a new utility function `orderStepsFromTaskRun` that reorders steps according to their definition order in the TaskRun's `taskSpec`. This ensures that steps are displayed in the correct execution order in the Tekton Dashboard UI.

### The problem

When a TaskRun has steps referring to StepAction, we encounter two issues that impact user experience in the Tekton Dashboard:
1. **Incorrect step ordering**: Steps referring to StepAction are always displayed at the top of the TaskRun steps list, regardless of their actual execution order defined in the TaskSpec. This creates confusion about the actual flow of execution.
2. **Misplaced step status**: When a step fails, the status information is sometimes misplaced, leading to poor visibility on which steps actually failed. This makes debugging and troubleshooting difficult for users trying to understand TaskRun execution failures.

These issues stem from the fact that steps are being displayed in an order that doesn't match their logical execution sequence as defined in the TaskRun's specification.

taskrunpanel--before
tasktree--before

### The changes

This PR introduces a new utility function `orderStepsFromTaskRun` that addresses both problems by ensuring steps are displayed in their correct execution order:
- Extract step order from `taskRun.status.taskSpec.steps` 
- Reorder steps array to match taskSpec definition order
- Preserve steps not found in taskSpec by appending at end
- Handle edge cases: missing taskSpec, empty arrays, invalid step names

This change should ensures that:
- Steps referring to StepAction are displayed in their correct position within the execution flow
- Step status information is properly aligned with the corresponding steps, improving failure visibility and debugging experience

taskrunpanel--after
tasktree--after

/kind misc

# Release Notes

```release-note
Add `orderStepsFromTaskRun` utility function to reorder steps according to TaskRun taskSpec definition, ensuring steps are displayed in correct execution order in the dashboard UI.
```

### DIFF
--- a/packages/components/src/components/TaskRunTabPanels/TaskRunTabPanels.jsx
+++ b/packages/components/src/components/TaskRunTabPanels/TaskRunTabPanels.jsx
@@ -24,6 +24,7 @@ import {
   getStatus,
   getStepDefinition,
   getStepStatusReason,
+  orderStepsFromTaskRun,
   updateUnexecutedSteps
 } from '@tektoncd/dashboard-utils';
 
@@ -288,9 +289,11 @@ function Logs({
     );
   }
 
+  const orderedSteps = orderStepsFromTaskRun(steps, taskRun);
+
   return (
     <Accordion align="end" className="tkn--task-logs" ordered size="md">
-      {steps.map(step => (
+      {orderedSteps.map(step => (
         <Step
           expandedSteps={{
             ...expandedSteps,
@@ -304,7 +307,7 @@ function Logs({
           selectedTaskId={selectedTaskId}
           taskRunReason={reason}
           step={step}
-          steps={steps}
+          steps={orderedSteps}
           task={task}
           taskRun={taskRun}
         />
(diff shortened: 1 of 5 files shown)


AlanGreene:
Thanks for the PR @Maximilien-R, however I believe this is actually a bug in Tekton Pipelines.

The step order in the status is expected to match the declared order in the spec to avoid just this type of issue across all clients.

Some previous related issues:
- https://github.com/tektoncd/pipeline/issues/3677
- https://github.com/tektoncd/pipeline/issues/3239
  - in this case, the order of these internal containers were explicitly added to the API compatibility policy: https://github.com/tektoncd/pipeline/issues/3239#issuecomment-693427310
  - The same should probably have been done for Step Actions but may have been missed
  - PR that was raised to ensure consistent ordering: https://github.com/tektoncd/pipeline/pull/3256

Can you open an issue against tektoncd/pipeline and include a sample reproduction please?

AlanGreene:
Closing as per comment above

AlanGreene:
This is fixed in Tekton Pipelines v1.6.0 which was released today: https://github.com/tektoncd/pipeline/releases/tag/v1.6.0
Relevant change: https://github.com/tektoncd/pipeline/pull/9039

PR #4906, opened 2026-04-21
Fix label links on PipelineRun and TaskRun detail pages

### LINKED ISSUE #4905: Clicking a label on a PipelineRun/TaskRun detail page links to /undefined?labelSelector=…
### Expected behavior

Clicking a label chip in the `Labels` section of the `RunHeader` on a PipelineRun (or TaskRun) detail page should navigate to the PipelineRuns/TaskRuns list filtered by that label.

### Actual behavior

The link is built as `…/pipelineruns/<name>/undefined?labelSelector=<key>%3D<value>`. Navigating there ends up rendering the PipelineRuns list with the filter applied, but also surfaces an `Error loading undefined` notification because the trailing route segment is interpreted as a resource name.

### Steps to reproduce

1. Open any PipelineRun detail page, e.g. `#/namespaces/<ns>/pipelineruns/<name>`.
2. In the `Labels` column of the header, click any label chip.
3. The URL becomes `#/namespaces/<ns>/pipelineruns/<name>/undefined?labelSelector=…` and an `Error loading undefined` notification appears.

### Root cause

`packages/components/src/components/LabelsWithOverflow/LabelsWithOverflow.jsx` determines the target list from the resource kind:

```js
const resourceType = resource.kind;
```

and `packages/utils/src/utils/router.js` only assigns a base URL when that value matches a known kind:

```js
urls.pipelineRuns.labels = ({ namespace, label, name, resourceType }) => {
  let baseURL;
  if (resourceType === 'PipelineRun') {
    baseURL = urls.pipelineRuns.byNamespace({ namespace });
  } else if (resourceType === 'TaskRun') {
    baseURL = urls.taskRuns.byNamespace({ namespace });
  }
  return filteredURL({ baseURL, label, name });
};
```

The resource passed in comes from `usePipelineRun({ name, namespace })` / `useTaskRun(...)` — i.e. the body returned directly by the Kubernetes API. Typed k8s API responses do not populate `TypeMeta` (`kind` / `apiVersion`) on list/get, so `resource.kind` is `undefined`. That makes `baseURL` `undefined`, `filteredURL` returns the literal string `\"undefined?labelSelector=…\"`, and the router resolves it relative to the current URL — producing the broken link.

Verified against a live cluster:

```
GET /apis/tekton.dev/v1/namespaces/<ns>/pipelineruns/<name>
→ response body keys: [\"metadata\", \"spec\", \"status\"]  // no \"kind\"
```

The existing test in `LabelsWithOverflow.test.jsx` hides the problem by hardcoding `resource.kind = 'PipelineRun'` on the fixture, which isn't representative of real API responses.

### Affected versions

v0.64.0 through v0.67.0 (latest at time of filing). The regression appears to have landed with #4198 \"new runMetadata header design\" (commit 65f7d24a, 2025-05-30), when `LabelsWithOverflow` began driving the header's label links.

### Suggested fix

Have the caller tell `LabelsWithOverflow` which kind the header is for, rather than reading `resource.kind`. The `PipelineRun` and `TaskRun` container components already know their kind via the route. Alternatively, give `urls.pipelineRuns.labels` a fallback when `resourceType` is neither `'PipelineRun'` nor `'TaskRun'`, and update the test fixture so it stops masking the issue.

Nepoxx:
## Changes

Fixes #4905.

`LabelsWithOverflow` read the resource kind from `resource.kind` when building the list page link for each label chip. Kubernetes does not populate `TypeMeta` (`kind` / `apiVersion`) on typed get/list responses, so `resource.kind` was `undefined` at runtime, and `urls.pipelineRuns.labels` returned the literal string `"undefined?labelSelector=…"`. The hash router resolved that relative to the current URL, producing links like `/namespaces/<ns>/pipelineruns/<name>/undefined?labelSelector=…` and surfacing an `Error loading undefined` notification when users clicked a label.

This PR stops depending on `resource.kind` and instead passes an explicit `kind` prop from the `PipelineRun` and `TaskRun` containers through `RunHeader` to `LabelsWithOverflow`. The callers already know which kind they are rendering.

The existing `LabelsWithOverflow` test masked the bug by hardcoding `kind: 'PipelineRun'` on the resource fixture; that has been removed, and a new test asserts the generated label-link URL for both kinds so the regression cannot return silently.

## Submitter Checklist

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing, including updates to minimum requirements e.g. Kubernetes version bumps
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Release notes block below has been filled in, or marked NONE

## Release Notes

```release-note
Fix label links on the PipelineRun and TaskRun detail pages, which previously pointed to `/namespaces/<ns>/<kind>/<name>/undefined?labelSelector=…` and produced an "Error loading undefined" notification when clicked.
```

### DIFF
--- a/packages/components/src/components/LabelsWithOverflow/LabelsWithOverflow.jsx
+++ b/packages/components/src/components/LabelsWithOverflow/LabelsWithOverflow.jsx
@@ -25,6 +25,7 @@ import { urls } from '@tektoncd/dashboard-utils';
 import Link from '../Link';
 
 export default function LabelsWithOverflow({
+  kind,
   namespace,
   resource,
   LinkComponent = Link
@@ -46,7 +47,10 @@ export default function LabelsWithOverflow({
     maxVisibleLabels + maxOverflowLabels
   );
   const totalHiddenLabels = labelEntries.length - maxVisibleLabels;
-  const resourceType = resource.kind;
+  // Kubernetes typed API responses typically do not populate `kind` on the
+  // resource body, so we rely on an explicit prop from the caller rather than
+  // reading `resource.kind` — see tektoncd/dashboard#4905.
+  const resourceType = kind;
   const [isModalOpen, setIsModalOpen] = useState(false);
   const [searchQuery, setSearchQuery] = useState('');
   const [open, setOpen] = useState(false);
--- a/packages/components/src/components/LabelsWithOverflow/LabelsWithOverflow.stories.jsx
+++ b/packages/components/src/components/LabelsWithOverflow/LabelsWithOverflow.stories.jsx
@@ -22,6 +22,7 @@ export default {
 
 export const Default = args => <LabelsWithOverflow {...args} />;
 Default.args = {
+  kind: 'PipelineRun',
   resource: {
     metadata: {
       labels: {
@@ -43,8 +44,7 @@ Default.args = {
         'triggers.tekton.dev/triggers-eventid3':
           'e9742d5b-00e4-4124-9aa1-da2fd670f2da5'
       }
-    },
-    kind: 'PipelineRun'
+    }
   },
   namespace: 'default'
 };
--- a/packages/components/src/components/LabelsWithOverflow/LabelsWithOverflow.test.jsx
+++ b/packages/components/src/components/LabelsWithOverflow/LabelsWithOverflow.test.jsx
@@ -34,8 +34,7 @@ const resource = {
       label14: 'value14',
       label15: 'value15'
     }
-  },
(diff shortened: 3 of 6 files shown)


Nepoxx:
This might still be worth exploring for different go k8s proxy, but closing for now

Issue #4540, opened 2025-10-30
Add eslint-plugin-react-hooks recommended lint settings to the eslint config


briangleeson:
# Description

As part of [the eslint v9 update](https://github.com/tektoncd/dashboard/pull/4511) the `eslint-plugin-react-hooks` package was removed from package.json as it didnt appear to be used in our existing lint config. However, we'd like to add this package back in and include its [recommended rules in our eslint config](https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks#flat-config-eslintconfigjsts) file. Doing so will introduce multiple lint errors and warnings that will need to reviewed and/or addressed

AlanGreene:
Hi @Agrawal-Anamika, thanks for taking a look at this.
Can you please comment `/assign` to assign this issue to yourself?

PR #4978, opened 2026-05-20
fix: add React Hooks es-linting

Agrawal-Anamika:
# Changes

Issue - https://github.com/tektoncd/dashboard/issues/4540
This PR adds React Hooks ESLint rules to the project to enforce proper usage of React Hooks and prevent common mistakes.

**Changes made:**
- Added `eslint-plugin-react-hooks` (v7.1.1) as a dev dependency
- Configured two React Hooks rules in `eslint.config.js`:
  - `react-hooks/rules-of-hooks: 'error'` - Enforces the Rules of Hooks (hooks must be called at the top level, only in React functions)
  - `react-hooks/exhaustive-deps: 'warn'` - Warns about missing dependencies in useEffect, useMemo, and useCallback hooks

**Why these settings:**
- `rules-of-hooks` is set to 'error' because violating the Rules of Hooks causes bugs
- `exhaustive-deps` is set to 'warn' (not 'error') to allow gradual adoption and intentional exceptions with eslint-disable comments

/kind misc
# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -25,6 +25,7 @@ import path from 'path';
 import prettier from 'eslint-plugin-prettier/recommended';
 /* eslint-disable-next-line depend/ban-dependencies */
 import react from 'eslint-plugin-react';
+import reactHooks from 'eslint-plugin-react-hooks';
 import storybook from 'eslint-plugin-storybook';
 
 const gitignorePath = path.join(import.meta.dirname, '.gitignore');
@@ -61,7 +62,8 @@ export default defineConfig([
 
     plugins: {
       // eslint-plugin-notice only works using this approach
-      notice
+      notice,
+      'react-hooks': reactHooks
     },
 
     settings: {
@@ -79,6 +81,10 @@ export default defineConfig([
     },
 
     rules: {
+      // React Hooks rules
+      'react-hooks/rules-of-hooks': 'error',
+      'react-hooks/exhaustive-deps': 'warn',
+
       curly: ['error', 'all'],
       'formatjs/enforce-description': 'off',
       'formatjs/enforce-default-message': 'error',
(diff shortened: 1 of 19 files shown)


eslint.config.js
```diff
@@ -61,7 +62,8 @@ export default defineConfig([
 
     plugins: {
       // eslint-plugin-notice only works using this approach
-      notice
+      notice,
+      'react-hooks': reactHooks
```
AlanGreene, line 66: Was there an issue with adding the [`recommended` config](https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks#flat-config-eslintconfigjsts) as suggested in the issue? It would be good to describe any issues encountered so we can plan how best to address them.

eslint.config.js
```diff
@@ -79,6 +81,10 @@ export default defineConfig([
     },
 
     rules: {
+      // React Hooks rules
+      'react-hooks/rules-of-hooks': 'error',
+      'react-hooks/exhaustive-deps': 'warn',
```
AlanGreene, line 86: See above re. `recommended` config

AlanGreene (commented):
Thanks for the PR @Agrawal-Anamika.

There's a lot of potential for unwanted behaviour changes with some of these updates so it may make sense to break it down into a series of PRs that can be tested, reviewed, and merged separately. That way if issues are encountered it's easier to revert a small set of them instead of the whole thing.

I'd also suggest trying to enable the `recommended` config as suggested in the issue description. We can disable rules for certain files or lines as needed and make incremental progress to having them fully enabled across the codebase.

AlanGreene:
Also when completing the submitter checklist it looks like you forgot to add the 'kind' label.

/kind misc

PR #2921, opened 2023-05-17
Fix AVT issues

kaushalnavneet:
# Issue: 
https://github.ibm.com/org-ids/roadmap/issues/24936#issuecomment-57086599 - issue 10 and issue 11

# Changes
To fix the AVT issues - `The ARIA attributes "aria-label" are not valid for the element <span> with implicit ARIA role "generic"`

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [ ] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [ ] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [ ] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [ ] Release notes block below has been updated with any user facing changes (new features, significant UI changes, API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

PR #2594, opened 2022-11-22
Fix accessibility violations

kaushalnavneet:
Issues:
https://github.ibm.com/org-ids/roadmap/issues/23831#issuecomment-50542297
https://github.ibm.com/org-ids/roadmap/issues/23622
https://github.ibm.com/org-ids/roadmap/issues/23623

PR #5019, opened 2026-06-03
AVT enhance date and duration format

### LINKED ISSUE #5018: Screen reader announces abbreviated date/time instead of descriptive format
During AVT (Accessibility Verification Testing) of the IBM Cloud Tekton page, it was identified that the Latest Run date and time information is not being announced descriptively by screen readers. Currently, screen readers announce the information as `Latest run Apr 7, 4:35 PM. 2m 55s` using abbreviated formats for both the month name and duration.

For optimal accessibility, screen readers should announce this information in a fully descriptive format: `Latest run April 7, 4:35 PM. 2 minutes 55 seconds`. This means using full month names (April instead of Apr) and spelling out time units (minutes and seconds instead of m and s).

Users who rely on assistive technologies such as screen readers may experience difficulty understanding the abbreviated format, leading to confusion and reduced ability to access the control's purpose effectively. 
Code reference: [FormattedDate](https://github.com/tektoncd/dashboard/blob/0cd080bd91e85750c6f2c295d24e6f1dd5d1013e/packages/components/src/components/FormattedDate/FormattedDate.jsx) and [FormattedDuration](https://github.com/tektoncd/dashboard/blob/0cd080bd91e85750c6f2c295d24e6f1dd5d1013e/packages/components/src/components/FormattedDuration/FormattedDuration.jsx) components.

This issue relates to WCAG 2.1 Success Criterion 1.3.1 Info and Relationships (Level A), which requires that information, structure, and relationships conveyed through presentation can be programmatically determined or are available in text.

Agrawal-Anamika:
# Changes
This PR improves accessibility for screen readers by updating date and duration components in the Tekton Dashboard. The changes use two separate span elements:
- One with `aria-hidden="true"` for sighted users (displays narrow/abbreviated format)
- One with `cds--assistive-text` class for screen readers (provides long format with context)

Visual users continue to see abbreviated formats (e.g., "Apr 7, 4:35 PM" and "2m 55s"), while screen readers announce full, descriptive formats (e.g., "April 7, 2024, 4:35 PM" and "Duration: 2 minutes 55 seconds"). The tooltip (title attribute) displays the same full format that screen readers announce, ensuring consistency.

**Example for duration:**
- Visual display: `2m 55s`
- Screen reader: `Duration: 2 minutes 55 seconds`
- Tooltip: `Duration: 2 minutes 55 seconds`

Fixes #5018

/kind bug

### Test Coverage
- ✅ All 617 tests passing
- ✅ 100% code coverage maintained for modified components

# Submitter Checklist

  functionality, content, code)

# Release Notes

None

### DIFF
--- a/packages/components/src/components/FormattedDuration/FormattedDuration.jsx
+++ b/packages/components/src/components/FormattedDuration/FormattedDuration.jsx
@@ -1,5 +1,5 @@
 /*
-Copyright 2019-2024 The Tekton Authors
+Copyright 2019-2026 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -42,33 +42,14 @@ class FormattedDurationWrapper extends Component {
   state = { tooltip: '' };
 
   componentDidMount() {
-    const { intl } = this.props;
-    const tooltip = intl.formatMessage(
-      {
-        id: 'dashboard.run.duration',
-        defaultMessage: 'Duration: {duration}'
-      },
-      {
-        duration: this.durationNode?.textContent
-      }
-    );
+    const tooltip = this.durationNode?.textContent;
     this.setState({
       tooltip
     });
   }
 
   componentDidUpdate() {
-    const { intl } = this.props;
-    const duration = this.durationNode.textContent;
-    const tooltip = intl.formatMessage(
-      {
-        id: 'dashboard.run.duration',
-        defaultMessage: 'Duration: {duration}'
-      },
-      {
-        duration
-      }
-    );
+    const tooltip = this.durationNode?.textContent;
     if (this.state.tooltip !== tooltip) {
       this.setState({ // eslint-disable-line
         tooltip
@@ -77,19 +58,38 @@ class FormattedDurationWrapper extends Component {
   }
 
   render() {
-    const { milliseconds } = this.props;
+    const { intl, milliseconds } = this.props;
     return (
-      <span
-        ref={ref => {
-          this.durationNode = ref;
-        }}
-        title={this.state.tooltip}
-      >
-        <FormattedDuration
-          format="{days} {hours} {minutes} {seconds}"
-          seconds={milliseconds / 1000}
-          unitDisplay="narrow"
-        />
+      <span title={this.state.tooltip}>
+        <span aria-hidden="true">
+          <FormattedDuration
+            format="{days} {hours} {minutes} {seconds}"
+            seconds={milliseconds / 1000}
+            unitDisplay="narrow"
+          />
+        </span>
+        <span
+          className="cds--assistive-text"
+          ref={ref => {
+            this.durationNode = ref;
+          }}
+        >
+          {intl.formatMessage(
+            {
+              id: 'dashboard.run.duration',
+              defaultMessage: 'Duration: {duration}'
+            },
+            {
+              duration: (
+                <FormattedDuration
+                  format="{days} {hours} {minutes} {seconds}"
+                  seconds={milliseconds / 1000}
+                  unitDisplay="long"
+                />
+              )
+            }
+          )}
+        </span>
       </span>
     );
   }
--- a/packages/components/src/components/FormattedDuration/FormattedDuration.test.jsx
+++ b/packages/components/src/components/FormattedDuration/FormattedDuration.test.jsx
@@ -1,5 +1,5 @@
 /*
-Copyright 2020-2024 The Tekton Authors
+Copyright 2020-2026 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -20,10 +20,10 @@ describe('FormattedDuration', () => {
       <FormattedDuration milliseconds={1000} />
     );
     expect(getByText(/1s/i)).toBeTruthy();
-    expect(getByTitle(/1s/i)).toBeTruthy();
+    expect(getByTitle(/Duration: 1 second/i)).toBeTruthy();
 
     render(<FormattedDuration milliseconds={2000} />, { rerender });
     expect(getByText(/2s/i)).toBeTruthy();
-    expect(getByTitle(/2s/i)).toBeTruthy();
+    expect(getByTitle(/Duration: 2 seconds/i)).toBeTruthy();
   });
 });
--- a/packages/components/src/components/LogFormat/LogFormat.test.jsx
+++ b/packages/components/src/components/LogFormat/LogFormat.test.jsx
@@ -1,5 +1,5 @@
 /*
-Copyright 2020-2025 The Tekton Authors
+Copyright 2020-2026 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -361,7 +361,7 @@ describe('LogFormat', () => {
       // accept anything (`.*`) for test purposes as it may be localised
       // and we're more concerned with the structure here
       new RegExp(
-        `<div class="tkn--log-line"><span class="tkn--log-line--timestamp"><span title="${timestamp}">.*</span> </span><span class="tkn--log-line--content">Hello</span></div>`
+        `<div class="tkn--log-line"><span class="tkn--log-line--timestamp"><span title="${timestamp}"><span aria-hidden="true">.*</span><span class="cds--assistive-text">${timestamp}</span></span> </span><span class="tkn--log-line--content">Hello</span></div>`
       )
     );
   });
--- a/packages/components/src/components/DetailsHeader/DetailsHeader.test.jsx
+++ b/packages/components/src/components/DetailsHeader/DetailsHeader.test.jsx
@@ -1,5 +1,5 @@
 /*
-Copyright 2019-2024 The Tekton Authors
+Copyright 2019-2026 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -165,10 +165,10 @@ describe('DetailsHeader', () => {
       }
     };
 
-    const { queryByText } = render(
+    const { getAllByText, queryByText } = render(
       <DetailsHeader {...props} stepStatus={stepStatus} />
     );
-    expect(queryByText(/duration/i)).toBeTruthy();
+    expect(getAllByText(/duration/i).length).toBeGreaterThan(0);
     expect(queryByText(/0s/i)).toBeTruthy();
   });
 
@@ -199,9 +199,9 @@ describe('DetailsHeader', () => {
       }
     };
 
-    const { queryByText } = render(
+    const { getAllByText } = render(
       <DetailsHeader {...props} taskRun={taskRun} type="taskRun" />
     );
-    expect(queryByText(/duration/i)).toBeTruthy();
+    expect(getAllByText(/duration/i).length).toBeGreaterThan(0);
   });
 });
--- a/packages/components/src/components/FormattedDate/FormattedDate.jsx
+++ b/packages/components/src/components/FormattedDate/FormattedDate.jsx
@@ -1,5 +1,5 @@
 /*
-Copyright 2019-2024 The Tekton Authors
+Copyright 2019-2026 The Tekton Authors
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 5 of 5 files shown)


kaushalnavneet (changes requested):
Thanks for working on this accessibility improvement, @Agrawal-Anamika! Looks like we can simplify the implementation. On the issue, @AlanGreene [suggested](https://github.com/tektoncd/dashboard/issues/5018) adding an `aria-label` attribute with the same value as the `title` attribute, which should address the screen reader concerns without requiring dual rendering.
Also, since we don't currently have accessibility-specific test patterns in this project, I'd suggest keeping the existing test structure focused on visual rendering and functional behavior. 
I'd suggest using the simpler `aria-label` approach and testing it first. Let me know if you have any questions!

Agrawal-Anamika:
> Thanks for working on this accessibility improvement, @Agrawal-Anamika! Looks like we can simplify the implementation. On the issue, @AlanGreene [suggested](https://github.com/tektoncd/dashboard/issues/5018) adding an `aria-label` attribute with the same value as the `title` attribute, which should address the screen reader concerns without requiring dual rendering. Also, since we don't currently have accessibility-specific test patterns in this project, I'd suggest keeping the existing test structure focused on visual rendering and functional behavior. I'd suggest using the simpler `aria-label` approach and testing it first. Let me know if you have any questions!

Thanks @kaushalnavneet 
Updated PR as requested

kaushalnavneet:
Thanks, @Agrawal-Anamika! Could you please update the copyright in `LogFormat.test.jsx`? Also, from the `Test Coverage` section, we can remove these:
>✅ New accessibility tests added for dual rendering
✅ Existing tests updated where needed

Agrawal-Anamika:
Screenshot with the accessibility tab
Screenshot 2026-06-04 at 8 16 42 PM

Agrawal-Anamika:
@kaushalnavneet updated PR to only capture format date since format duration was not resolved. Will raise a separate pr for the same

Agrawal-Anamika:
**Screenshot with updated code changes:**
Screenshot 2026-06-05 at 11 00 26 PM

AlanGreene:
Thanks to @kaushalnavneet for reminding me on Slack of some of the previous context and changes around this, e.g. https://github.com/tektoncd/dashboard/pull/2921 and earlier https://github.com/tektoncd/dashboard/pull/2594.

As you can see from those PRs, we previously attempted adding `aria-label` but it raised some additional issues and unfortunately wasn't read by JAWS which is one of the most popular screen readers on Windows. Adding `role="text"` is likely to cause another set of problems and doesn't address the underlying issue that `aria-label` is really only valid on interactive elements. The previous attempt relied on inconsistent screen reader behaviour when encountering it on other elements.

In order to have a single consistent approach to surfacing time and duration across the entire application, while ensuring full accessibility, we should review the current design and make changes to simplify our current approach. I'll create an issue with details of the plan for that including the additional accessibility considerations.

In the meantime, we can address this specific issue by providing two separate elements, one for sighted users, and another for screen readers to ensure they don't have problems with the abbreviated / narrow strings.

Taking `FormattedDuration` as an example, it would look something like this:

```js
return (
  <span title={this.state.tooltip}>
    <span
      aria-hidden="true"
      ref={ref => {
        this.durationNode = ref;
      }}
    >
      <FormattedDuration
        format="{days} {hours} {minutes} {seconds}"
        seconds={milliseconds / 1000}
        unitDisplay="narrow"
      />
    </span>
    <span className="cds--assistive-text">
      {intl.formatMessage({
          id: 'dashboard.run.duration',
          defaultMessage: 'Duration: {duration}'
        },
        {
          duration: (
            <FormattedDuration
              format="{days} {hours} {minutes} {seconds}"
              seconds={milliseconds / 1000}
              unitDisplay="long"
            />
          )
        }
      )}
    </span>
  </span>
);
```

- each string is wrapped in a span so they can be treated appropriately
- the existing `ref` used to populate the tooltip is moved to the new inner span for the visible content to maintain the existing behaviour
- `aria-hidden` is used on this to hide the narrow version from screen readers, this was the problematic content
- `cds--assistive-text` is a utility class provided by Carbon to visually hide an element while allowing it to be processed by screen readers, this is applied to the new long formatted version to ensure it's read correctly (minutes instead of meters etc.)
- does not use the `hidden` attribute which also hides the content from screen readers

A similar approach could be taken for `FormattedDate` if needed.

Agrawal-Anamika:
Thanks @AlanGreene for the review. I updated code as per suggestion
Tested with [IBM equal access accessibility checker](https://chromewebstore.google.com/detail/ibm-equal-access-accessib/lkcagbfjnkomcinoddgooolagloogehp?hl=en-GB&utm_source=ext_sidebar)

Screenshot 2026-06-15 at 12 41 44 PM

packages/components/src/components/FormattedDuration/FormattedDuration.jsx
```diff
@@ -79,17 +80,32 @@ class FormattedDurationWrapper extends Component {
...
+            this.fullDurationNode = ref;
+          }}
+          hidden
+        >
+          <FormattedDuration
+            format="{days} {hours} {minutes} {seconds}"
+            seconds={milliseconds / 1000}
+            unitDisplay="long"
+          />
+        </span>
```
AlanGreene, line 92: This additional third element is unnecessary.

We can achieve the desired outcome with just two elements (one visible, a second for screen readers) as shown in https://github.com/tektoncd/dashboard/pull/5019#issuecomment-4652490019

However, if you want to update the tooltip to include the long format instead of the current narrow format as you have done in this PR you can move the ref to the assistive node which allows for additional simplification of the code:

```diff
diff --git a/packages/components/src/components/FormattedDuration/FormattedDuration.jsx b/packages/components/src/components/FormattedDuration/FormattedDuration.jsx
index 0b98063d..5fda91bb 100644
--- a/packages/components/src/components/FormattedDuration/FormattedDuration.jsx
+++ b/packages/components/src/components/FormattedDuration/FormattedDuration.jsx
@@ -42,34 +42,14 @@ class FormattedDurationWrapper extends Component {
   state = { tooltip: '' };
 
   componentDidMount() {
-    const { intl } = this.props;
-    const fullDuration = this.fullDurationNode?.textContent || '';
-    const tooltip = intl.formatMessage(
-      {
-        id: 'dashboard.run.duration',
-        defaultMessage: 'Duration: {duration}'
-      },
-      {
-        duration: fullDuration
-      }
-    );
+    const tooltip = this.durationNode?.textContent;
     this.setState({
       tooltip
     });
   }
 
   componentDidUpdate() {
-    const { intl } = this.props;
-    const fullDuration = this.fullDurationNode?.textContent || '';
-    const tooltip = intl.formatMessage(
-      {
-        id: 'dashboard.run.duration',
-        defaultMessage: 'Duration: {duration}'
-      },
-      {
-        duration: fullDuration
-      }
-    );
+    const tooltip = this.durationNode?.textContent;
     if (this.state.tooltip !== tooltip) {
       this.setState({ // eslint-disable-line
         tooltip
@@ -78,33 +58,37 @@ class FormattedDurationWrapper extends Component {
   }
 
   render() {
-    const { milliseconds } = this.props;
+    const { intl, milliseconds } = this.props;
     return (
       <span title={this.state.tooltip}>
-        <span
-          aria-hidden="true"
-          ref={ref => {
-            this.durationNode = ref;
-          }}
-        >
+        <span aria-hidden="true">
           <FormattedDuration
             format="{days} {hours} {minutes} {seconds}"
             seconds={milliseconds / 1000}
             unitDisplay="narrow"
           />
         </span>
-        <span className="cds--assistive-text">{this.state.tooltip}</span>
         <span
+          className="cds--assistive-text"
           ref={ref => {
-            this.fullDurationNode = ref;
+            this.durationNode = ref;
           }}
-          hidden
         >
-          <FormattedDuration
-            format="{days} {hours} {minutes} {seconds}"
-            seconds={milliseconds / 1000}
-            unitDisplay="long"
-          />
+          {intl.formatMessage(
+            {
+              id: 'dashboard.run.duration',
+              defaultMessage: 'Duration: {duration}'
+            },
+            {
+              duration: (
+                <FormattedDuration
+                  format="{days} {hours} {minutes} {seconds}"
+                  seconds={milliseconds / 1000}
+                  unitDisplay="long"
+                />
+              )
+            }
+          )}
         </span>
       </span>
     );
```

This way we still have just two elements, one ref, and eliminate the duplicate `intl.formatMessage` for populating the tooltip.

packages/components/src/components/FormattedDuration/FormattedDuration.test.jsx
```diff
@@ -20,10 +20,10 @@ describe('FormattedDuration', () => {
       <FormattedDuration milliseconds={1000} />
     );
     expect(getByText(/1s/i)).toBeTruthy();
-    expect(getByTitle(/1s/i)).toBeTruthy();
+    expect(getByTitle(/Duration:.*1.*second/i)).toBeTruthy();
```
AlanGreene: We should assert an exact match here as this is what we're expecting in the title attribute:
```suggestion
    expect(getByTitle(/Duration: 1 second/i)).toBeTruthy();
```

packages/components/src/components/FormattedDuration/FormattedDuration.test.jsx
```diff
@@ -20,10 +20,10 @@ describe('FormattedDuration', () => {
...
 
     render(<FormattedDuration milliseconds={2000} />, { rerender });
     expect(getByText(/2s/i)).toBeTruthy();
-    expect(getByTitle(/2s/i)).toBeTruthy();
+    expect(getByTitle(/Duration:.*2.*second/i)).toBeTruthy();
```
AlanGreene: Same as above on the exact match. We should also ensure we're asserting the correct plural form (i.e. seconds instead of second):
```suggestion
    expect(getByTitle(/Duration: 2 seconds/i)).toBeTruthy();
```

packages/components/src/components/LogFormat/LogFormat.test.jsx
```diff
@@ -361,7 +361,7 @@ describe('LogFormat', () => {
       // accept anything (`.*`) for test purposes as it may be localised
       // and we're more concerned with the structure here
       new RegExp(
-        `<div class="tkn--log-line"><span class="tkn--log-line--timestamp"><span title="${timestamp}">.*</span> </span><span class="tkn--log-line--content">Hello</span></div>`
+        `<div class="tkn--log-line"><span class="tkn--log-line--timestamp"><span title="${timestamp}"><span aria-hidden="true">.*<\\/span><span class="cds--assistive-text">${timestamp}<\\/span><\\/span> <\\/span><span class="tkn--log-line--content">Hello<\\/span><\\/div>`
```
AlanGreene: The backslashes are unnecessary here:

```suggestion
        `<div class="tkn--log-line"><span class="tkn--log-line--timestamp"><span title="${timestamp}"><span aria-hidden="true">.*</span><span class="cds--assistive-text">${timestamp}</span></span> </span><span class="tkn--log-line--content">Hello</span></div>`
```

AlanGreene (changes requested):

packages/components/src/components/FormattedDuration/FormattedDuration.test.jsx
Agrawal-Anamika: done

packages/components/src/components/FormattedDuration/FormattedDuration.test.jsx
```diff
@@ -20,10 +20,10 @@ describe('FormattedDuration', () => {
       <FormattedDuration milliseconds={1000} />
     );
     expect(getByText(/1s/i)).toBeTruthy();
-    expect(getByTitle(/1s/i)).toBeTruthy();
+    expect(getByTitle(/Duration: 1 second/i)).toBeTruthy();
 
     render(<FormattedDuration milliseconds={2000} />, { rerender });
     expect(getByText(/2s/i)).toBeTruthy();
-    expect(getByTitle(/2s/i)).toBeTruthy();
+    expect(getByTitle(/Duration: 2 second/i)).toBeTruthy();
```
AlanGreene:
```suggestion
    expect(getByTitle(/Duration: 2 seconds/i)).toBeTruthy();
```

AlanGreene (changes requested):
One last change missed from the review.

Can you also update the commit message to include the description you've added to the PR? This is useful when reviewing the change history outside GitHub. The example given for duration should also be updated to use the long format.

packages/components/src/components/FormattedDuration/FormattedDuration.test.jsx
Agrawal-Anamika: done

AlanGreene (approved):
Looks good. Thanks @Agrawal-Anamika 👍

kaushalnavneet (approved):
/lgtm

kaushalnavneet:
Thanks, folks! @Agrawal-Anamika @AlanGreene

Issue #4539, opened 2025-10-30
Update React Router


AlanGreene:
~~Blocked by #3492 as React v18 is the minimum supported version for React Router v7~~

See migration guide: https://reactrouter.com/7.18.3/upgrading/v6

This can be tackled in phases:
1. update to latest v6
2. enable all feature flags (can be split as needed)
3. update to v7 (no breaking changes)

iamrajiv:
hi @AlanGreene can i work on this?

AlanGreene:
Hi @iamrajiv, thanks for your interest. Yes please reach out if you have any questions, and I would strongly suggest following the phased approach above if possible to aid in easier review and testing of the changes.

/assign @iamrajiv

PR #5210, opened 2026-09-01
chore: enable react-router v7 future flags

kaushalnavneet:
# Changes

Addressing phase 2 work of https://github.com/tektoncd/dashboard/issues/4539. The first phase is already done since we are using the [latest v6 version](https://github.com/remix-run/react-router/releases#release-react-router@6.30.6) anyway - `react-router-dom@^6.30.6`

## Enable future flags

Enable 5 of the 6 react-router v7 future flags in preparation for the v7 migration.

- `App.jsx`: pass a `future` options object to `createHashRouter` with the 5 applicable flags
- `packages/components/src/utils/test.jsx`: mirror the same flags on `createBrowserRouter` so unit tests run under the same router conditions as production

Tested locally to confirm the basic navigation works along with namespace switching, etc.

### v7_startTransition - intentionally excluded

Enabling this flag wraps `navigate()` calls in `React.startTransition`, making router state updates non-urgent. This breaks the global namespace dropdown: `selectNamespace()` (external context) fires synchronously while the navigation is deferred, causing `useLocation()` / `useMatches()` to reflect the old route on re-render. The result is the URL updates correctly but the table doesn't refresh until a second interaction.

The react-router migration guide provides an explicit opt-out for this flag in v7 for exactly this kind of scenario. 

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

## TODO for the next phase (Upgrade to v7):

With all flags enabled, the upgrade itself has no breaking changes in the router behaviour. However, there are few changes needed:
- Package swap: `react-router-dom` is replaced by `react-router`. Update peer deps (add v7 and remove v5 & v6)
- Update imports ^
- In v7, the `state` prop is removed from `<Navigate>`. The code in `App.jsx` needs to be fixed accordingly
- `useLinkClickHandler` state option in `Link.jsx` needs to be fixed as well - pass it as a part of the `to` object
- `devDependency` update - can be done later as well: `storybook-addon-remix-react-router@^3.1.0` is used in 9 story files. Version 3.x supports `react-router` v6. For v7, need to upgrade to v4 of that addon (which added v7 support).


AlanGreene (changes requested):
Enabling these flags has broken the global namespace dropdown. For example:
1. load the app, verify namespaces dropdown shows 'All Namespaces' by default
2. navigate to PipelineRuns page, verify URL is as expected
3. select a namespace from the global namespace dropdown in the app header
4. verify the contents are filtered as expected and the URL is also updated to reflect the selected namespace
5. click the 'X' to clear the selected namespace
6. note the URL has been updated to remove the namespace as expected
7. table not updated to show runs from all namespaces
8. click the 'X' a second time and see the table updates

Considering this breaking change, we should also enable the flags in our use of `createBrowserRouter` for the unit tests to ensure consistency and see if it helps to identify the source of the issue.

kaushalnavneet:
Hi @AlanGreene, followed the steps from 1 to 6 and tried to reproduce it locally with the changes we have. After selecting the `tekton-dashboard` namespace, it shows the runs for that namespace. And when I click the `x` icon, the URL is updated to remove the namespace, and the table is also updated to show runs from all namespaces:

https://github.com/user-attachments/assets/e7eb4b9d-5292-43b1-a59f-426de78547cc

AlanGreene:
Are you running against a dev or production build? I see `:9097` in your recording so assuming a prod build. Could be a difference there as I was testing against a local dev build.

Can you confirm if you can reproduce it against your local dev env (`npm start`, localhost:8000)?

If it's dev-only, it could be quite annoying as that would mean we can no longer trust local dev behaviour so we'll definitely need to fix it.

kaushalnavneet:
You're right! Somehow my local dev was terminated, and started testing where it was port-forwarding to, i.e. `9097`.
I was able to reproduce it on my local dev. I've made a fix that seems to be working. Have enabled the future flags in the tests as well.

AlanGreene:
Thanks for the update. Can you also fix the PR title, description, and commit comment to match the recent changes and squash your commits when done?

Disabling the transition flag is OK as a short-term workaround to get the React Router update in but we'll need to revisit in future if we decide to use any view transitions. It will also need to be documented for consumers of the components / utils packages.

AlanGreene (approved):
Thanks Nav 👍 
/lgtm

PR #5227, opened 2026-09-11
feat(logs): pass preferences to external providers. Fixes #4981

### LINKED ISSUE #4981: Update external log support to pass timestamp and log level preferences to the log provider
# Feature request

Update the API layer to support passing the user's timestamp and log level settings on the request to the external log provider.

https://github.com/tektoncd/dashboard/blob/6fb7878f7c5d99bd034c232ccc5084e780b225d8/src/api/index.js#L131

The `queryParams` included on the URL should be updated to include:
- `timestamps`: boolean
- `logLevel`: string. This should be repeated for each selected log level, e.g. `…&logLevel=info&logLevel=error`
  - `queryParams.append('logLevel', value)` can be used to achieve this

Documentation at https://github.com/tektoncd/dashboard/blob/6fb7878f7c5d99bd034c232ccc5084e780b225d8/docs/logs.md#logs-persistence should also be updated to reflect the new functionality.

Update callers to provide these new fields.

Note: External log providers may choose to ignore these fields, they are provided as an optional additional capability if they wish to use it. Therefore, the client code should make no assumptions about resulting changes to the returned content, its only responsibility is to ensure the query params are set correctly on the URL.

# Use case

Some logs contain a large amount of data that may not be relevant to all users. In some cases, a user may only be interested in errors or warnings for example, or not want to see lower level debug or trace content. We support filtering these in the client but this still requires the external log provider to return all of the content first which can be slow and/or wasteful. Being able to specify the desired log levels on the API call to the external provider can provide a much more efficient response, resulting in better performing user experience and less load on backend systems.

Similarly, to better understand where time is spent, it can be useful to display timestamps on log lines. This may not be helpful in all cases though, so being able to disable this can produce cleaner / simpler logs in some cases. While we support toggling this in the client, some users may use a different date/time format in their logs with the external provider and being able to toggle this at the API layer allows those users to still benefit from the functionality.

# Alternatives

n/a

# Additional context

akashchamp:
# Changes

Passes the user's timestamp and selected log-level preferences to external log providers.

- Adds `timestamps=<true|false>` and repeated `logLevel` parameters to external log URLs.
- Propagates the preferences through PipelineRun and TaskRun fallback requests and external raw-log links.
- Documents the optional parameters for external providers.
- Adds URL, fallback, and toolbar coverage.

Fixes #4981

Validation: targeted tests (137), the full test suite (614 passing, 1 skipped), lint, and production build pass. A live UI check requires an authenticated Tekton cluster and configured external log provider; neither is available on the remote host, so no cluster state or credentials were changed.

# Submitter Checklist

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included for this user-facing change
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included for changed functionality
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Has a kind label
- [x] Release notes block below has been updated for this user-facing change
- [x] Release notes do not require action from users switching to the new release

# Release Notes

```release-note
External log providers can now receive the user's timestamp and log-level preferences.
```

### DIFF
--- a/docs/logs.md
+++ b/docs/logs.md
@@ -90,7 +90,7 @@ It can be enabled by providing the `--external-logs` flag to the installer scrip
 When configured, the Dashboard will first attempt to load pod logs normally, and if they're unavailable will fallback to the provided external logs service by making a `GET` request to the provided endpoint with the following format:
 
 ```
-GET <external-logs>/<namespace>/<podName>/<container>?startTime=<stepStartTime>&completionTime=<stepCompletionTime>
+GET <external-logs>/<namespace>/<podName>/<container>?startTime=<stepStartTime>&completionTime=<stepCompletionTime>&timestamps=<true|false>&logLevel=<selectedLogLevel>&logLevel=<selectedLogLevel>
 ```
 
 - `namespace`: the namespace containing the run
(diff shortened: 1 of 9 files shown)


AlanGreene:
Thanks for the PR. Please see the [Tekton community AI contribution policy](https://github.com/tektoncd/community/blob/main/ai-contribution-policy.md) and update the PR accordingly.

